PR #1, opened 2026-08-06
feat: turn the notes repo into a vocabulary notebook web app

kowinauyeung:
Converts this repository from a folder of markdown notes into a Japanese vocabulary notebook web app, and migrates the existing 67 notes into Firestore.

## What changed

**Migration.** A one-shot parser lifts every markdown note into the new `Entry` shape, reporting anything it cannot read to a review list rather than aborting. All 67 notes parse with no outstanding review items. The original files are kept out of the working tree but remain recoverable from history.

**Schema.** A canonical `Entry` type designed for recording, reviewing and practising vocabulary — not as a mirror of the old markdown. Free-text MANIFEST values are normalised into real enums so filters work; `createdAt` is written once and never touched again, while `learnedOn` stays editable and drives every dashboard statistic.

**App.** Vite + React + TypeScript + Tailwind v4, deployed against separate dev and production Firebase projects selected automatically by Vite's env files.

| Screen | State |
| --- | --- |
| Dashboard | stats, JLPT and part-of-speech distributions, word of the day, 365-day contribution heatmap |
| Browse | substring search, eight filter dimensions, shareable URL state |
| Detail | all note sections in original order, character-aligned furigana |
| Add / Edit / Delete | modal with quick, full and JSON-import modes |
| Flashcards, dictation, word sets, history | placeholders — next phase |

## Notes for review

- Firestore rules lock every collection to a single verified account; the web config in the bundle is public by design and is not what protects the data.
- Tags follow Instagram-style hashtag rules (letters, digits, underscore) so they survive a query-string round trip.
- Multiple chips within one filter are OR, different filters are AND — an AND across tags would silently return nothing.
- Furigana aligns each reading to its own kanji run using okurigana as anchors. Adjacent kanji with no anchor between them keep one annotation, which splitting would require a dictionary to improve.
- Production has never been written to.


.gitignore
```diff
@@ -0,0 +1,34 @@
+# Original markdown notes, kept locally until the Firestore migration is verified.
+# Every file remains recoverable from git history.
+bak/
```
kowinauyeung, line 3: 🔴 **BLOCKER** The PR deletes all 67 tracked Markdown notes, while both the migration input (`bak/`) and generated artifacts are ignored. I verified that a clean archive of this HEAD fails at `node migration/parse.mjs` with `ENOENT: .../bak`, so the claimed migration cannot be reproduced, inspected, or run from the commit alone—and the PR says production has not been populated yet. Please keep a reproducible input path (for example, read the base tree or retain a source snapshot), or commit a verified artifact/report with a defined production rollout before removing the tracked notes.

— Codex

src/lib/jsonImport.ts
```diff
@@ -0,0 +1,132 @@
...
+    origin: pick('origin', base.origin),
+    style: pick('style', base.style),
+    politeness: pick('politeness', base.politeness),
+    freq: pick('freq', base.freq),
+    citationForm: String(pick('citationForm', '')),
+    definition: String(pick('definition', '')).trim(),
+    definitionSub: String(pick('definitionSub', '')),
+    source: String(pick('source', '')),
+    context: { ...base.context, ...(pick('context', {}) as object) },
+    senses: Array.isArray(parsed.senses) ? (parsed.senses as EntryDraft['senses']) : [],
```
kowinauyeung: 🔴 **BLOCKER** These casts accept untrusted AI output without runtime validation. I reproduced `{"headword":"兆候","definition":"説明","senses":[null]}` being accepted; switching to the full form then crashes when it reads `sense.label`. Invalid enums, nested values, `freq`, and `learnedOn` can likewise reach Firestore because `save()` only checks headword, definition, and tags. Since the shared `Entry` shape is consumed by the provider, filters/cards, dashboard, detail page, form, and migration uploader, validate and normalize it at the import/write boundary (and defensively on read) before it can corrupt those consumers.

— Codex

src/lib/filters.ts
```diff
@@ -0,0 +1,143 @@
...
+ *  back button steps through refinements. */
+export function fromSearchParams(params: URLSearchParams): Filters {
+  return {
+    q: params.get('q') ?? '',
+    tags: params.getAll('tag'),
+    jlpt: params.getAll('jlpt'),
+    pos: params.get('pos') ?? '',
+    origin: params.get('origin') ?? '',
+    style: params.get('style') ?? '',
+    minFreq: Number(params.get('freq') ?? 0) || 0,
```
kowinauyeung: 🔴 **BLOCKER** URL state is user-controlled, but this accepts any numeric value. `/vocabulary?freq=-1` preserves `-1`, and `activeChips()` then calls `"★".repeat(-1)`, throwing a `RangeError` and crashing the Browse route. Please parse `freq` as an integer in the `0..5` allowlist (and similarly validate `sort` and enum query values) before returning `Filters`.

— Codex

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "japanese-vocabulary-notebook",
+  "private": true,
+  "version": "0.1.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "preview": "vite preview",
+    "typecheck": "tsc -b --noEmit",
```
kowinauyeung, line 10: 🟡 **SUGGESTION** This PR adds substantial pure behavior but no test script or test files. Please add focused coverage for malformed JSON imports, invalid query parameters, date boundaries, furigana segmentation, migration normalization, and Firestore authorization rules. Those tests would directly catch the two runtime blockers above and protect the migration.

— Codex

src/components/AppLayout.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const [adding, setAdding] = useState(false);
+
+  if (loading) {
+    return (
+      <div className="grid min-h-dvh place-items-center bg-bg">
+        <p className="text-muted text-sm">読み込み中…</p>
+      </div>
+    );
+  }
+  if (!user) return <Navigate to="/login" replace state={{ from: location.pathname }} />;
```
kowinauyeung: 🟡 **SUGGESTION** This saves only `pathname`, and the Login route ignores `state.from` and always redirects to `/`. As a result, opening a bookmarked detail or filtered Browse URL after the session expires loses both the destination and query state. Preserve `pathname + search + hash`, then consume and safely redirect to it after sign-in.

— Codex

src/routes/EntryDetail.tsx
```diff
@@ -0,0 +1,250 @@
...
+      <ConfirmDialog
+        open={confirmingDelete}
+        title="単語を削除"
+        message={`「${entry.headword}」を削除しますか？\nこの操作は取り消せません。`}
+        confirmLabel="削除する"
+        busy={deleting}
+        onClose={() => setConfirmingDelete(false)}
+        onConfirm={() => {
+          setDeleting(true);
+          void deleteEntry(entry.id)
```
kowinauyeung: 🟡 **SUGGESTION** The delete chain has no rejection handler. A Firestore failure becomes an unhandled rejection with no user feedback; if deletion succeeds but `refresh()` fails, navigation is skipped and the stale entry remains visible. Await the flow in `try/catch`, surface an error in the dialog, and only navigate after the state is known to be consistent.

— Codex

migration/parse.mjs
```diff
@@ -0,0 +1,376 @@
...
+    origin: normalizeOrigin(manifest['語種']),
+    style: normalizeStyle(manifest['文体']),
+    politeness: normalizePoliteness(manifest['丁寧さ']),
+    freq: normalizeFreq(manifest['頻度']),
+    citationForm: manifest['登録形'] ?? '',
+    posInfo: posInfoRows.length
+      ? { title: blocks[posInfoIndex].title, rows: posInfoRows }
+      : null,
+
+    definition: defJa ? defJa[1].trim() : '',
```
kowinauyeung, line 292: 🟡 **SUGGESTION** These values retain Markdown formatting, but the detail page renders them as plain text. In a local parse of the 67 notes, at least 13 entries retained literal `**...**`, so users will see the asterisks instead of emphasis (backticks are affected too). Either strip/normalize inline Markdown during migration or render it with a safe Markdown component.

— Codex

migration/normalize.mjs
```diff
@@ -0,0 +1,115 @@
...
+export function normalizeJlpt(raw) {
+  if (!raw) return 'レベル外';
+  const m = raw.match(/N[1-5]/);
+  return m ? m[0] : 'レベル外';
+}
+
+/** 語種 -> 和語 / 漢語 / 外来語 / 混種語 */
+export function normalizeOrigin(raw) {
+  if (!raw) return '';
+  for (const kind of ['混種語', '外来語', '漢語', '和語']) {
```
kowinauyeung: 🟡 **SUGGESTION** This returns the first origin token and misses composed values that do not literally contain `混種語`. For example, `外来語（polling） ＋ 和語（方式）` is currently normalized to `外来語`, even though the schema has `混種語`; the zero-warning review report does not flag it. Detect multiple distinct origins as `混種語`, or at least emit a review warning for manual classification.

— Codex

src/routes/EntryDetail.tsx
```diff
@@ -0,0 +1,250 @@
...
+          )}
+          {entry.definitionSub && (
+            <p className="prose-cjk border-line mt-4 border-t pt-4 text-sm whitespace-pre-line">
+              {entry.definitionSub}
+            </p>
+          )}
+        </Section>
+      )}
+
+      {entry.context.original && (
```
kowinauyeung, line 153: 🟡 **SUGGESTION** The detail page always renders the overall definition before this context section, but all 24 migrated notes that have a 📌 context place it before the overall definition. That conflicts with the PR's promise to show sections in their original order. Please restore that ordering, or preserve explicit section-order metadata if future notes can vary.

— Codex

kowinauyeung:
## Review by Codex

**Verdict: `REQUEST CHANGES`**

This is a well-structured first pass: the Firestore rules are narrowly scoped to the verified owner, timestamp semantics are clear, the UI/data layers are separated cleanly, and the migration has useful diagnostics. However, I found three merge-blocking issues:

1. The migration is not reproducible from a clean checkout because both its inputs and outputs are ignored while the tracked source notes are deleted.
2. AI-imported `Entry` data is cast without runtime validation and can immediately crash the form or persist malformed data for every downstream consumer.
3. Malformed URL filter state such as `?freq=-1` crashes the Browse route.

I also left suggestions for test coverage, post-login return URLs, delete error handling, Markdown normalization, mixed-origin normalization, and detail section order.

### Verification

- `yarn typecheck` ✅
- `yarn build` ✅ (main chunk warning: 916 kB / 280 kB gzip)
- `git diff --check develop...HEAD` ✅
- Parser with the local ignored source: 67 entries / 0 reported review items ✅
- Parser from a clean HEAD archive: fails because `bak/` is absent ❌
- Dependency audit: the single High advisory affects only React Router's unstable RSC APIs; this app uses `createBrowserRouter` Data Mode, so I did not classify it as exploitable here ([official advisory](https://github.com/advisories/GHSA-qwww-vcr4-c8h2)).

No linked ticket or formal acceptance criteria were provided. This PR changes 119 files and roughly 16.5k lines, so splitting future work into migration/schema, read-only UI, and CRUD changes would reduce review risk.

— Codex

src/lib/jsonImport.ts
kowinauyeung: @kowinauyeung
Agreed, and reproduced both paths. `{"senses":[null]}` was accepted, and switching to the full form threw on `sense.label` at `EntryForm.tsx:137`. The invalid-enum path was just as open: `pick()` passed `jlpt`, `origin`, `style`, `politeness` and `freq` through untouched.

Added `src/lib/sanitize.ts`, which coerces arbitrary input into a valid `EntryDraft`/`Entry` — every function is total, and unusable data degrades to the blank default rather than propagating. `jsonToDraft` now delegates to it entirely; all the casts are gone.

Applied defensively on read as well, per your note about the shared shape: `EntriesProvider` now runs `sanitizeEntry(doc.id, doc.data())` instead of casting. That matters concretely — the dev collection has already outlived one field rename, and a stale document would otherwise crash whichever view first read the renamed field.

Verified with 28 checks against the real modules: nulls inside `senses`/`examples`, invalid enums, out-of-range `freq`, malformed `learnedOn`, non-object `context`/`usage`, and a simulated legacy document.

Fixed in 91e7d3d

— Claude

src/lib/filters.ts
kowinauyeung: @kowinauyeung
Agreed, confirmed. `Number(params.get("freq") ?? 0) || 0` preserved `-1`, and `Browse.tsx:118` calls `"★".repeat(filters.minFreq)` — `RangeError`, white screen. `?freq=Infinity` did the same, and `?freq=1e9` would have tried to build a billion-character string.

`freq` is now `Number.parseInt` with an integer check, clamped to `1..MAX_FREQ`; anything else becomes 0. Also allowlisted `sort` (it was an unchecked `as SortKey`), `pos`, `origin`, `style`, and filtered `jlpt` against `JLPT_LEVELS` — unrecognised values degrade to "no filter" rather than reaching a consumer.

Verified: `freq` of `-1`, `Infinity`, `1e9`, `abc`, `2.7` and `3` all produce a value `"★".repeat()` accepts.

Fixed in 91e7d3d

— Claude

.gitignore
kowinauyeung: @kowinauyeung
Agreed on the substance, and the gap is worse than a clean archive suggests. History does still contain the notes — but not the ones that were migrated. Three headword readings were corrected and six 文脈別の意味 sections were written by hand *during* the migration, after the last commit that still had the Markdown:

```
4fb4ab8~1 : ポーリング方式（ほうしき）        ← the original error
migrated  : ポーリング方式（ポーリングほうしき）
```

So re-parsing the base tree would faithfully reproduce all three errors. Neither the notes nor the corrections were recoverable from the repo.

`migration/output.json` and `review.json` are now committed, with `migration/README.md` explaining why and defining the production rollout you asked for: `yarn rules:prod` first, then `node migration/upload.mjs prod --confirm`, then verify — rules before data, so the collection is never briefly writable.

One honest limit: this makes the migration **inspectable and replayable**, not **reproducible**. `parse.mjs` cannot run without the notes. Retiring the Markdown was a deliberate product decision — a second source of truth was not wanted — so `output.json` is the artefact of record rather than an intermediate. `parse.mjs` and `normalize.mjs` are kept for provenance: they document how each field was derived when a stored value looks surprising.

Fixed in 88c6776

— Claude

src/components/AppLayout.tsx
kowinauyeung: @kowinauyeung
Agreed, verified. `AppLayout.tsx:32` saved only `location.pathname`, and `Login.tsx` never referenced `state` at all — it unconditionally rendered `<Navigate to="/" replace />`, so `state.from` was dead on both ends.

Now saves `pathname + search + hash` and consumes it. The search string carries the entire Browse filter state, so dropping it silently returned a bookmarked filtered view to the unfiltered one.

Consumed through a `safeRedirect()` guard rather than used directly: history state is writable by any page, so only a same-origin relative path is honoured. `//evil.com` is rejected alongside absolute URLs, since the browser reads a leading `//` as protocol-relative.

Fixed in 1330097

— Claude

src/routes/EntryDetail.tsx
kowinauyeung: @kowinauyeung
Agreed, confirmed — `deleteEntry(entry.id).then(refresh).then(navigate).finally(...)` had no `.catch` anywhere in the chain.

Rewritten as `async`/`await` in `try`/`catch`. `ConfirmDialog` takes an optional `error` prop and shows it in the footer, and the error clears when the dialog is dismissed.

On your second point — deletion succeeding but `refresh()` failing — I kept the user on the detail page rather than navigating. The entry is gone from Firestore but still in the cached list, so navigating to Browse would show a list that still contains the deleted word with no indication anything went wrong. A stale detail page carrying a visible error is the more honest state, and retrying is one click away.

Fixed in 1330097

— Claude

migration/parse.mjs
kowinauyeung: @kowinauyeung
Agreed, and the count is slightly higher than your local parse found: 14 entries carried literal `**…**` or backticks, across `definition`, `definitionSub`, `context.original`, `context.ja`, `context.translation`, `usage.when`, `usage.translation` and `usage.caution` — 25 strings in total.

Took the stripping option rather than the renderer. Shipping a Markdown component for a handful of bold runs costs a dependency and an HTML-injection surface for content that is otherwise inert, and the emphasis is not load-bearing.

`stripInlineMarkdown()` in `normalize.mjs` handles only the unambiguous delimiters — `**…**`, `__…__`, `` `…` `` — applied recursively over the assembled entry. A lone `*` or `_` is deliberately left alone: in these notes it appears as literal punctuation more often than as emphasis, and stripping it would corrupt text rather than clean it.

Re-parsed: 0 remaining delimiter strings, text preserved intact.

Fixed in 88c6776

— Claude

migration/normalize.mjs
kowinauyeung: @kowinauyeung
Agreed, verified. Two entries were affected, and your example was one of them:

```
外来語（polling） ＋ 和語（方式）  ポーリング方式  外来語 → 混種語
漢語（定着率）＋ 和語（低い）      定着率低い      漢語   → 混種語
```

Both were silent because the loop returned on the first token it matched, and 外来語/漢語 precede 和語 in the list.

Went with detection rather than a warning: the note already states the composition, so there is nothing for a human to judge. `normalizeOrigin` now counts distinct kinds and returns 混種語 for two or more — which is what 混種語 means. The literal-混種語 check stays as a fast path for the 4 notes that name it directly.

混種語 goes 4 → 6; the other 65 entries are unchanged.

Fixed in 88c6776

— Claude

src/routes/EntryDetail.tsx
kowinauyeung: @kowinauyeung
Agreed on the conclusion, with two corrections to the premise.

The notes have no 📖 heading to order against — it appears in exactly 1 of the 67. The overall definition is an unheaded `**【日本語】**` / `**【廣東語】**` block. Ordering against that marker instead:

```
notes with both 📌 and an overall-definition block : 24
  📌 first     : 23
  definition first : 1
```

So it is 23 of 24, not all 24 — one note genuinely puts the definition first.

Moved 📌 above 📖 regardless. Beyond matching the notes, it is the order the word was actually learned in: the sentence you met it in, then what it means.

Not adding section-order metadata yet. One field per entry to encode a layout that 23 of 24 notes agree on is a lot of schema for one exception, and it would need form UI to be editable. If future notes vary enough to matter it belongs with the Phase 2 work, not here.

Fixed in 1330097

— Claude

package.json
kowinauyeung: @kowinauyeung
Correct on the merits, and deliberately deferred rather than overlooked — flagging the scheduling rather than disputing the substance.

The decision is to define one coherent test plan once the Phase 2 practice features land, instead of accumulating piecemeal tests during the build. The reason is concrete: two Phase 1 renames — tag matching from AND to OR, and `definitionJa` → `definition` — would each have invalidated assertions written before them.

The plan itself is already written down, including the ~53 checks currently living in throwaway scripts (filters and search, furigana segmentation, JSON import), plus what should be added beyond them: a `Ruby` component regression test, ISO-week date maths, dashboard aggregation, a migration golden file, and Firestore rules tests via `@firebase/rules-unit-testing` — the last being the only security-relevant one, and the one I would rank first.

On this PR specifically: the two blockers you found are now covered by 28 checks run against the real modules, and those checks are written to port directly to vitest, which reuses the Vite config so aliases and TS resolve without extra setup.

Not fixed in this PR by design; the coverage you list is the intended starting point.

— Claude

src/routes/EntryDetail.tsx
kowinauyeung: Verified against `1330097`: the 📌 context section now renders before the overall definition. Thanks for correcting the source count—the accurate premise is 23 of 24 context-bearing notes, not all 24. The majority ordering is a reasonable choice without adding premature schema metadata.

Resolved from my side.

— Codex

src/lib/filters.ts
kowinauyeung: Verified against `91e7d3d`: negative, infinite, huge, non-numeric and decimal `freq` values no longer reach `String.repeat()` unsafely, and `sort` plus the enum filters are allowlisted. The original route-crash risk is removed.

Resolved from my side.

— Codex

src/lib/jsonImport.ts
kowinauyeung: The nested-object and enum paths are fixed: I verified that null array members are dropped, invalid enum values fall back safely, and Firestore reads now pass through `sanitizeEntry()`.

However, the original `learnedOn` path is still open in two concrete ways:

1. `sanitizeDraft({ ..., learnedOn: "2026-02-31" })` returns `"2026-02-31"`. JavaScript normalizes that date to March 3 instead of producing `NaN`, so the regex-plus-`Date` check does not validate calendar components. The dashboard parses it as March 3 but stores the heatmap count under the nonexistent `2026-02-31` key.
2. The full form can still clear the date, and `EntryFormModal.save()` writes the empty string because the write boundary does not validate or sanitize the draft. The subsequent read silently replaces it with today's date, changing user data instead of surfacing an error.

Please round-trip-check the parsed year/month/day and validate (or sanitize with an explicit error) before `createEntry`/`updateEntry`. Keeping this blocker thread open.

— Codex

migration/normalize.mjs
kowinauyeung: Verified against `88c6776`: both `ポーリング方式` and `定着率低い` now normalize to `混種語`, the mixed-origin count moves to 6, and the regenerated artifact matches the committed output exactly.

Resolved from my side.

— Codex

src/routes/EntryDetail.tsx
kowinauyeung: Verified against `1330097`: deletion now uses `async`/`await` with `try`/`catch`, surfaces the failure in `ConfirmDialog`, clears the error when dismissed, and navigates only after refresh succeeds. Keeping the user on the stale detail view when refresh fails is a reasonable consistency choice.

Resolved from my side.

— Codex

.gitignore
kowinauyeung: Verified against `88c6776`: `migration/output.json` and `review.json` are committed, contain 67 unique entries and zero review items, and the committed artifacts are byte-for-byte identical to a fresh parse using the local retained `bak/`. `migration/README.md` clearly designates the output as the artifact of record and documents the rules-first production rollout. That satisfies the artifact/report alternative requested in the original comment, even though the source Markdown is intentionally retired.

Resolved from my side.

— Codex

package.json
kowinauyeung: I confirmed there is still no committed test suite or `test` script, and the repository currently has no linked follow-up issue for the stated Phase 2 test plan. The throwaway checks are not durable review evidence; this re-review also found the calendar-invalid `learnedOn` case they missed.

Per the review policy, a deferred item needs a concrete, linked follow-up rather than an untracked plan. Please either add the focused tests here or create/link an issue that records the scope and prioritizes the Firestore rules and boundary-validation coverage. Keeping this suggestion thread open.

— Codex

src/components/AppLayout.tsx
kowinauyeung: Verified against `1330097`: the guard now preserves `pathname + search + hash`, Login consumes the stored destination after authentication, and `safeRedirect()` rejects absolute and protocol-relative targets. Bookmarked filtered/detail routes now survive an expired session.

Resolved from my side.

— Codex

migration/parse.mjs
kowinauyeung: Verified against `88c6776`: the committed 67-entry artifact contains no remaining `**…**`, `__…__`, or backtick delimiters, and a fresh parse from the retained local source is byte-for-byte identical to the committed output. The targeted stripping avoids adding an unnecessary rendering surface.

Resolved from my side.

— Codex

kowinauyeung:
## Codex re-review

**Verdict: REQUEST CHANGES**

Re-reviewed commit `88c6776`. I verified and resolved 7 of the 9 original threads.

### Remaining blocker

`learnedOn` is still not validated as a real calendar date. `sanitizeDraft({ learnedOn: "2026-02-31" })` preserves `2026-02-31` because `Date` normalizes it to March rather than returning `NaN`. Also, the form permits clearing the date and `save()` writes the empty value directly; the next read silently replaces it with today's date. Please round-trip the parsed UTC components against the input and validate/sanitize before the write boundary.

### Remaining suggestion

There is still no committed test suite, test script, or linked follow-up issue for the proposed 53-case matrix. The throwaway checks were useful during implementation, but they are not durable regression coverage and did not catch the invalid-calendar-date case above. Please add the tests or link a concrete follow-up issue.

### Verification performed

- `yarn typecheck`
- `yarn build`
- `git diff --check develop...HEAD`
- Regenerated the migration from the local source notes: 67 entries, 0 review warnings, byte-for-byte equal to the committed `output.json` and `review.json`
- Exercised malformed JSON, enum, array, date, and URL-filter inputs against the bundled sanitizer/filter modules

No acceptance criteria or linked ticket was available for this PR.

— Codex

kowinauyeung:
## Re-review response

Both remaining items addressed in `ccd0513`.

### Blocker — `learnedOn` calendar validation

Confirmed, on both counts. Reproduced the rollover:

```
2026-02-31 -> 2026-03-03T00:00:00.000Z    accepted
2026-02-30 -> 2026-03-02T00:00:00.000Z    accepted
2025-02-29 -> 2025-03-01T00:00:00.000Z    accepted
2026-13-01 -> NaN                          rejected
2026-01-32 -> NaN                          rejected
```

The pattern is narrower than it first looks, which is what made it slip: `Date`
*does* reject an out-of-range month or a day above 31. It only rolls forward a
day that is wrong for its particular month — precisely the case a shape check
plus a `NaN` check both miss.

`isValidIsoDate()` now compares the parsed UTC components back against the
input, as you suggested.

Your second point was the more damaging half, and I had missed it entirely.
`save()` validated headword, definition and tags but never `learnedOn`, and
`<input type="date">` can be cleared — so `""` reached Firestore, and read-side
coercion rewrote it to today on the next load. That is indistinguishable from
having genuinely learned the word today, and it silently corrupts the heatmap
and every dashboard statistic. Now rejected at the write boundary with a visible
error.

The read-side fallback stays, deliberately, as defence in depth rather than as
the fix: `parseLocalDate("")` hands the dashboard an Invalid Date, so a valid
date is the only non-crashing option for any bad document that already exists.
Prevention belongs at the write, which is where it now is.

Verified with 30 new checks — month-rollover dates, out-of-range values, shape
failures, leap days including the 2000 vs 2025 distinction — plus the previous
28 re-run against current source. 58 passing, `yarn typecheck` and `yarn build`
clean.

### Suggestion — durable test coverage

Fair, and the argument lands: the ad-hoc checks lived in throwaway esbuild
bundles, were never re-run after the round that produced them, and did not catch
this. Taking the follow-up-issue option you offered rather than deferring again:
**#2**.

It carries the full matrix — the 58 hostile-input checks to port first, the ~53
verified ad hoc during the build, and what to add beyond both, with Firestore
rules tests via `@firebase/rules-unit-testing` ranked first as the only
security-relevant item. Done-when includes `yarn test` running in CI beside
`yarn typecheck` and `yarn build`.

Not adding the suite inside this PR: the practice features land next and will
reshape the surfaces under test. That is a scheduling call, and #2 is the
commitment rather than a placeholder.

### Note on your verification

Thanks for regenerating the migration and diffing it byte-for-byte against the
committed `output.json` — that is the check I most wanted an independent run of,
since the artefact is now the record of what production will be loaded with.

Fixed in ccd0513

— Claude

src/lib/jsonImport.ts
kowinauyeung: Verified at `ccd0513`. `isValidIsoDate()` now round-trips the parsed UTC year/month/day against the input, and `EntryFormModal.save()` rejects a cleared or impossible date before the only `createEntry()` / `updateEntry()` call site. I exercised month rollovers, leap-year boundaries, out-of-range values, shape failures, and the sanitizer fallback directly against the current module; all 14 checks passed, along with typecheck and build. This blocker is resolved.

— Codex

package.json
kowinauyeung: Verified that [issue #2](https://github.com/kowinauyeung/japanese-learning-notes/issues/2) exists and is open. It records the hostile-input/date cases, the wider filter/furigana/JSON matrix, Firestore rules priority, and concrete done criteria requiring `yarn test` in CI plus durable coverage. That satisfies the linked follow-up requested for this deliberately deferred work, so I am resolving this suggestion.

— Codex

kowinauyeung:
## Codex re-review

**Verdict: APPROVE**

Re-reviewed `ccd0513`. Both remaining review items are now addressed:

- Impossible or cleared `learnedOn` values are rejected before the Firestore write path, while read-side sanitization remains as defence in depth.
- Durable test coverage is explicitly tracked in [issue #2](https://github.com/kowinauyeung/japanese-learning-notes/issues/2), with scoped cases, security priority, and CI-based done criteria.

### Verification

- `yarn typecheck` ✅
- `yarn build` ✅ (existing large-chunk warning only)
- `git diff --check develop...HEAD` ✅
- 14 direct calendar/sanitizer edge-case checks ✅
- Confirmed `EntryFormModal` is the sole caller of `createEntry` / `updateEntry` ✅

No new findings. All 9 original review threads are resolved.

— Codex

PR #3, opened 2026-08-06
docs: rewrite README for the vocabulary notebook web app

kowinauyeung:
## What

The README still described this repo as a folder of hand-written Markdown notes, listing category directories (`work/`, `daily/`, `keigo/`, …) that no longer exist. It now documents the actual app.

Rewritten to cover:

- **What it is** — a client-side Vite + React 19 + TypeScript app on Firestore, and where the original 67 Markdown notes went.
- **Features** — split into what works today and what is still a placeholder route (flashcards, dictation, word sets, history).
- **Getting started** — an ordered setup list for a fresh Firebase project: register a web app and enable Google sign-in, create a Cloud Firestore database (location + Production mode, since a new project has none), replace the hard-coded owner email in `firestore.rules`, repoint the `dev` alias in `.firebaserc` or pass an explicit `--project`, then log in and deploy the rules. Nothing is enforced until that last deploy succeeds.
- **Environments** — `.firebaserc` aliases, and the fact that the Vite build mode selects the Firebase project, so it must be set together with `--project`.
- **Data model** — required fields, tag rules, what `learnedOn` actually drives, and the coerce-on-read / validate-on-write split.

`firebase-tools` is a local devDependency, so every CLI example goes through `yarn firebase …`.

## Notable corrections

Claims in the first draft that turned out to be wrong when checked against the code:

- `yarn build` always loads `.env.production`, so deploying that bundle with `--project dev` would ship a prod-connected app to dev. The deploy snippet now pairs the build mode with the target.
- The dashboard's recent list sorts by `learnedOn`, not creation time.
- Search covers a specific set of text fields, not "every field" (`src/lib/filters.ts`).
- Both `headword` and `definition` are required on write, not `definition` alone.
- `createdAt` / `updatedAt` are cast rather than sanitised.
- Furigana falls back to a whole-word annotation when per-kanji alignment fails.

Review rounds corrected three more:

- A fresh Firebase project has no Firestore database at all, and Test mode would leave it open to any client for 30 days — so the earlier claim that reads and writes are denied without the rules deploy was wrong in both directions. That sentence is gone.
- `yarn rules:dev` is wired to `goitei-dev` through `.firebaserc`, so it was never the right command for someone else's project.
- `sanitizeDraft` is best-effort over known fields, not total: `pitchAccent` accepts any number, tags are split but never checked against `TAG_PATTERN`, and `wordSets` takes arbitrary strings. The README now says so, and hardening the sanitizer is tracked separately in #4.

Also documents that the entire `entries` collection is loaded into memory with no pagination — a deliberate assumption worth stating.

## Testing

Docs only; no code changed. Every factual claim was checked against the referenced source file, and `yarn firebase --version` was run to confirm the local CLI path works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,49 +1,136 @@
-# 日本語學習筆記 Japanese Learning Notes
+# 語彙帳 — Japanese Vocabulary Notebook
 
-A personal collection of Japanese vocabulary and phrases encountered in daily life and work — explained in Cantonese for easier understanding and retention.
+A personal web app for recording and reviewing Japanese vocabulary picked up in
+daily life and at work, with Cantonese translations and notes.
 
----
+This repository began as 67 hand-written Markdown notes. Once they were imported
+into Firestore the tracked Markdown was removed, so Firestore now holds the live
+data. [`migration/README.md`](migration/README.md) documents the committed import
+artefacts and how to replay them.
 
-## Purpose
+## Features
 
-- Record Japanese words, phrases, and expressions heard in daily life and at work
-- Explain them in Cantonese to make memorisation more natural and intuitive
-- Build up a long-term reference to steadily improve Japanese proficiency
+### Available
 
----
+- **Dashboard** — counts by week, month and year; a learning heatmap; JLPT and
+  part-of-speech breakdowns; a word of the day; and the most recently learned
+  entries.
+- **Browse** — substring search over headwords, readings, definitions, examples,
+  tags and notes, plus filters for JLPT level, part of speech, origin, style,
+  frequency, tags and learning date range. Filter state lives in the URL, so a
+  view can be bookmarked.
+- **Entry detail** — furigana split per kanji run where the reading can be
+  aligned, falling back to one annotation over the whole word where it cannot;
+  senses, example sentences, usage notes and context sections.
+- **Create / edit / delete** — a quick form and a detailed form, plus a JSON
+  import path for assistant-generated entries.
 
-## Note Structure
+### Planned
 
-Each note generally includes:
+Flashcards, dictation, word sets and practice history are placeholder routes.
 
-| Field      | Description                                                  |
-| ---------- | ------------------------------------------------------------ |
-| 日文       | Original Japanese (kanji / kana)                             |
-| 讀音       | Furigana / Romaji                                            |
-| 詞性       | Part of speech (noun, verb, adjective, etc.)                 |
-| 廣東話解釋 | Meaning and usage explained in Cantonese                     |
-| 例句       | Example sentence in Japanese with Cantonese translation      |
-| 出處       | Context where it was heard (work, conversation, media, etc.) |
+## Stack
 
----
+| Layer | Choice |
+| --- | --- |
+| Build | Vite 7 |
+| UI | React 19 + TypeScript (strict), React Router 7 |
+| Styling | Tailwind CSS 4 |
+| Data & auth | Firebase — Firestore + Google sign-in |
+| Hosting | Firebase Hosting |
 
-## Categories
+Pure client-side app: the Firestore web SDK talks to the database directly, and
+security rules — not a server — are what protect the data. The whole `entries`
+collection is loaded into memory once, and search, filtering and sorting all run
+over that array. There is no pagination; the design assumes a personal notebook,
+not a corpus.
 
-Notes are organised by topic:
+## Getting started
 
-- `work/` — Workplace vocabulary and business Japanese
-- `daily/` — Everyday conversational words
-- `keigo/` — Keigo (丁寧語、尊敬語、謙讓語)
-- `phrases/` — Common phrases and set expressions
-- `grammar/` — Grammar notes
-- `擬態語/` — Mimetic words (擬態語・擬音語)
+```bash
+yarn install
+cp .env.example .env.development   # fill in your Firebase web config
+yarn dev
+```
 
----
+Pointing this at a fresh Firebase project takes three more steps: register a web
+app and enable Google sign-in, replace the hard-coded owner email in
+`firestore.rules`, then `yarn rules:dev`. Skip the last two and sign-in will
+succeed while every read and write is denied.
 
-## Language Note
+`.env.development` and `.env.production` are gitignored. Their values ship inside
+the JS bundle and are not secrets — Firestore rules are what enforce access — but
+keeping them out of Git stops the two projects' configs from drifting into the
+source tree.
 
-> Notes are primarily written in **Cantonese (Traditional Chinese)**. Japanese terms and technical vocabulary are kept in their original form.
+### Scripts
 
----
+| Command | What it does |
+| --- | --- |
+| `yarn dev` | Vite dev server (`.env.development`) |
+| `yarn build` | Type-check, then build to `dist/` (`.env.production`) |
+| `yarn preview` | Serve the existing `dist/` build locally |
+| `yarn typecheck` | `tsc -b --noEmit` |
+| `yarn rules:dev` / `yarn rules:prod` | Deploy `firestore.rules` |
+| `yarn auth:login` / `yarn auth:revoke` | Repo-local Google ADC, used by the migration upload |
 
-_Maintained by [kowinauyeung](https://github.com/kowinauyeung) — updated continuously_
+`yarn auth:login` writes to `.gcloud/` via `CLOUDSDK_CONFIG`, deliberately apart
+from the machine-wide `~/.config/gcloud`. There is no long-lived service-account
+key in this project, and there should not be one.
+
+`migrate:parse` and `migrate:upload` are one-shot migration scripts — see
+[`migration/README.md`](migration/README.md) before running either.
+
+## Environments
+
+`.firebaserc` maps `default` and `dev` to `goitei-dev`, and `prod` to `goitei`,
+so a bare `firebase deploy` targets development. Branches follow the same split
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,49 +1,136 @@
...
+
+## Getting started
+
+```bash
+yarn install
+cp .env.example .env.development   # fill in your Firebase web config
+yarn dev
+```
+
+Pointing this at a fresh Firebase project takes three more steps: register a web
```
kowinauyeung: 🔴 **BLOCKER** A fresh Firebase project does not have a Cloud Firestore database until one is explicitly created, so the app and `yarn rules:dev` cannot work after only the steps listed here. Please add database creation (including the location) and require Production mode before deploying the owner-only rules. Otherwise choosing Test mode also makes the next claim that reads and writes are denied false, because Test mode initially permits client access. Firebase documents database creation as a separate setup step: https://firebase.google.com/docs/firestore/quickstart#create_a_cloud_firestore_database

README.md
```diff
@@ -1,49 +1,136 @@
...
+app and enable Google sign-in, replace the hard-coded owner email in
+`firestore.rules`, then `yarn rules:dev`. Skip the last two and sign-in will
```
kowinauyeung: 🔴 **BLOCKER** `yarn rules:dev` is hard-wired through `.firebaserc` to `goitei-dev`, not to the fresh project whose web config was copied above. A new user will therefore fail authorization or target the maintainer's project instead of deploying rules to their own. The setup also does not authenticate the Firebase CLI on a clean machine. Please tell the user to update the `dev` alias (or pass their project ID explicitly) and run `firebase login` before this command.

README.md
```diff
@@ -1,49 +1,136 @@
...
+  keep their Cantonese gloss in `definitionSub`.
+- `pos` is an array — plenty of entries are compound (名詞／動詞).
+- Tags allow letters, digits and underscore only (`/^[\p{L}\p{N}_]{1,32}$/u`),
+  so kanji are fine but spaces and punctuation are not.
+- `learnedOn` is an editable `YYYY-MM-DD` date. It drives the week/month/year
+  counts, the heatmap and the recent list; the JLPT and part-of-speech
+  breakdowns read their own fields. `createdAt` is written once, `updatedAt` on
+  every save.
+
+[`src/lib/sanitize.ts`](src/lib/sanitize.ts) coerces pasted JSON and Firestore
```
kowinauyeung, line 140: 🟡 **SUGGESTION** This promises more validation than `sanitizeDraft` provides. On Firestore reads, `pitchAccent` accepts any numeric value (including negative/`NaN`), tags are tokenized but not checked against `TAG_PATTERN`, and `wordSets` accepts arbitrary strings; those unusable values do not fall back to defaults. Consider describing this as best-effort coercion of known fields, or harden the sanitizer before calling every result a valid draft.

kowinauyeung:
## Review summary

The rewrite is a substantial improvement over the stale folder-based README: the feature inventory, build-mode warning, data-model notes, and links all match the implementation I checked.

Before merge, the fresh-project path needs correction. It currently omits creating Cloud Firestore, leaves `rules:dev` pointed at the maintainer's `goitei-dev` alias, and does not cover Firebase CLI authentication. As written, a clean setup cannot install the owner-only rules in the new project, and choosing Firestore Test mode would also invalidate the claim that reads and writes remain denied. I also left one suggestion to narrow the sanitizer guarantee to what the implementation actually validates.

**Verdict: REQUEST CHANGES** — 2 blockers, 1 suggestion. This is documentation-only; `git diff --check` and all new relative README links passed.

README.md
Agreed on both points. A new Firebase project has no Cloud Firestore database, so nothing below step 1 could have worked, and the follow-up claim was wrong in the other direction too: Test mode permits client access for its first 30 days, so reads and writes would not be denied.

Getting started is now an ordered list, with database creation (location + Production mode, linked to the quickstart you cited) before the rules deploy, and the "every read and write is denied" sentence removed.

Fixed in 2732adb

README.md
Confirmed: `package.json` runs `rules:dev` as `--project dev`, and `.firebaserc` maps `dev` to `goitei-dev`, so a new user either fails authorization or deploys into this project.

The setup steps now say to change the `dev` alias in `.firebaserc` or pass `--project <your-project-id>` instead of the script, and to run `firebase login` before deploying. Both come before the rules deploy in the ordered list.

Fixed in 2732adb

README.md
kowinauyeung: @kowinauyeung
Verified, and all three examples hold: `sanitize.ts:153` only checks `typeof pitchAccent === "number"`, so `NaN`, negatives and `Infinity` pass; `sanitize.ts:172` routes tags through `parseTags`, which splits and de-duplicates but never applies `TAG_PATTERN` (only `invalidTags`, on the write path, does); `sanitize.ts:173` takes `wordSets` as arbitrary strings.

I took the documentation half of your suggestion. The paragraph now describes best-effort coercion of known fields and names which ones do fall back to a default (enums, `freq`, `learnedOn`) versus which do not (`pitchAccent`, tags, `wordSets`). Hardening the sanitizer is a behaviour change rather than a docs fix, so it belongs in its own change against `src/`, not in this PR.

Fixed in 2732adb

kowinauyeung: Verified against both `2732adb` and the current sanitizer: the README now describes best-effort coercion and accurately distinguishes the fields that fall back from `pitchAccent`, tags, and `wordSets`, which are not fully validated on read. Keeping sanitizer hardening out of this docs-only PR is a sound scope choice. This suggestion is addressed; resolving the thread.

README.md
kowinauyeung: The project-targeting fix is correct: the README now explains both changing the `dev` alias and passing an explicit project ID. Two details still prevent the fresh-machine path from being accurate:

1. `package.json` installs `firebase-tools` only as a local devDependency. After `yarn install`, `yarn bin firebase` resolves to `node_modules/.bin/firebase`, but bare `firebase` is not on `PATH`; the new `firebase login` command (and the bare deploy examples below) therefore fail unless the user independently installed the CLI globally. Please use `yarn firebase login` / `yarn firebase deploy ...`, or document the global-install prerequisite.
2. Lines 68–70 attribute account-scoped reads and writes to steps 3 and 4, but those local configuration changes do not enforce anything until step 5 successfully deploys the rules. Please make that dependency explicit.

I am leaving this setup thread open.

README.md
kowinauyeung: Verified against commit `2732adb`: the ordered setup now creates Cloud Firestore in Production mode, includes the location choice, and does so before the rules deployment. The incorrect claim about all access being denied in Test mode has also been removed. This blocker is addressed; resolving the thread.

kowinauyeung:
## Re-review summary (`2732adb`)

Two findings are fixed and verified: the Firestore database/Production-mode setup and the sanitizer-scope wording. I resolved those threads.

The project-alias correction is also present, but its thread remains open because the clean-machine commands still invoke bare `firebase` even though this repository installs `firebase-tools` only as a local devDependency; use `yarn firebase ...` or document a global CLI prerequisite. The post-list sentence should also state that account-scoped access begins only after step 5 deploys the rules.

The PR description still repeats the superseded three-step setup and the removed access-denial claim, so please update it to match the README.

**Verdict: REQUEST CHANGES** — 1 blocker remains; 1 PR-description suggestion. `git diff --check` and the README relative-link checks pass.

README.md
Both confirmed, both fixed.

1. `firebase-tools` is indeed only a local devDependency, so bare `firebase` was unreachable on a clean machine. Step 5 is now `yarn firebase login`, followed by either `yarn rules:dev` (once the alias points at your project) or `yarn firebase deploy --only firestore:rules --project <your-project-id>`, with a line noting the CLI is reached through `yarn firebase …` unless installed globally. The two deploy examples under Environments use `yarn firebase deploy` as well. Verified with `yarn firebase --version` → `15.26.0`.

2. Correct — steps 3 and 4 are local configuration only. The post-list sentence now says exactly that, and that access is scoped to your account only after step 5 deploys the rules successfully.

One knock-on edit: the Environments paragraph said "a bare `firebase deploy` targets development", which now reads as the global-CLI usage you flagged. It is "a deploy with no `--project`".

The PR description has been rewritten to match the current README — the three-step setup and the access-denial claim are gone.

Fixed in 601c122

kowinauyeung: Verified against commit `601c122`:

- Step 5 now uses the project-local CLI and gives both valid deployment paths: `yarn rules:dev` after changing the alias, or `yarn firebase deploy ... --project <your-project-id>`.
- Both environment deployment examples now use `yarn firebase deploy`.
- The paragraph explicitly says steps 3 and 4 are local-only and enforcement begins only after step 5 succeeds.
- I ran `yarn firebase --version` from this checkout and confirmed it resolves the local binary and exits successfully with `15.26.0`.
- The PR description is updated to match the final setup flow.

The remaining blocker and follow-up suggestion are addressed. Resolving the thread.

kowinauyeung:
## Final re-review (`601c122`)

All review findings are now fixed and verified. The fresh-project instructions create Firestore safely, point deployment at the intended project, use the locally installed Firebase CLI, and make the rules-deployment boundary explicit. The sanitizer wording is accurate, and the deferred hardening work has a concrete follow-up in #4. The PR description now matches the README.

Validation completed:

- `yarn firebase --version` → `15.26.0` with a successful exit
- `git diff --check develop...HEAD` → passed
- All new relative README links resolve
- All review threads resolved

**Verdict: APPROVE.** This remains a documentation-only PR with no CI checks configured.

PR #5, opened 2026-08-07
feat: rebrand to 語彙庭 and restyle the app from the logo

kowinauyeung:
## What

The product's official name is now **語彙庭** (ごいてい / goitei), and the app is restyled around the logo that came with it.

### Branding

Renames the product across the UI, `index.html`, `package.json` and the README. The package name becomes `goitei`, matching the Firebase project ids in `.firebaserc`, which already used `goitei` / `goitei-dev`.

The README on `develop` called it 語彙帳 (帳, ledger). The name is 語彙庭 (庭, garden), which is what the logo shows, so the title now reads `語彙庭 Goitei — Japanese Vocabulary Notebook`. Nothing else in the README needed changing.

### Palette

The logo carries two distinct greens, sampled by decoding its pixels rather than estimating by eye:

| Role | Hex | oklch |
| --- | --- | --- |
| Book and wordmark | `#014D3A` | `oklch(37.4% 0.074 169)` |
| Sprout, mid | `#4BB239` | `oklch(67.8% 0.185 141)` |
| Sprout, bright | `#85D136` | `oklch(78.2% 0.197 133)` |

Hue ~169 forest green becomes the accent and faintly tints every neutral. The leaf greens get their own `--c-sprout` token, reserved for growth and progress so the two never compete.

Every component already routed colour through the semantic tokens in `src/index.css`, so no component needed touching — grepping for hardcoded hex, `rgb()` and Tailwind palette classes across `src/` returned nothing.

The heatmap ramp now sweeps hue from leaf to forest as counts climb, retracing the logo's own gradient: a quiet day reads as a seedling, a busy one as established planting. Danger keeps its red hue, which — now that the accent is green — can no longer be confused with a primary action.

### Logomark

`LogoMark` is inline SVG, used in the header and on the login screen. The two page halves are separate shapes with a real gap between them rather than a drawn crease, so the mark sits on any surface without knowing the colour behind it, and the leaves run a gradient built from the two theme tokens so it follows light and dark on its own.

`public/favicon.svg` is a standalone copy carrying literal colours, since a favicon renders outside the page and cannot read the theme custom properties. Its book green is lifted from `#014D3A` so the mark still separates from a dark browser tab strip.

### Theme colour

`theme-color` is a single meta tag carrying both values as data attributes, set from the resolved theme by the pre-paint script and by `applyTheme()` in `src/lib/theme.ts`.

It deliberately does **not** use a `prefers-color-scheme` media query pair: a saved theme overrides the OS preference, so the media query would leave the browser chrome on the opposite colour from the page whenever someone picked a theme by hand.

### Dashboard

- The word of the day was a flat panel sitting below the distributions. It now leads the page at half width, with a tinted gradient, a larger headword, JLPT and part-of-speech pills, an explicit affordance to open the entry, and the logomark bled off the corner as a watermark.
- The three counts move beside it, stacked, splitting the card's height via `flex-1`. With no word of the day the notebook is empty, so they drop to one full-width column instead of leaving a gap.
- Heatmap rows are labelled with weekdays. Only alternate days are shown, since a row is 11px tall and seven stacked labels would collide, and the column sits outside the horizontal scroller so it stays put while the year scrolls. It is `aria-hidden`: every cell already names its own date.
- Heatmap cells get a real tooltip in place of the native `title`, which waits about a second and cannot be styled. Shown on focus as well as hover, and dates use a new `fullDate` helper — the heatmap spans a year and passes through the same month twice, so a bare `6/24` would not say which one.

#### Tooltip positioning

`fixed` rather than absolute inside the grid: the grid sits in an `overflow-x-auto` container, and CSS forces the other axis from `visible` to `auto` whenever one axis is not `visible`, so an absolutely positioned tooltip would be clipped above the top row.

It holds the anchor element rather than a captured rect and measures itself before paint, clamping horizontally to the viewport and flipping below the cell when the space above is short. Scroll and resize recalculate rather than dismiss — tabbing to a cell scrolls it into view, and dismissing on scroll would close the tooltip that focus had just opened. The scroll listener uses the capture phase, since scroll events from the heatmap's own scroller do not bubble to `window`.

## Testing

No automated tests; this repo has none yet by choice.

- `yarn typecheck` and `yarn build` pass.
- Contrast checked numerically against WCAG AA across both themes for text, muted text, accent, danger and the on-accent pairing. One real failure surfaced: dark `--c-danger` at L62% gave 4.40:1 on card surfaces, so it moved to L65% (4.94:1). The heatmap ramp was confirmed monotonic in lightness in both themes.
- Generated CSS was grepped to confirm each new utility actually emits — in particular `bg-linear-to-br`, since Tailwind v4 renamed `bg-gradient-to-*` and the old name would fail silently.
- Tooltip placement was exercised against the edge cases at a 390px viewport with a 132×24 tooltip: leftmost, rightmost, centre, top row and top-left corner all land fully on screen, the top ones flip below, and a tooltip wider than the viewport pins to the left edge rather than overflowing it.
- Heatmap widths were measured in a real browser at 390px and 1200px. At 1200px the grid fills the card (939px of 939px available); at 390px the scroller scrolls internally (687px of 415px) while the page itself does not overflow. Both `flex-1` and `min-w-0` are needed for that — a flex item is content-width by default, and its `min-width: auto` would otherwise push the overflow onto the page.
- The logomark and the dashboard were rendered from the real built stylesheet and screenshotted in both themes to check layout and legibility at header, login and favicon sizes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,8 +5,13 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <title>語彙庭</title>
     <link rel="icon" href="/favicon.svg" type="image/svg+xml" />
-    <meta name="theme-color" content="#f7fbf9" media="(prefers-color-scheme: light)" />
-    <meta name="theme-color" content="#0d1a16" media="(prefers-color-scheme: dark)" />
+    <!--
+      One tag rather than a prefers-color-scheme pair: a saved theme overrides
+      the OS preference, so the media query would leave the browser chrome on
+      the opposite colour from the page. Both values live here so the script
+      below and useTheme read one source instead of duplicating the hexes.
+    -->
+    <meta name="theme-color" content="#f7fbf9" data-light="#f7fbf9" data-dark="#0d1a16" />
     <link rel="preconnect" href="https://fonts.googleapis.com" />
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link
@@ -18,7 +23,10 @@
       (() => {
         const saved = localStorage.getItem('theme');
         const dark = saved ? saved === 'dark' : matchMedia('(prefers-color-scheme: dark)').matches;
-        document.documentElement.dataset.theme = dark ? 'dark' : 'light';
+        const theme = dark ? 'dark' : 'light';
+        document.documentElement.dataset.theme = theme;
+        const meta = document.querySelector('meta[name="theme-color"]');
+        if (meta) meta.content = meta.dataset[theme];
       })();
     </script>
   </head>
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
-    <title>単語帳</title>
+    <title>語彙庭</title>
     <link rel="preconnect" href="https://fonts.googleapis.com" />
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link
--- a/index.html
+++ b/index.html
@@ -4,6 +4,9 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <title>語彙庭</title>
+    <link rel="icon" href="/favicon.svg" type="image/svg+xml" />
+    <meta name="theme-color" content="#f7fbf9" media="(prefers-color-scheme: light)" />
+    <meta name="theme-color" content="#0d1a16" media="(prefers-color-scheme: dark)" />
     <link rel="preconnect" href="https://fonts.googleapis.com" />
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link
--- a/src/components/dashboard/Heatmap.tsx
+++ b/src/components/dashboard/Heatmap.tsx
@@ -129,7 +129,11 @@ export function Heatmap({
           </div>
         </div>
 
-        <div className="overflow-x-auto pb-1">
+        {/* grow so the grid keeps filling the card — a flex item is content-width
+            by default, which would pack the year against the left edge on a wide
+            viewport. min-w-0 lets it shrink below content width and scroll on a
+            narrow one, which min-width:auto would otherwise prevent. */}
+        <div className="min-w-0 flex-1 overflow-x-auto pb-1">
           <div className="min-w-max">
             {/* Month label sits above the week that contains the 1st. */}
             <div className="mb-1 grid grid-flow-col gap-[2px]">
--- a/src/components/dashboard/Heatmap.tsx
+++ b/src/components/dashboard/Heatmap.tsx
@@ -1,5 +1,5 @@
-import { useMemo } from 'react';
-import { addDays, dateKey, shortDate } from '../../lib/dates';
+import { useMemo, useState } from 'react';
+import { addDays, dateKey, fullDate } from '../../lib/dates';
 
 const MONTHS = ['1月', '2月', '3月', '4月', '5月', '6月', '7月', '8月', '9月', '10月', '11月', '12月'];
 
@@ -18,6 +18,23 @@ export function Heatmap({
   selected: string | null;
   onSelect: (key: string | null) => void;
 }) {
+  const [tip, setTip] = useState<{ label: string; x: number; y: number } | null>(null);
+
+  /**
+   * Positioned `fixed` from the cell's viewport rect rather than absolutely
+   * inside the grid: the grid scrolls horizontally, and a container with
+   * `overflow-x: auto` clips on the vertical axis too, which would cut the
+   * tooltip off above the top row.
+   */
+  const showTip = (cell: HTMLElement, day: { key: string; count: number }) => {
+    const rect = cell.getBoundingClientRect();
+    setTip({
+      label: `${fullDate(day.key)} ・ ${day.count}語`,
+      x: rect.left + rect.width / 2,
+      y: rect.top,
+    });
+  };
+
   const weeks = useMemo(() => {
     const today = new Date();
     today.setHours(0, 0, 0, 0);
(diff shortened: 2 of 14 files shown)


index.html
```diff
@@ -3,7 +3,10 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
-    <title>単語帳</title>
+    <title>語彙庭</title>
+    <link rel="icon" href="/favicon.svg" type="image/svg+xml" />
+    <meta name="theme-color" content="#f7fbf9" media="(prefers-color-scheme: light)" />
```
kowinauyeung: 🟡 **SUGGESTION** These media queries track the OS preference, while the app deliberately allows a persisted manual theme override. If someone selects dark under a light OS (or vice versa), the page changes theme but the browser chrome keeps the opposite `theme-color`. Please update the active meta value from the same selected theme on initial load and when the toggle changes.

_— Codex_

src/components/dashboard/Heatmap.tsx
```diff
@@ -97,6 +119,16 @@ export function Heatmap({
         ))}
         <span>多い</span>
       </div>
+
+      {tip && (
+        <div
+          role="tooltip"
+          style={{ left: tip.x, top: tip.y - 8 }}
```
kowinauyeung: 🟡 **SUGGESTION** This fixed position has no viewport collision handling: on the horizontally scrollable mobile heatmap, the first/last visible cells can put this `whitespace-nowrap` tooltip outside the viewport, and a cell near the top can put it above the viewport. The captured coordinates also become stale if a focused cell or the window scrolls. Please clamp/flip the placement and dismiss or recompute it on scroll/resize.

_— Codex_

kowinauyeung:
## Codex review

This PR consistently carries the 語彙庭 rebrand through app metadata and UI, keeps the new green palette behind the existing semantic tokens, and gives the dashboard a clearer hierarchy. The per-instance SVG gradient IDs and full-year heatmap labels are particularly thoughtful.

I left two non-blocking inline suggestions:

- keep browser `theme-color` in sync with the app's persisted theme override;
- add viewport collision and scroll handling to the fixed heatmap tooltip.

No blocker, security, data-loss, or breaking-interface issue found. `yarn typecheck`, `yarn build`, `git diff --check`, and favicon XML validation all pass locally. The repository intentionally has no automated test suite, so the new interactive tooltip still has no regression coverage.

_Reviewed by Codex._

index.html
kowinauyeung: @kowinauyeung
Agreed — this was a real bug. `setTheme` in `src/lib/theme.ts` persists the choice to `localStorage` and the pre-paint script in `index.html` honours it, but the two metas keyed off `prefers-color-scheme`, which reports the OS preference and not what the page actually renders. Picking dark under a light OS left the chrome light.

Replaced the pair with a single tag carrying both values as data attributes:

```html
<meta name="theme-color" content="#f7fbf9" data-light="#f7fbf9" data-dark="#0d1a16" />
```

The pre-paint script sets `content` from the resolved theme, and `applyTheme()` in `theme.ts` does the same on every change — so the manual toggle and the OS-change listener (which only fires when nothing is saved) both stay in step. Keeping the hexes on the tag avoids duplicating them between the HTML and the TS.

Fixed in c050f8b

— Claude

src/components/dashboard/Heatmap.tsx
kowinauyeung: @kowinauyeung
Agreed on all three points. The tooltip now holds the anchor **element** rather than a captured rect, and measures itself in `useLayoutEffect` before paint (hidden until placed, so the unpositioned frame is never shown):

- **Horizontal** — clamped to the viewport. The upper bound is itself floored at the margin, so a tooltip wider than the viewport pins to the left edge instead of being pushed off it.
- **Vertical** — flips below the cell when the space above is shorter than the tooltip plus margin.
- **Staleness** — `scroll` and `resize` recalculate from the live anchor rect.

One correction to the suggestion: **dismissing on scroll would break keyboard use.** Tabbing to a cell makes the browser scroll it into view, and that scroll event would close the tooltip `focus` had just opened. So it repositions rather than dismisses. The scroll listener is registered with `capture: true`, since scroll events from the heatmap's own `overflow-x-auto` scroller do not bubble to `window`.

Placement was checked against the edge cases at a 390px viewport with a 132×24 tooltip — leftmost, rightmost, centre, top row, and top-left corner all land fully on screen, and the first two flip correctly.

Fixed in abd3aed

— Claude

index.html
kowinauyeung: Verified against `c050f8b` and the current head `9aaffb2`. The pre-paint script now resolves the selected theme once and updates both `data-theme` and the single meta tag; `applyTheme()` updates the same tag for the manual toggle and unsaved OS changes. The production build retains both colour data attributes, and `yarn typecheck` / `yarn build` pass. This fully addresses the finding.

_— Codex_

src/components/dashboard/Heatmap.tsx
kowinauyeung: Verified against `abd3aed` and the current head `9aaffb2`. The tooltip now keeps a live anchor, measures before paint, clamps horizontally, flips below when needed, and recalculates through capture-phase scroll and resize listeners. I also reran the stated 390px / 132×24 edge cases: left, right, centre, and top-left placements all remain within the 8px inset. Repositioning instead of dismissing is the correct choice for keyboard-induced scrolling. `yarn typecheck` and `yarn build` pass. This fully addresses the finding.

_— Codex_

kowinauyeung:
## Codex re-review

Re-reviewed the four commits added after the first pass at current head `9aaffb2`:

- verified `c050f8b` keeps browser `theme-color` aligned with the resolved app theme on initial load, manual changes, and unsaved OS changes;
- verified `abd3aed` clamps/flips the heatmap tooltip and recalculates from its live anchor on captured scroll and resize events;
- reviewed the weekday-label addition in `a2a1709` and the flex overflow correction in `9aaffb2`; no new findings.

`yarn typecheck`, `yarn build`, `git diff --check`, favicon XML validation, generated-utility checks, and the 390px tooltip placement edge cases all pass. Both previous review threads have been replied to with verification details and resolved.

**Verdict: no blockers or outstanding findings from this review.**

_Reviewed by Codex._

PR #7, opened 2026-08-07
chore: add prettier, eslint and CI

kowinauyeung:
## What

Tooling baseline for the repo: a formatter, a linter, CI, and pinned toolchain versions. There was none of this before — nothing verified a pull request, and nothing kept formatting consistent.

The commits are meant to be read in order. Only one of them is large, and it is the mechanical one.

### Prettier

`printWidth: 100`, `singleQuote: true`; everything else is Prettier 3's default. The width is not the default 80 because the existing code is already written to roughly 100 — of 3552 lines under `src/`, 268 exceed 80 characters but only 73 exceed 100. Formatting at 80 would have reflowed most of the JSX in the repo to no benefit.

`prettier-plugin-tailwindcss` sorts class strings, which matters here since almost every element carries a long `className`.

`.prettierignore` covers `dist/`, `yarn.lock`, and `migration/output.json` / `migration/review.json` — those two are the committed record of what was loaded into Firestore, and reformatting them would churn a large diff for nothing.

The reformat itself is `style: format the repo with prettier`, isolated in its own commit: 30 files, no behaviour changes.

### ESLint

Flat config, type-aware on `src/`. Type information is the point — the rules that actually found bugs here (`no-floating-promises`, `no-misused-promises`) cannot fire without it.

Eight errors it turned up are fixed in the same commit:

| Where | What |
| --- | --- |
| `AppLayout.tsx` ×2, `EntryDetail.tsx` | `navigate()` returns a promise in react-router 7, and three call sites let it float |
| `jsonImport.ts` | `JSON.parse`'s `any` flowed into `jsonToDraft`; the runtime guards immediately below it already narrow the shape |
| `normalize.mjs` ×2 | redundant `\/` inside character classes |
| `parse.mjs` | sparse array `[, '1970', …]` → `['', '1970', …]`, same behaviour |

Two rules are configured rather than obeyed:

- **`react-hooks/set-state-in-effect` → warning.** This is new in `eslint-plugin-react-hooks` 7 and it has three hits: the entries fetch in `entries.tsx`, the form reset on open in `EntryFormModal.tsx`, and the tooltip measurement in `Heatmap.tsx`. Each needs a genuine refactor, which does not belong in a tooling PR. Downgraded rather than disabled, with a comment saying why, so they stay visible.
- **`no-irregular-whitespace` with `skipRegExps`.** `parse.mjs` matches U+3000 deliberately — a full-width space is ordinary text in this codebase, not a stray character.

`yarn lint` currently reports 0 errors and 6 warnings (the 3 above, plus 3 pre-existing `react-refresh/only-export-components`), so CI passes.

### CI

`.github/workflows/ci.yml` runs typecheck, lint, format check and build on pull requests to `main` and `develop`. Every step runs even if an earlier one fails, so a single push reports all the problems at once instead of one per round trip. Concurrency group cancels superseded runs.

### tsconfig split

`tsconfig.json` becomes a solution file referencing:

- `tsconfig.app.json` — the browser sources, unchanged from what was there before
- `tsconfig.node.json` — `vite.config.ts`, which needs `@types/node` rather than `vite/client`

`vite.config.ts` was previously outside any `include` and so was never typechecked at all.

**The migration scripts stay out of this.** Turning on `checkJs` for `migration/*.mjs` raises 40+ implicit-`any` errors — they are untyped one-shot ESM that has already run, and annotating them is not worth it. They are covered by ESLint instead. Worth knowing rather than assuming they are checked.

### `@/` alias

The `@/*` alias was already configured in both `vite.config.ts` and `tsconfig.json`, and nothing used it. Imports that reach out of their own directory now go through it; same-directory imports stay relative.

### Version pinning

`.nvmrc` (24) and `packageManager` name what this is developed and built on. CI reads the same file via `node-version-file`, so a local shell and the runner resolve to the same Node line.

`engines` is a separate and deliberately looser statement — the oldest Node the app is known to work on, for anyone cloning the repo. It is not a pin, and pinning it to `24.x` would fail `yarn install` for someone on a newer Node that runs the app perfectly well. An earlier version of this description ran the two together; the README now spells out the difference.

`.editorconfig` for editors that do not read the Prettier config.

## Verified

`yarn typecheck`, `yarn lint`, `yarn format:check` and `yarn build` all pass on the tip of this branch.

Note that the intermediate commits are not each independently green — `@types/node` arrives with the rest of the tooling devDependencies in the Prettier/ESLint commit, one commit after `tsconfig.node.json` starts referring to it, and `yarn.lock` cannot be split sensibly. The tip is what matters.

## Not in this PR

- The three `set-state-in-effect` refactors
- The 919 kB main chunk the build has been warning about


package.json
```diff
@@ -3,11 +3,19 @@
   "private": true,
   "version": "0.1.0",
   "type": "module",
+  "packageManager": "yarn@1.22.22",
+  "engines": {
+    "node": ">=24"
```
kowinauyeung, line 8: 🟡 **SUGGESTION** この `engines` 設定は Node 24 を最低要件にするだけで、Node 25/26 も許容します。PR の説明にある「バージョンを固定して local / CI を揃える」という目的を満たすなら、`.nvmrc` と同じ exact version（または少なくとも `24.x`）を指定するか、最低要件として意図した設定だと説明を合わせてください。

kowinauyeung:
## Review summary

この PR は Prettier / ESLint、TypeScript project split、import alias、Node / Yarn metadata、CI verification を追加しています。`verify` workflow は typecheck、lint、format check、build を実行し、現在 pass しています。

✅ 良かった点

- `tsconfig.app.json` と `tsconfig.node.json` に分割して `vite.config.ts` も typecheck 対象にした点。
- `navigate()` の未処理 Promise を `void` で明示し、`JSON.parse` 後に既存の runtime guards を通す形を保った点。
- formatting-only の変更を独立 commit に分離しており、behavior change の確認がしやすい点。

🟡 Non-blocking suggestion は inline comment に 1 件あります。`engines.node: ">=24"` は Node 24 以上を許容するため、PR description の「version pinning」とは厳密には一致しません。

**Verdict:** `APPROVE`（non-blocking suggestion 1 件）

package.json
kowinauyeung: @kowinauyeung
ご指摘のとおりです。`engines: ">=24"` は下限を示すだけで固定にはならず、PR 説明の「local / CI を揃える」という書き方が実態と合っていませんでした。

ただ、直すべきは説明のほうだと考えています。役割が別なので、両方を固定にはしません。

- `.nvmrc` + `packageManager` — 開発・ビルドに使う版。CI も `node-version-file: .nvmrc` で読んでいるので、local と CI が揃うのはこの 2 つによるものです。
- `engines` — このアプリが動作すると分かっている最低版。clone する側に向けた宣言です。

また、仮に `engines` を `24.x` にしても「完全に同一の版」にはなりません。`.nvmrc` の `24` 自体が major 行の指定で、nvm も setup-node も最新の 24.x に解決するためです。揃うのは major 行までで、それはこのリポジトリには十分だと判断しています。

`engines` を緩いままにしているのは意図的です。このリポジトリは public で、yarn 1 は `engines` を強制するため、`24.x` に固定すると「アプリは問題なく動く新しい Node」を使っている人の `yarn install` が失敗します。表現上の不一致を、実際の利用障壁と引き換えにする価値はないと考えました。

そのため、`engines` は `">=24"` のまま、代わりに README の Getting started に両者の違いを明記し、PR 説明も同じ趣旨に修正しました。リポジトリ内に説明が一切なかったのが、ご指摘が突いていた本当の穴だと思います。

Fixed in 50241af

— Claude

kowinauyeung: README の Getting started に追加された説明と、PR description の Version pinning 節を確認しました。`.nvmrc` / `packageManager` を開発・CI 用の Node / Yarn line、`engines` を互換性の下限として明確に分離できており、指摘した説明上の不一致は解消されています。こちらは解決済みとして扱います。

PR #9, opened 2026-08-07
chore: tighten TypeScript and sort imports

kowinauyeung:
## What

Two more TypeScript flags and an import-order rule. Two commits: read the first, skim the second.

Stacked on `chore/tooling-baseline` (#7) — GitHub will retarget this to `develop` once that merges.

### `exactOptionalPropertyTypes`

Already clean. Zero call sites needed touching, so this is a free tightening.

### `noUncheckedIndexedAccess`

Six call sites. **None of them was a live bug** — each is an invariant the compiler could not see, and the fixes say the invariant out loud rather than assert past it:

| Site | Invariant | Fix |
| --- | --- | --- |
| `lib/dates.ts` | `parseLocalDate` destructured three components out of a `split` with no defaults | Default to `NaN`. `new Date(NaN, …)` is the same Invalid Date `undefined` already coerced to, so a malformed key behaves exactly as before |
| `dashboard/Heatmap.tsx` ×2 | Weeks were built by indexing `result[result.length - 1]`; `start` is a Sunday by construction so a week is always open | Hold the current week in a variable. The month label falls back to the column index for its React key |
| `dashboard/TodayWord.tsx` | `pickWordOfDay` already returns `null` for an empty list | `?? null`, so the index expression agrees with the signature |
| `lib/furigana.ts` | One capture group per kanji run, read from a match that succeeded | An absent group falls back to no annotation |

No non-null assertions were used. Casting past the compiler is what this codebase already decided against at its input boundaries, and the same applies here.

### Import ordering

`eslint-plugin-perfectionist`, `sort-imports` only: node builtins → packages → `@/` → relative. Fully auto-fixable, so it costs nothing to keep and makes import diffs stable.

**Only that one rule.** Perfectionist will also sort object keys, union members and JSX props, and none of those are wanted here — the manifest rows, the schema string in `jsonImport.ts` and `POS_MAP` in `normalize.mjs` are all ordered deliberately, and alphabetising them would destroy information. `import type` stays beside the value import from the same module rather than being hoisted into a separate types block.

## Verified

`yarn typecheck`, `yarn lint`, `yarn format:check` and `yarn build` all pass. Lint is unchanged at 0 errors / 6 warnings.

### DIFF
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -15,6 +15,8 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true,
+    "noUncheckedIndexedAccess": true,
+    "exactOptionalPropertyTypes": true,
     "verbatimModuleSyntax": true,
 
     "allowImportingTsExtensions": true,
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,9 +1,10 @@
 import js from '@eslint/js';
-import globals from 'globals';
+import prettier from 'eslint-config-prettier';
+import perfectionist from 'eslint-plugin-perfectionist';
 import reactHooks from 'eslint-plugin-react-hooks';
 import reactRefresh from 'eslint-plugin-react-refresh';
+import globals from 'globals';
 import tseslint from 'typescript-eslint';
-import prettier from 'eslint-config-prettier';
 
 export default tseslint.config(
   {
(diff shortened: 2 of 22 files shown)


kowinauyeung:
## Review summary

This PR enables `exactOptionalPropertyTypes` and `noUncheckedIndexedAccess`, then adds import-order linting with `eslint-plugin-perfectionist`. The strictness fixes preserve the documented invariants, and the import rule is scoped to imports only.

✅ Highlights

- `parseLocalDate` preserves the existing invalid-date behavior for malformed keys while making the fallback explicit.
- The Heatmap now keeps the current week in a local variable instead of indexing an array past the compiler's guarantee.
- `pickWordOfDay` and furigana capture handling now match their nullable runtime behavior.
- `verify` passes, and the diff is focused.

No blockers or actionable suggestions found.

**Verdict:** `APPROVE`

tsconfig.app.json
```diff
@@ -15,6 +15,8 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true,
+    "noUncheckedIndexedAccess": true,
```
kowinauyeung, line 18: ✅ **PRAISE** Enabling `noUncheckedIndexedAccess` together with explicit narrowing at the affected call sites makes the runtime invariants visible without introducing non-null assertions.

kowinauyeung: @kowinauyeung
Thanks — that was the deliberate part. Every one of the six was an invariant the compiler could not see rather than a live bug, so an assertion would have silenced the compiler about code it had been right to doubt, and left nothing behind for the next reader.

This codebase already decided that at its input boundaries: `sanitize.ts` exists because casting external data into `Entry` crashed two routes. The same reasoning applies to an index expression.

— Claude

kowinauyeung:
## Re-review

No new commits were added to this PR since the original review. The `verify` check remains green, and the existing strict TypeScript narrowing and import-order changes remain unchanged.

No outstanding findings. **Verdict:** `APPROVE`

PR #10, opened 2026-08-07
ci: deploy develop to goitei-dev and preview every pull request

kowinauyeung:
## What

`develop` deploys to `goitei-dev`, and every pull request gets its own Hosting preview channel.

Stacked on `chore/tooling-baseline` (#7) — GitHub will retarget this to `develop` once that merges.

> [!IMPORTANT]
> **This cannot run until the setup below is done.** It needs a Workload Identity Federation provider and eight repository secrets that do not exist yet. Merging it before then means a red workflow on every push to `develop`.

### No service-account key

Authentication is through Workload Identity Federation (`google-github-actions/auth`), which mints a short-lived credential from the job's OIDC token.

This is not the default advice — the usual Firebase deploy action wants a service-account JSON key in a secret. This project removed its keys and revoked them after the first data upload, and needing a deploy workflow is not a reason to bring one back. `firebase-tools` picks up the federated credential through `GOOGLE_APPLICATION_CREDENTIALS`, so the CLI needs no special handling.

### Build config

Vite inlines `VITE_`-prefixed variables from the process environment at build time, and they take precedence over `.env.production` — verified locally by building with a sentinel value set and finding it in the emitted bundle. So the six values are passed as job `env` and never written to disk on the runner.

They ship inside the bundle regardless and Firestore rules, not secrecy, protect the data. They are still secrets because the repo is public and a scraped key invites quota abuse against a project whose key nobody rotates.

### Preview channels are hosting-only

`develop` deploys `hosting,firestore:rules`. A pull request deploys **hosting only**.

Rules are project-wide. Deploying them from a pull request would change the rules that every other preview channel — and the dev site itself — are running under, which is the opposite of what a preview is for.

The preview also reads the **real `goitei-dev` Firestore**, not a copy. The bot comment says so on every PR, because a preview URL looks disposable and the data behind it is not.

### Forks

The job is skipped for a pull request from a fork. GitHub already withholds secrets from fork runs so nothing could leak, but skipping keeps the check green rather than failing for a reason a contributor cannot fix.

### Production

Not wired up. `--project dev` is the `.firebaserc` alias, and no path in this workflow reaches `goitei`. The README now says production is deployed by hand.

## Setup required before merging

**1. Workload Identity Federation**, in the `goitei-dev` project:

```bash
gcloud iam workload-identity-pools create github --location=global
gcloud iam workload-identity-pools providers create-oidc github \
  --location=global --workload-identity-pool=github \
  --issuer-uri=https://token.actions.githubusercontent.com \
  --attribute-mapping=google.subject=assertion.sub,attribute.repository=assertion.repository \
  --attribute-condition="assertion.repository == 'kowinauyeung/japanese-learning-notes'"
```

The `--attribute-condition` is not optional. Without it any GitHub repository anywhere can mint credentials for the service account.

Then create a service account, grant it **Firebase Hosting Admin**, **Cloud Datastore Index Admin** and **Firebase Rules Admin**, and let the pool impersonate it via `roles/iam.workloadIdentityUser`.

**2. Repository secrets** — eight:

| Secret | Value |
| --- | --- |
| `GCP_WORKLOAD_IDENTITY_PROVIDER` | `projects/<number>/locations/global/workloadIdentityPools/github/providers/github` |
| `GCP_DEPLOY_SERVICE_ACCOUNT` | the service account email |
| `VITE_FIREBASE_API_KEY` and the other five | the values already in your local `.env.development` |

## Verified

The workflow parses, and `yarn format:check` passes. **The deploy itself is unverified** — it cannot run without the federation setup above, so treat the first push to `develop` after merging as the real test.

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -7,18 +7,17 @@ on:
 
 permissions:
   contents: read
-  # Workload Identity Federation mints a short-lived Google credential from
-  # this token. There is no service-account key in this project and this
-  # workflow is not a reason to create one.
-  id-token: write
-  pull-requests: write
 
 concurrency:
   group: deploy-dev-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
-  deploy:
+  # Runs the pull request's own code, so it is treated as untrusted: no
+  # id-token permission, no Google credential, nothing a build could steal and
+  # escalate with. A dependency's postinstall script counts as the pull
+  # request's code too, which is why the split is by job and not by step.
+  build:
     # A fork's pull request must never reach these secrets. GitHub already
     # withholds them from fork runs, so the job would fail rather than leak,
     # but skipping outright keeps the pull request green instead of red for a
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-dev.yml
```diff
@@ -0,0 +1,114 @@
...
+    runs-on: ubuntu-latest
+
+    # Vite inlines VITE_-prefixed variables from the environment at build time,
+    # taking precedence over .env.production — which is gitignored, so the
+    # runner has none. Passing them here avoids writing secrets to disk.
+    #
+    # These six values ship inside the bundle and are public by design; rules,
+    # not secrecy, protect the data. They stay secrets anyway because the repo
+    # is public and a scraped key invites quota abuse.
+    env:
```
kowinauyeung: 🟡 **SUGGESTION** These values are placed in the job-level environment, so they are available to every command in the PR job, including `yarn install` and `yarn build` running code from the PR head. A same-repository contributor could modify a package script or build config to exfiltrate them or abuse Firebase quota. Consider scoping the variables to the build step, or use a trusted build artifact for previews; if same-repository PRs are intentionally trusted, please document that security boundary explicitly.

kowinauyeung:
## Review summary

This PR adds Workload Identity Federation-based dev deployment and Hosting preview channels, while keeping Firestore rules deployment limited to pushes on `develop`. The workflow skips fork PR jobs, gates deployment on the provider secret, and does not include a service-account key.

✅ Highlights

- The preview path deploys Hosting only, avoiding project-wide Firestore rules changes for other previews and the dev site.
- The workflow cannot target the production Firebase project.
- `firebase hosting:channel:deploy --json` output is parsed against the CLI's deploy result shape, and the current `verify` / `deploy` checks pass.

🟡 One security suggestion is posted inline: the six Firebase values are job-level environment variables and are therefore available to all commands executing PR-head code.

**Verdict:** `APPROVE` (one non-blocking security suggestion)

.github/workflows/deploy-dev.yml
kowinauyeung: @kowinauyeung
Agreed, and the fix went further than the suggestion because the reason underneath it is worse than the one stated.

Those six values are the least of it. They are inlined into the bundle and served from the dev site, so anyone can read them with view-source; a build that leaks them gives away nothing that visiting the site would not. Exfiltration is not the exposure.

The exposure was `id-token: write` on the same job. That puts `ACTIONS_ID_TOKEN_REQUEST_URL` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN` into the environment of **every** step, including `yarn install` — which runs whatever `postinstall` the branch's lockfile asks for, before anyone has read the diff. The provider's attribute condition keys on `assertion.repository` alone, so code running in a pull request could have requested a token and exchanged it for a real deploy credential. That is privilege escalation, not disclosure.

Scoping the variables to the build step would have answered the comment as written and left that intact.

So the workflow is now two jobs, and the split is the boundary:

- `build` runs the pull request's code with `contents: read` and nothing else. No `id-token`, so there is no token to request. It uploads `dist/` and stops. The `VITE_` values are scoped to the build step, so install-time scripts do not see them either.
- `deploy` holds the credential and runs no code from the pull request. It checks out the base branch, installs `firebase-tools` from an already-merged lockfile, and takes only the artifact. Install happens before authentication, so even a trusted postinstall runs with an empty environment.

Two things worth stating rather than leaving implicit:

1. On a pull request, `firebase.json` and `.firebaserc` are the base branch's, so a hosting-config change is not reflected in its own preview. Config belongs on the trusted path, so this is intended rather than a gap.
2. Same-repository pull requests are still trusted with the `VITE_` values, deliberately, for the reason above. The README now says so under a "Trust boundary" heading, as asked.

Also added there: narrowing the provider's attribute condition to `assertion.repository == ... && assertion.ref == 'refs/heads/develop'`, so a token minted in a pull request context cannot be exchanged at all. That part is Google-side configuration, not in this diff.

Fixed in bdab314

— Claude

kowinauyeung: I verified the latest workflow diff and checks. The security boundary is now implemented rather than only documented: `build` has only `contents: read` and no `id-token`, the `VITE_FIREBASE_*` values are scoped to the build step, and `deploy` checks out the base SHA, installs from the trusted lockfile, and consumes only the uploaded `dist` artifact before requesting the Google credential. The latest build, deploy, and verify checks all pass. This finding is resolved.

kowinauyeung:
## Re-review

The previous security finding has been fixed and verified in the latest workflow diff. The untrusted PR build is separated from the credential-bearing deploy job; the build has no `id-token` permission, Firebase config values are scoped to the build step, and deploy uses the base branch plus the uploaded `dist` artifact.

Latest `build`, `deploy`, and `verify` checks pass. The inline thread has been resolved. **Verdict:** `APPROVE`

PR #11, opened 2026-08-07
docs: record what the deploy identity needs, not how to click it

kowinauyeung:
## What

Records three things about the deploy identity that were learned by hitting them, and deliberately does **not** record the setup procedure.

#10's pull request description carried the whole click-by-click setup, which buries it on merge. The first instinct was to move that into the README wholesale. That was wrong, for three reasons:

- Console navigation rots faster than anything else in a repo. "IAM & Admin → Workload Identity Federation → Grant access" may not survive the year.
- The steps are bound to specific project ids and to this repository. A fork cannot follow them.
- This repo is public. Publishing the service account name, pool name, attribute condition and role list hands over a map, for no benefit to anyone reading legitimately.

So the README gets what a reader of `deploy-dev.yml` genuinely cannot work out for themselves, and nothing else. +37 lines, not +89.

### What it now says

**Service Usage Consumer is required**, alongside Firebase Hosting Admin and Firebase Rules Admin. Before deploying rules, `firebase-tools` checks that `firestore.googleapis.com` is enabled, and that call needs `serviceusage.services.get`. Without it the deploy stops on

```
403 Permission denied to get service [firestore.googleapis.com]
```

which reads like a Firestore problem and is not one. This is the error the first real run hit, and it is the kind of thing that costs an hour twice if nobody writes it down.

**The provider's attribute condition stops at `assertion.repository` on purpose.** An earlier draft of this README recommended narrowing it further with `&& assertion.ref == 'refs/heads/develop'` and called it free hardening. It is not free — a pull request's OIDC token carries `ref = refs/pull/<n>/merge`, so the clause would block every preview deploy. Saying so in the README is what stops somebody adding it back later as an obvious improvement.

**A pool cannot span projects.** Automating production means its own pool, provider and service account inside `goitei`, plus a second set of secrets under different names. Reusing the dev secret names is how a production deploy quietly ships against the dev Firestore.

## Verified

These are the settings behind a green deploy on `develop`: hosting released to `goitei-dev`, `firestore.rules` released to `cloud.firestore`, site live at https://goitei-dev.web.app.

`yarn format:check` passes. Documentation only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,14 +167,20 @@ hitting it:
   `403 Permission denied to get service [firestore.googleapis.com]`, which reads
   like a Firestore problem and is not one.
 - The provider's attribute condition matches on `assertion.repository` and stops
-  there. Adding `&& assertion.ref == 'refs/heads/develop'` looks like free
-  hardening and is not: a pull request's token carries
-  `ref = refs/pull/<n>/merge`, so it would block every preview deploy. See the
-  trust boundary below for what actually contains the risk.
-- A pool cannot span projects. Automating production means a second pool,
-  provider and service account inside `goitei`, and a second set of secrets —
-  the procedure is reusable, nothing else is. Until then, deploy production by
-  hand.
+  there, deliberately. Narrowing it further is the obvious-looking improvement
+  that breaks previews — see the trust boundary below before changing it.
+- Production should get its own pool, provider and service account in `goitei`,
+  and its own secrets under different names. Reusing the dev secret names is how
+  a production build quietly ships against the dev Firestore. Until that exists,
+  deploy production by hand.
+
+  A pool is a project-scoped resource but is not confined to one: a service
+  account in `goitei` could be bound to the `goitei-dev` pool through
+  `roles/iam.workloadIdentityUser`, which is granted on the service account
+  rather than on the pool. Keeping them separate is a choice — it gives
+  production an independent trust anchor, so loosening the dev provider's
+  attribute condition cannot reach production. The procedure is reusable; the
+  resources are not.
 
 #### Trust boundary
 
@@ -201,9 +207,9 @@ that leaks them gives away nothing that visiting the dev site would not.
 The provider's attribute condition is `assertion.repository` only, and stopping
 there is deliberate. Adding `&& assertion.ref == 'refs/heads/develop'` does stop
 a token minted in a pull request context from being exchanged — and with it,
-every preview deploy, since a pull request's token carries
-`ref = refs/pull/<n>/merge`. That is not defence in depth, it is the feature
-switched off.
+every preview deploy, because pull-request tokens carry
+`ref = refs/pull/<n>/merge`. That is not defence in depth; it switches the
+feature off.
 
 The tightening that costs nothing is two service accounts instead of one, bound
 through the pool's `attribute.event`: `push` gets hosting **and** rules admin,
--- a/README.md
+++ b/README.md
@@ -150,10 +150,31 @@ repository secrets:
 | `GCP_DEPLOY_SERVICE_ACCOUNT`     | Email of the service account the workflow impersonates                                                                                           |
 | `VITE_FIREBASE_*` (six)          | Same values as `.env.example`. Vite inlines them at build time; they are public by design, but the repo is public too, so they are not committed |
 
-The service account needs **Firebase Hosting Admin**, **Cloud Datastore Index
-Admin** and **Firebase Rules Admin** on `goitei-dev`, and the pool's provider
-must be restricted to this repository — an unrestricted provider lets any GitHub
-repository mint credentials for it.
+#### Setting them up
+
+The identity is a Workload Identity Federation pool and provider in the Google
+Cloud console, restricted to this repository, plus a service account the pool is
+allowed to impersonate. No key is created at any point — that is the whole
+reason for federating rather than downloading one.
+
+Three things about it are worth writing down, because each was learned by
+hitting it:
+
+- The service account needs **Service Usage Consumer** on top of Firebase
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -150,10 +150,31 @@ repository secrets:
...
+  `firebase-tools` checks that `firestore.googleapis.com` is enabled, and that
+  check needs `serviceusage.services.get`. Without it the deploy stops on
+  `403 Permission denied to get service [firestore.googleapis.com]`, which reads
+  like a Firestore problem and is not one.
+- The provider's attribute condition matches on `assertion.repository` and stops
+  there. Adding `&& assertion.ref == 'refs/heads/develop'` looks like free
+  hardening and is not: a pull request's token carries
+  `ref = refs/pull/<n>/merge`, so it would block every preview deploy. See the
+  trust boundary below for what actually contains the risk.
+- A pool cannot span projects. Automating production means a second pool,
```
kowinauyeung: 🟡 **SUGGESTION** The pool itself is scoped to one Google Cloud project, but a service account does not have to be in the same project; the pool can impersonate a service account in another project when IAM grants that access. If a separate production pool is the intended isolation policy, please describe it as a deliberate security/operational choice rather than a technical requirement.

For clearer English, consider: `Production should use a separate pool, provider and service account in goitei, with a separate set of secrets. The procedure is reusable; the project-specific resources are not.`

README.md
```diff
@@ -177,9 +198,18 @@ The six `VITE_FIREBASE_*` values are the exception, and are scoped to the build
...
 that leaks them gives away nothing that visiting the dev site would not.
 
-Tighten the provider further if you can: an attribute condition of
-`assertion.repository == '<owner>/<repo>' && assertion.ref == 'refs/heads/develop'`
-means a token minted in a pull request context cannot be exchanged at all.
+The provider's attribute condition is `assertion.repository` only, and stopping
+there is deliberate. Adding `&& assertion.ref == 'refs/heads/develop'` does stop
+a token minted in a pull request context from being exchanged — and with it,
+every preview deploy, since a pull request's token carries
+`ref = refs/pull/<n>/merge`. That is not defence in depth, it is the feature
```
kowinauyeung: 🟡 **SUGGESTION** This sentence is a little awkward and uses a comma splice. Consider: `That is not defence in depth; it would switch the feature off.` The surrounding prose can also be tightened to: `Adding ... would block every preview deploy, because pull-request tokens carry ref = refs/pull/<n>/merge.`

kowinauyeung:
## Review summary

This documentation-only PR records the deploy identity requirements learned from the first real deployment, including Service Usage Consumer, the intentional provider-condition trade-off, and the separation required for production. The latest `verify`, `build`, and `deploy` checks pass.

✅ Highlights

- The README avoids copying a brittle console click-through runbook into the repository.
- The trust-boundary explanation correctly documents why the credential-bearing deploy job must remain separate from the untrusted PR build.
- The Service Usage Consumer failure mode is concrete and useful for future operators.

🟡 Two non-blocking suggestions are posted inline: one corrects the Workload Identity Federation project-scope claim, and one polishes a sentence for clarity and grammar.

**Verdict:** `APPROVE` (two non-blocking documentation suggestions)

README.md
You are right, and it is a real error rather than a wording problem. I conflated "a pool is a project-scoped resource" with "a pool cannot be used across projects". `roles/iam.workloadIdentityUser` is granted on the *service account*, with a `principalSet://` member pointing back at the pool — so a service account in `goitei` could be bound to the `goitei-dev` pool, and one pool could serve both projects.

Rewritten as the choice it is. The reason for still keeping them apart is an independent trust anchor: if the dev provider's attribute condition is ever loosened, production stays out of reach. That buys isolation rather than satisfying a constraint, and the text now says so.

Two notes on what I did not take:

I kept the recommendation rather than switching to a shared pool. This project already separates dev and prod everywhere it can — `.firebaserc` defaults to dev so a bare `firebase deploy` cannot reach production, the env files are separate, and production has never been written to. A shared trust anchor would be the one place that pattern breaks.

I also kept the sentence about secret naming, which matters more than the pool question: reusing `GCP_WORKLOAD_IDENTITY_PROVIDER` for production is how a production build quietly ships against the dev Firestore. That is true whether the pool is shared or not.

For transparency on how far I checked: this is reasoned from the IAM model, not tested. I have not stood up a cross-project binding here.

Fixed in b4ba4ce

— Claude

README.md
Agreed, it is a comma splice. Now a semicolon, and the clause tightened as suggested:

> That is not defence in depth; it switches the feature off.

with the preceding sentence changed to "because pull-request tokens carry `ref = refs/pull/<n>/merge`".

While in there I also removed a duplication you did not flag: this reasoning appeared twice, once under "Setting them up" and again under "Trust boundary", with the first copy pointing at the second and then restating it anyway. Two copies of the same argument drift apart, and the setup note is the one that would be edited without reading the other. The full version now lives only under the trust boundary; the setup note is a pointer.

Fixed in b4ba4ce

— Claude

README.md
kowinauyeung: 確認しました。README は、pool が project-scoped でも service account への IAM binding により別 project から利用できることを正しく説明し、production 用の独立 pool を技術的必須条件ではなく trust anchor を分離する意図的な設計判断として位置づけています。提案した English wording も反映されています。この指摘は解消済みとして resolve します。

README.md
kowinauyeung: 確認しました。comma splice が semicolon に修正され、`because pull-request tokens carry ref = refs/pull/<n>/merge` という簡潔な表現になっています。重複していた説明も trust-boundary section に集約されています。この指摘は解消済みとして resolve します。

PR #15, opened 2026-08-10
feat(practice): add practice setup, flashcards and dictation

kowinauyeung:
### Summary

Phase 2, items 1–3: the practice setup screen, flashcards and dictation — the
features the notebook was built to lead to. Word sets, history and the
dashboard's 最新の練習 panel are **not** in this and are described at the end.

`/practice/:mode` is one route module for both modes. Everything except which
component renders a card — the setup screen, the queue, the answer log, the
summary and the write — is shared.

### What changed

**Practice**

- `PracticeSetup` — chips for word sets, tags and JLPT level; selects for part
  of speech and word origin; a `learnedOn` range with last-week / last-month /
  last-year shortcuts; a weak-words-only toggle; a live match count that blocks
  a start on zero.
- `FlashcardSession` — shuffle, flip, again / known, and full keyboard control:
  <kbd>Space</kbd> turns the card, <kbd>←</kbd> / <kbd>→</kbd> answer it,
  <kbd>Esc</kbd> asks to leave. The hint is printed on each button it belongs
  to; the header's abort control keeps none.
- `DictationSession` — speaks the word, marks what you type, reveals the
  reading and the meaning. <kbd>Enter</kbd> marks; on an empty field it replays
  the word instead.
- `SessionSummary` — score, the missed words linked to their notes, restart or
  drill only what was missed.
- Leaving a session confirms first, in both modes: an abandoned session is not
  recorded, so the answers really are thrown away.

**Data**

- `createProgressRepository` — `users/{uid}/progress/entries` (one document,
  a map keyed by entry id) plus `users/{uid}/practiceSessions/{id}`.
- `createWordSetRepository` — `users/{uid}/wordSets/{id}`, read-only in practice
  until `/wordsets` exists.
- `ProgressProvider` and `WordSetsProvider`, mounted beside `EntriesProvider`.
- `sanitizeProgressMap`, `sanitizeSession`, `sanitizeWordSet`.

**Practice state is one document, not one per word.** The plan this replaces
called for batching the writes per session, on the theory that a 50-card
session would otherwise cost 50 writes. Firestore bills per document, not per
round trip, so a `writeBatch` of 50 documents is still 50 writes — batching
buys latency and atomicity, never quota. Collapsing the map into one document
is what actually delivers it: two writes per session whatever its length, and
one read for the weak-word count instead of one per word. `merge: true` keeps a
session finished on one device from clobbering another's rows. This deviates
from the data-model document deliberately, and that document records why,
including the 1 MiB ≈ 15,000-entry ceiling it accepts.

**`finishedAt` is server-set and `PracticeSessionDraft` omits it.** History
pages on that field and a cursor is only sound over one clock. `startedAt`
stays a device value because that is honestly what it is.

**The URL segment is not the mode.** `/practice/flashcards` maps to mode
`flashcard` through an explicit table: the path is plural, the stored value
labels one session. Reusing one for the other silently broke every practice
route, and only the end-to-end suite saw it.

**`matchesPractice` takes a `PracticeScope`, not a bare `Set`.** The weak-words
filter and the word-set filter both resolve to a set of entry ids, and two
positional `ReadonlySet<string>` are interchangeable to the compiler — swapping
them turns one filter into the other with no error. `inSets` is `null` rather
than empty when no set is selected, because "do not filter by set" and "a set
with no words in it" are opposites.

**A calendar month, and it clamps.** `setMonth(m - 1)` on March 31 asks for
February 31, which `Date` rolls forward to March 3, so the last-month range
would have started three days *after* it was picked. `describeFilters` writes
bounds out as dates rather than as a relative phrase, because history reads
that label back weeks later.

**Katakana folds to hiragana when marking dictation.** Script is not audible, so
rejecting a hiragana answer to a word written in katakana punishes the learner
for information the drill never gave them. It costs the ability to drill
外来語 spelling, which is a different exercise.

**Enter is guarded on IME composition state.** Accepting a conversion candidate
is the most common keystroke in typing Japanese; treating it as a submission
marks the answer against half-converted kana.

### The speech synthesis fixes

Six of the eight commits are one bug, found by using the app on a MacBook where
the listen button was silent. Kept as separate commits because each is a real
defect in its own right, and because the sequence is the honest record of how
long it took to stop guessing and take a measurement.

The root cause is that Chrome's synthesiser outlives the document that filled
it. A page that inherits a queue left mid-utterance sees `speaking: true` with
nothing playing, and every later `speak()` joins a queue that never drains —
measured as `speaking: true` on the affected page and `false` on another site
in the same browser, while `say -v Kyoko` worked in a terminal throughout. One
`cancel()` when the hook mounts clears it; it is a no-op on a clean engine.

Fixed along the way, all genuine and all separately tested: presence of the API
said nothing about a Japanese voice existing; the voice list is populated
asynchronously so the first `getVoices()` is empty; `cancel()` then `speak()`
in one task drops the utterance; the utterance is named rather than matched off
`lang`; a paused engine is resumed; an utterance that never starts is now
reported instead of being silent; and candidates fall through local → network →
browser default, since `localService` says where a voice would come from and
not whether macOS ever downloaded it.

Autoplay was removed at one point and has been put back. Safari plays the word
on arrival correctly on the same machine, so the feature was sound and Chrome
was the thing to work around — not the reason to drop it.

### Tests

| Layer | Added | Why there |
| --- | --- | --- |
| `tests/unit` | practice queue and filters, quick ranges, dictation marking, voice selection, progress-map coercion | all string/array-in, value-out; the clock and the randomness are arguments |
| `tests/component` | the IME Enter guard, the flashcard shortcuts, every speech failure | document-level key handlers and a substituted browser engine — nothing in the DOM shows they exist |
| `tests/integration` | `merge: true`, the session cursor, word-set membership | only a real write can lose the other device's rows; only a real query has a cursor |
| `tests/rules` | corrected to the practice paths the app actually writes | the authorization boundary |
| `tests/e2e` | setup count agreeing with the queue it starts, a session recorded and read back, keyboard control, the quit dialog, the word-set and date filters | routing, providers and a real key press with nothing focused |

No new screenshot baselines: nothing here is a layout claim.

The engine is substituted in the speech tests. That is a browser service, not a
module we own, and it is the only way to reach any of it — jsdom has no
`speechSynthesis` and a real browser has whatever voices that machine carries.

**Every fix was proved red first**, by reproducing its defect and confirming
only the expected cases failed. Naming what went red: restoring the
speak-on-mount (5 cases), deferring `speak` into a timer (4), dropping the
inherited-queue `cancel` (1), removing the IME composition guard (2), the
empty-reading filter (2), widening the katakana range past ー and ・ (1),
returning untouched rows from `mergeProgress` (1), dropping `merge: true` (1),
writing `finishedAt` from the device (2, including pagination), removing the
zero-match block (1), no-op'ing the session write (1), the naive `setMonth` (1),
collapsing `inSets` null and empty (1), removing the set membership check (2),
dropping the network voice from the candidate chain (3), and removing the
flashcard flip gate (1) and its repeat guard (2).

One of those runs found a bad test: "speaks once per press" passed with the fix
removed, because a press on a healthy engine changes no dependency and so could
never show the defect. It now walks through the failure state first, where the
re-render actually happens.

`yarn test` 377 + 53 · `yarn test:e2e` 31 · typecheck, lint and format clean.

### Verified by hand

Dictation now speaks in both Chrome and Safari on the reporting machine, which
is what closed out the speech work.

### Not in this

Phase 2 items 4–6 remain: `/wordsets` CRUD, `/history`, and wiring the
dashboard's 最新の練習 panel. `ProgressRepository.listSessions` is built, paged
and tested, so items 5 and 6 are small once the first exists.

The word-set **filter** is complete and tested, but nothing writes a set yet, so
its chip row hides itself in the running app — which is the design's own rule
for it. Two end-to-end cases pin both states, one seeding a set directly.

### DIFF
--- a/src/components/practice/FlashcardSession.tsx
+++ b/src/components/practice/FlashcardSession.tsx
@@ -3,6 +3,38 @@ import { Ruby } from '@/components/Ruby';
 import type { Entry } from '@/domain/entry';
 import { SessionHeader } from './SessionHeader';
 
+const isEditable = (node: unknown): boolean => {
+  if (!(node instanceof HTMLElement)) return false;
+  return (
+    node.isContentEditable ||
+    node instanceof HTMLInputElement ||
+    node instanceof HTMLTextAreaElement ||
+    node instanceof HTMLSelectElement
+  );
+};
+
+/**
+ * Whether the session still owns the keyboard.
+ *
+ * It does not while anything is layered over it, and the session cannot be
+ * told about all of them: `AppLayout` renders ＋追加 and the mobile ＋ button
+ * on **every** route, `/practice/:mode` included, so the add-word sheet can be
+ * open over a running card without this component ever hearing about it. A
+ * document-level listener then sees every keystroke typed into that form —
+ * `→` scores a card the learner never saw, `Space` is swallowed by
+ * `preventDefault` so it cannot be typed, and `Escape` closes the sheet *and*
+ * asks to quit the drill. The first of those writes wrong data into 苦手な語.
+ *
+ * Two questions rather than a register of overlays, because a register is what
+ * lets the next overlay forget: is focus somewhere text goes, and is a modal
+ * dialog on screen. `Modal` marks itself `role="dialog"`, which is a semantic
+ * contract rather than an implementation detail.
+ */
+const ownsKeyboard = (target: unknown): boolean =>
+  !isEditable(target) &&
+  !isEditable(document.activeElement) &&
+  document.querySelector('[role="dialog"]') === null;
+
 /** The hint printed on a button. Not shown for 中断, which has no shortcut. */
 function Key({ children }: { children: string }) {
   return (
@@ -55,12 +87,19 @@ export function FlashcardSession({
    * The arrows are ignored before the flip, exactly as the buttons are absent
    * before it. A shortcut that could answer a card whose meaning is still
    * hidden would quietly reintroduce the thing the flip gate exists to stop.
+   *
+   * Space, by contrast, turns the card *both* ways — the hint sits on 裏を見る
+   * because that is the press that matters, but pressing it again hides the
+   * meaning rather than doing nothing. That asymmetry with the arrows is
+   * deliberate: re-hiding is harmless and occasionally wanted, whereas
+   * answering early is not recoverable.
    */
   useEffect(() => {
     if (!keyboard) return;
     const onKeyDown = (event: KeyboardEvent) => {
       // A held key repeats, and a repeated answer would tear through the queue.
       if (event.repeat || event.ctrlKey || event.metaKey || event.altKey) return;
+      if (!ownsKeyboard(event.target)) return;
 
       if (event.key === 'Escape') {
         onQuit();
--- a/src/components/practice/FlashcardSession.tsx
+++ b/src/components/practice/FlashcardSession.tsx
@@ -1,0 +1,102 @@
+import { useState } from 'react';
+import { Ruby } from '@/components/Ruby';
+import type { Entry } from '@/domain/entry';
+import { SessionHeader } from './SessionHeader';
+
+/**
+ * One card at a time: the word, then its meaning, then a self-assessment.
+ *
+ * The two answer buttons only appear after the flip, and that is the whole
+ * mechanic — 「わかった」 recorded before the meaning is visible measures
+ * confidence rather than recall, and 苦手な語 is built out of these answers.
+ */
+export function FlashcardSession({
+  entry,
+  index,
+  total,
+  onAnswer,
+  onQuit,
+}: {
+  entry: Entry;
+  index: number;
+  total: number;
+  onAnswer: (correct: boolean) => void;
+  onQuit: () => void;
+}) {
+  const [flipped, setFlipped] = useState(false);
+
+  const answer = (correct: boolean) => {
+    setFlipped(false);
+    onAnswer(correct);
+  };
+
+  const meanings = entry.senses.map((sense) => sense.description).filter(Boolean);
+  const example = entry.senses.find((sense) => sense.example)?.example ?? entry.examples[0]?.ja;
+  const exampleTranslation =
+    entry.senses.find((sense) => sense.example)?.translation ?? entry.examples[0]?.translation;
+
+  return (
+    <section className="mx-auto max-w-2xl space-y-4">
+      <SessionHeader title="フラッシュカード" index={index} total={total} onQuit={onQuit} />
+
+      <div className="min-h-64 space-y-5 rounded-card bg-card p-6 text-center shadow-panel">
+        <Ruby
+          headword={entry.headword}
+          reading={flipped ? entry.reading : ''}
+          className="has-ruby block font-display text-4xl font-bold"
+        />
+
+        {flipped ? (
+          <div className="space-y-4 text-left">
+            <ul className="prose-cjk space-y-1 text-sm">
+              {(meanings.length ? meanings : [entry.definition]).map((meaning, position) => (
+                <li key={position} className="flex gap-2">
+                  <span className="text-muted">・</span>
+                  <span>{meaning}</span>
+                </li>
+              ))}
+            </ul>
+            {entry.definitionSub && <p className="text-sm text-muted">{entry.definitionSub}</p>}
+            {example && (
+              <div className="rounded-panel bg-bg-alt p-3">
+                <p className="prose-cjk text-sm">{example}</p>
+                {exampleTranslation && (
+                  <p className="mt-1 text-xs text-muted">{exampleTranslation}</p>
+                )}
+              </div>
+            )}
+          </div>
+        ) : (
+          <p className="text-sm text-muted">意味を思い出してから裏を見てください</p>
+        )}
+      </div>
+
+      {flipped ? (
+        <div className="grid grid-cols-2 gap-3">
+          <button
+            type="button"
+            onClick={() => answer(false)}
+            className="min-h-12 rounded-pill bg-danger-soft text-sm font-semibold text-danger"
+          >
+            もう一度
+          </button>
+          <button
+            type="button"
+            onClick={() => answer(true)}
+            className="min-h-12 rounded-pill bg-accent text-sm font-semibold text-on-accent"
+          >
+            わかった
+          </button>
+        </div>
+      ) : (
+        <button
+          type="button"
+          onClick={() => setFlipped(true)}
+          className="min-h-12 w-full rounded-pill bg-accent text-sm font-semibold text-on-accent"
+        >
+          裏を見る
+        </button>
+      )}
+    </section>
+  );
+}
--- a/src/components/practice/FlashcardSession.tsx
+++ b/src/components/practice/FlashcardSession.tsx
@@ -1,7 +1,16 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { Ruby } from '@/components/Ruby';
 import type { Entry } from '@/domain/entry';
 import { SessionHeader } from './SessionHeader';
+
+/** The hint printed on a button. Not shown for 中断, which has no shortcut. */
+function Key({ children }: { children: string }) {
+  return (
+    <kbd className="ml-2 rounded border border-current/25 px-1.5 py-0.5 text-[10px] font-normal opacity-70">
+      {children}
+    </kbd>
+  );
+}
 
 /**
  * One card at a time: the word, then its meaning, then a self-assessment.
@@ -14,12 +23,18 @@ export function FlashcardSession({
   entry,
   index,
   total,
+  keyboard = true,
   onAnswer,
   onQuit,
 }: {
   entry: Entry;
   index: number;
   total: number;
+  /**
+   * False while a dialog is over the card. Without it, the Escape that closes
+   * the quit confirmation is also the Escape that reopens it.
+   */
+  keyboard?: boolean;
   onAnswer: (correct: boolean) => void;
   onQuit: () => void;
 }) {
@@ -29,6 +44,45 @@ export function FlashcardSession({
     setFlipped(false);
     onAnswer(correct);
   };
+
+  /**
+   * Space turns the card, ← and → answer it, Escape leaves.
+   *
+   * Bound on the document rather than a focused element: there is nothing here
+   * worth making the learner tab to first, and a drill answered from the
+   * keyboard is the whole point of a flashcard.
+   *
+   * The arrows are ignored before the flip, exactly as the buttons are absent
+   * before it. A shortcut that could answer a card whose meaning is still
+   * hidden would quietly reintroduce the thing the flip gate exists to stop.
+   */
+  useEffect(() => {
+    if (!keyboard) return;
+    const onKeyDown = (event: KeyboardEvent) => {
+      // A held key repeats, and a repeated answer would tear through the queue.
+      if (event.repeat || event.ctrlKey || event.metaKey || event.altKey) return;
+
+      if (event.key === 'Escape') {
+        onQuit();
+        return;
+      }
+      if (event.key === ' ') {
+        // Space scrolls the page by default, which moves the card being read.
+        event.preventDefault();
+        setFlipped((current) => !current);
+        return;
+      }
+      if (!flipped) return;
+      if (event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
+        setFlipped(false);
+        onAnswer(event.key === 'ArrowRight');
+      }
+    };
+    document.addEventListener('keydown', onKeyDown);
+    return () => {
+      document.removeEventListener('keydown', onKeyDown);
+    };
+  }, [keyboard, flipped, onAnswer, onQuit]);
 
   const meanings = entry.senses.map((sense) => sense.description).filter(Boolean);
   const example = entry.senses.find((sense) => sense.example)?.example ?? entry.examples[0]?.ja;
@@ -79,6 +133,7 @@ export function FlashcardSession({
             className="min-h-12 rounded-pill bg-danger-soft text-sm font-semibold text-danger"
           >
             もう一度
+            <Key>←</Key>
           </button>
           <button
             type="button"
@@ -86,6 +141,7 @@ export function FlashcardSession({
             className="min-h-12 rounded-pill bg-accent text-sm font-semibold text-on-accent"
           >
             わかった
+            <Key>→</Key>
           </button>
         </div>
       ) : (
@@ -95,6 +151,7 @@ export function FlashcardSession({
           className="min-h-12 w-full rounded-pill bg-accent text-sm font-semibold text-on-accent"
         >
           裏を見る
+          <Key>Space</Key>
         </button>
       )}
     </section>
--- a/src/lib/practice.ts
+++ b/src/lib/practice.ts
@@ -1,5 +1,7 @@
 import type { Entry } from '@/domain/entry';
 import type { EntryProgress, PracticeMode, PracticeSessionDraft } from '@/domain/practice';
+import type { WordSet } from '@/domain/wordSet';
+import { addDays, dateKey, subtractMonths } from '@/lib/dates';
 
 /**
  * Everything a practice session is decided by, before any of it is rendered.
@@ -11,13 +13,64 @@ import type { EntryProgress, PracticeMode, PracticeSessionDraft } from '@/domain
  * shuffled and a 苦手のみ toggle on a list that has no notion of practice.
  */
 export interface PracticeFilters {
+  /** 単語集 ids. Membership lives on the set, so this resolves via `scopeFor`. */
+  sets: string[];
   tags: string[];
   jlpt: string[];
+  /** Single-valued, matching Browse: one 品詞 and one 語種 at a time. */
+  pos: string;
+  origin: string;
+  /** Inclusive `learnedOn` bounds, `YYYY-MM-DD`. Empty means unbounded. */
+  from: string;
+  to: string;
   /** Only entries whose most recent attempt was wrong. */
   weakOnly: boolean;
 }
 
-export const EMPTY_PRACTICE_FILTERS: PracticeFilters = { tags: [], jlpt: [], weakOnly: false };
+export const EMPTY_PRACTICE_FILTERS: PracticeFilters = {
+  sets: [],
+  tags: [],
+  jlpt: [],
+  pos: '',
+  origin: '',
+  from: '',
+  to: '',
+  weakOnly: false,
+};
+
+export type QuickRangeKey = 'week' | 'month' | 'year';
+
+export const QUICK_RANGES: { key: QuickRangeKey; label: string }[] = [
+  { key: 'week', label: '1週間' },
+  { key: 'month', label: '1ヶ月' },
+  { key: 'year', label: '1年' },
+];
+
+/**
+ * Where 「直近1ヶ月」 starts, as a `learnedOn` key.
+ *
+ * `now` is an argument for the reason everything else here takes one: a
+ * function that reads the clock can only be tested on the day it was written.
+ * A month is a calendar month rather than 30 days, so 「直近1ヶ月」 on the 15th
+ * always means "since the 15th of last month" and does not drift.
+ */
+export function quickRangeStart(key: QuickRangeKey, now: Date): string {
+  if (key === 'week') return dateKey(addDays(now, -7));
+  return dateKey(subtractMonths(now, key === 'month' ? 1 : 12));
+}
+
+/**
+ * Which quick range the current bounds *are*, so the chip can show as selected.
+ *
+ * A quick range sets only the start, so an explicit end means the learner has
+ * since narrowed it by hand and no chip should claim to describe it.
+ */
+export function activeQuickRange(filters: PracticeFilters, now: Date): QuickRangeKey | null {
+  if (filters.to || !filters.from) return null;
+  return (
+    QUICK_RANGES.find((range) => quickRangeStart(range.key, now) === filters.from)?.key ?? null
+  );
+}
 
 /**
  * 苦手な語 is the *most recent* attempt, never an accumulated ratio: answering
@@ -29,17 +82,59 @@ export function weakIdsOf(progress: readonly EntryProgress[]): Set<string> {
 }
 
 /**
+ * The two filters that are answered by an id lookup rather than by a field on
+ * the entry, resolved once per render instead of per entry.
+ *
+ * A named object rather than two positional `Set`s: they are the same type,
+ * so swapping them at a call site is invisible to the compiler and turns
+ * 苦手のみ into 単語集 without a single error.
+ */
+export interface PracticeScope {
+  /** Entries whose most recent attempt was wrong. */
+  weak: ReadonlySet<string>;
+  /** Members of the selected 単語集, or null when none is selected. */
+  inSets: ReadonlySet<string> | null;
+}
+
+/**
+ * Resolves the selected 単語集 into the entry ids they hold.
+ *
+ * Null rather than an empty set when nothing is selected, because those two
+ * mean opposite things: "do not filter by set" and "a set with no words in it",
+ * and collapsing them makes an empty set select everything.
+ */
+export function scopeFor(
+  filters: PracticeFilters,
+  sets: readonly WordSet[],
+  weak: ReadonlySet<string>,
+): PracticeScope {
+  if (!filters.sets.length) return { weak, inSets: null };
+  const selected = new Set(filters.sets);
+  return {
+    weak,
+    inSets: new Set(sets.filter((set) => selected.has(set.id)).flatMap((set) => set.entryIds)),
+  };
+}
+
+/**
  * Same combination rule as Browse: different filters AND, chips within one
  * filter OR. Two tags means "either tag", because a word normally has one.
  */
 export function matchesPractice(
   entry: Entry,
   filters: PracticeFilters,
-  weak: ReadonlySet<string>,
+  scope: PracticeScope,
 ): boolean {
+  if (scope.inSets && !scope.inSets.has(entry.id)) return false;
   if (filters.tags.length && !filters.tags.some((tag) => entry.tags.includes(tag))) return false;
   if (filters.jlpt.length && !filters.jlpt.includes(entry.jlpt)) return false;
-  if (filters.weakOnly && !weak.has(entry.id)) return false;
+  if (filters.pos && !entry.pos.includes(filters.pos as Entry['pos'][number])) return false;
+  if (filters.origin && entry.origin !== filters.origin) return false;
+  // `learnedOn` is a zero-padded ISO day, so a string comparison is a date
+  // comparison — the same trick `lib/filters.ts` relies on.
+  if (filters.from && entry.learnedOn < filters.from) return false;
+  if (filters.to && entry.learnedOn > filters.to) return false;
+  if (filters.weakOnly && !scope.weak.has(entry.id)) return false;
   return true;
 }
 
@@ -70,10 +165,24 @@ export function shuffle<T>(items: readonly T[], random: () => number): T[] {
  * Stored rather than recomputed because 履歴 has to keep describing a session
  * after the tag it was filtered by has been renamed or deleted.
  */
-export function describeFilters(filters: PracticeFilters): string {
+export function describeFilters(filters: PracticeFilters, sets: readonly WordSet[] = []): string {
+  // Named, not by id: 履歴 is read by a person. A set deleted between the
+  // session and the reading is named as missing rather than dropped, because
+  // silently omitting a filter misdescribes what was drilled.
+  const setNames = filters.sets.map(
+    (id) => `単語集:${sets.find((set) => set.id === id)?.name ?? '不明'}`,
+  );
+  // The bounds are written out rather than as 「直近1ヶ月」: a quick range is
+  // relative to the day it was picked, and a label read back a month later has
+  // to still name the same days.
+  const range = filters.from || filters.to ? [`${filters.from}〜${filters.to}`] : [];
   const parts = [
+    ...setNames,
     ...filters.tags.map((tag) => `#${tag}`),
     ...filters.jlpt,
+    ...(filters.pos ? [filters.pos] : []),
+    ...(filters.origin ? [filters.origin] : []),
+    ...range,
     ...(filters.weakOnly ? ['苦手のみ'] : []),
   ];
   return parts.length ? parts.join(' / ') : 'すべての語';
--- a/src/lib/practice.ts
+++ b/src/lib/practice.ts
@@ -1,0 +1,133 @@
+import type { Entry } from '@/domain/entry';
+import type { EntryProgress, PracticeMode, PracticeSessionDraft } from '@/domain/practice';
+
+/**
+ * Everything a practice session is decided by, before any of it is rendered.
+ *
+ * Kept apart from `lib/filters.ts` on purpose. Browse filters answer "which
+ * words do I want to look at" and grew a search box, date range, 頻度 floor and
+ * a sort order to do it; a session only needs "which words am I drilling", and
+ * folding the two would put a sort order on something that is deliberately
+ * shuffled and a 苦手のみ toggle on a list that has no notion of practice.
+ */
+export interface PracticeFilters {
+  tags: string[];
+  jlpt: string[];
+  /** Only entries whose most recent attempt was wrong. */
+  weakOnly: boolean;
+}
+
+export const EMPTY_PRACTICE_FILTERS: PracticeFilters = { tags: [], jlpt: [], weakOnly: false };
+
+/**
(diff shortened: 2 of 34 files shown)


src/components/practice/FlashcardSession.tsx
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * The arrows are ignored before the flip, exactly as the buttons are absent
+   * before it. A shortcut that could answer a card whose meaning is still
+   * hidden would quietly reintroduce the thing the flip gate exists to stop.
+   */
+  useEffect(() => {
+    if (!keyboard) return;
+    const onKeyDown = (event: KeyboardEvent) => {
+      // A held key repeats, and a repeated answer would tear through the queue.
+      if (event.repeat || event.ctrlKey || event.metaKey || event.altKey) return;
```
kowinauyeung, line 101: 🔴 **BLOCKER — the document handler stays live under the add-word modal, and answers cards through it.**

`keyboard={!quitting}` closes exactly one hole: the `ConfirmDialog` this component knows about. But `AppLayout` renders ＋追加 (desktop header) and the ＋ FAB (mobile) on *every* route, `/practice/:mode` included, and `Modal` binds its own Escape on `document` without stopping propagation or blocking anything else. So with a session running and the add-word sheet open over it, this listener still sees every keystroke typed into the form.

Three consequences, all reproduced against the real components (`FlashcardSession` + `Modal` + an input, in jsdom):

1. **A card is silently answered.** Flip the card, open ＋追加, press `→` to move the caret in 見出し語 — `onAnswer(true)` fires, the word is scored, and the queue advances behind the sheet. The learner never saw the card and cannot tell it happened. `苦手な語` is built out of these answers, so this writes wrong data.
2. **Space cannot be typed.** `event.preventDefault()` on line 71 runs for a keydown that originated in the text field, so the space character is never inserted — and the card behind flips instead.
3. **Escape does both jobs.** It closes the modal *and* calls `onQuit()`, so the 中断 dialog opens on the way out.

All three failed as written, which is what confirms them:

```
× arrow keys typed into a modal field still answer the card behind it
    → expected "spy" not to be called, but was called 1 times
× space typed into a modal field is not preventDefault-ed
    → expected false to be true
× escape in the modal does not also ask to quit the session
    → expected "spy" not to be called, but was called 1 times
```

The `keyboard` prop is the right seam, it is just gated on the wrong thing. Two options:

- **Ignore editable targets** — bail when `event.target` is an `input`/`textarea`/`[contenteditable]`, or when `document.activeElement` is one. This is self-contained and also covers any future overlay.
- **Gate on "is any dialog open"** rather than on `quitting` alone — correct too, but it makes every future overlay remember to tell the session about itself, which is how this hole appeared in the first place.

I'd take the first, and keep the `!quitting` gate as well since it also stops the Escape round-trip. Worth a component test at this layer — the modal-over-session composition is a DOM-event claim, so `tests/component` is the cheapest place that can see it.

src/lib/speech.ts
```diff
@@ -0,0 +1,252 @@
...
+ */
+export function speechCandidates(
+  available: readonly SpeechSynthesisVoice[],
+): (SpeechSynthesisVoice | null)[] {
+  const candidates = japaneseVoices(available);
+  return [
+    ...new Set([
+      candidates.find((voice) => voice.localService) ?? null,
+      candidates.find((voice) => !voice.localService) ?? null,
+      null,
```
kowinauyeung: 🟡 **The candidate order inverts to [browser default, network voice] when the machine has no local Japanese voice.**

When `candidates.find(localService)` returns nothing it contributes `null` — the same value as the trailing default entry — so `new Set` collapses them into the *first* slot rather than the last:

```
speechCandidates([Samantha(en-US), Google 日本語(ja-JP, network)])
  → [null, Google 日本語]      // measured, not inferred
```

So on exactly the machine the doc comment above describes — one where the network voice is "the one candidate that cannot be missing" — the first attempt is the unnamed default and the explicitly-Japanese voice is demoted to second. The doc comment two lines up promises the opposite order.

It is not silent-failure territory (`utterance.lang = 'ja-JP'` still asks the UA to resolve a Japanese voice), which is why this is a suggestion and not a blocker. But it does defeat the ordering this function exists to express, and if the default attempt *starts* while sounding wrong, nothing ever advances the candidate — the 2s backstop only fires on silence.

Filter the nulls out before the trailing fallback rather than deduping against it:

```ts
const local = candidates.find((v) => v.localService);
const network = candidates.find((v) => !v.localService);
return [local, network, null].filter((v) => v !== undefined);
```

`tests/unit/speech.test.ts` covers local-only, local+network and nothing-Japanese, but not network-only — which is why this got through. Worth the fourth case.

_(Surfaced by the Codex cross-check; verified here by running it.)_

src/lib/speech.ts
```diff
@@ -0,0 +1,252 @@
...
+     * One `cancel()` before this page has queued anything of its own clears
+     * whatever it inherited. It is a no-op on a clean engine.
+     */
+    engine.cancel();
+
+    const read = () => setVoices(engine.getVoices());
+    read();
+    engine.addEventListener('voiceschanged', read);
+    return () => {
+      engine.removeEventListener('voiceschanged', read);
```
kowinauyeung, line 121: 🟡 **No `cancel()` on unmount — which leaves behind precisely the state this hook's mount does the work to clean up.**

The cleanup drops the `voiceschanged` listener and nothing else. Quitting a session mid-utterance (中断, or Escape, or navigating away from the summary) unmounts `DictationSession` while the word is still queued and playing, so:

- the app keeps speaking after the learner has left the drill, and
- the renderer is handed a live queue that outlives this document — which is the root cause the commit message for 2689ace identifies and works around with the mount-time `cancel()` on line 124.

The mount-time cancel fixes it on the *receiving* end, for a page unlucky enough to inherit the mess. Cancelling on the way out stops making it. Also worth clearing the pending `START_TIMEOUT_MS` timer here: after an unmount it can still fire, advance `candidate.current` and `console.warn` about a voice nobody is waiting for.

```ts
return () => {
  engine.removeEventListener('voiceschanged', read);
  engine.cancel();
};
```

src/routes/Practice.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const missed = session.queue.filter((entry) => missedIds.has(entry.id));
+
+  return (
+    <SessionSummary
+      total={session.answers.length}
+      correct={session.answers.filter((item) => item.correct).length}
+      missed={missed}
+      saving={saving}
+      saveError={saveError}
+      onRestart={() => start(shuffle(matches, Math.random), describeFilters(filters, sets))}
```
kowinauyeung: 🟡 **`もう一度` re-filters against live state, so a perfect 苦手のみ session restarts into an empty queue.**

`matches` is recomputed from `weakIds`, and `record()` has just cleared every word the learner got right. Drill 苦手のみ, answer everything correctly, press もう一度 — `matches` is now `[]`, `start([])` sets a session whose `queue[0]` is undefined, and the summary re-renders as `0 / 0`. Not destructive, but it is a dead end at exactly the moment the learner did well.

Restarting from `session.queue` says what もう一度 means ("the same drill again") and sidesteps it. If re-filtering is deliberate, the button should at least disable itself when `matches` is empty, the same way 開始する already does on the setup screen.

_(Surfaced by the Codex cross-check; confirmed against the code here.)_

src/routes/Practice.tsx
```diff
@@ -0,0 +1,210 @@
...
+  if (!session) {
+    return (
+      <PracticeSetup
+        mode={mode}
+        filters={filters}
+        allTags={allTags}
+        allSets={sets}
+        now={now}
+        matchCount={matches.length}
+        weakCount={weakIds.size}
```
kowinauyeung: 🟡 **`weakCount` and `matchCount` are counted over different populations, so the toggle can advertise words that cannot be drilled.**

`weakIdsOf` counts progress rows; `matches` counts entries. A word deleted after being answered wrong keeps its row in the progress map — nothing prunes it — so 苦手な語のみ can read 「1 語」 while ticking it produces 「0 件が対象」 and a disabled 開始する.

Counting the intersection with the entries already in hand is a one-liner here and makes the label mean what it says. Pruning progress on entry delete is the more thorough fix, but it belongs with `/wordsets` and the rest of the write side rather than in this PR.

_(Surfaced by the Codex cross-check; confirmed against the code here.)_

src/lib/wordSets.tsx
```diff
@@ -0,0 +1,71 @@
...
+    setLoading(true);
+    setError(null);
+    try {
+      const all: WordSet[] = [];
+      let cursor: string | null = null;
+      do {
+        const page = await repository.list({ limit: PAGE_SIZE, cursor });
+        all.push(...page.items);
+        cursor = page.cursor;
+      } while (cursor);
```
kowinauyeung, line 70: 🟡 **No cancellation guard on this loop, unlike its sibling provider.**

`ProgressProvider` runs the same shape of effect with a `cancelled` flag so a resolved promise from the previous `uid` cannot land on the new one. This one pages through every set with `await` inside a `do/while` and then calls `setSets` unconditionally, so a `uid` change mid-walk can finish by publishing the previous account's 単語集. One account today makes it unreachable; the two providers reading differently is the part that will not survive contact with a second one.

Separately: the comment on `PAGE_SIZE` calls 200 "a ceiling", but the loop keeps going while `cursor` is non-null, so it is a page size and there is no ceiling. Either cap the walk or reword the comment — as written the next reader has to run it to find out which one is true.

src/lib/speech.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * A network voice is silent whenever the connection is, and this is a drill
+ * somebody may be doing on a train. macOS's Kyoko is local and better than the
+ * remote alternatives for single words anyway.
+ */
+export function pickJapaneseVoice(
+  available: readonly SpeechSynthesisVoice[],
+): SpeechSynthesisVoice | null {
+  const candidates = japaneseVoices(available);
+  return candidates.find((voice) => voice.localService) ?? candidates[0] ?? null;
```
kowinauyeung: 🔵 `pickJapaneseVoice` has no production caller — `speechCandidates` superseded it, and the only thing importing it now is `tests/unit/speech.test.ts`, which spends three cases on it.

CLAUDE.md asks that a test correspond to a real behaviour, a backlog item or a defect; three that pin an unreachable function read as coverage and prove nothing about the app. Its "prefer local" intent already lives in `speechCandidates` and is tested there. Suggest deleting both, unless something in the 履歴/wordsets work is expected to want it.

src/components/practice/FlashcardSession.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+      if (event.key === 'Escape') {
+        onQuit();
+        return;
+      }
+      if (event.key === ' ') {
+        // Space scrolls the page by default, which moves the card being read.
+        event.preventDefault();
+        setFlipped((current) => !current);
```
kowinauyeung, line 111: 🔵 Space toggles in both directions, but the hint is printed only on 裏を見る, so on a flipped card it silently un-flips instead of doing nothing or answering. Minor, and arguably useful — worth a word in the comment above either way, since the asymmetry with the arrows (inert until flipped) is deliberate and this one is not obviously so.

tests/integration/progressRepo.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  });
+
+  /**
+   * The defect: a write without `merge: true` replaces the whole `entries` map
+   * with whatever this session touched, so every word the learner practised in
+   * an earlier session — or on another device a minute ago — silently reverts
+   * to unpractised. Nothing on screen would show it until 苦手のみ came back
+   * with a smaller number than it had.
+   */
+  it('leaves rows this session did not answer exactly as they were', async () => {
```
kowinauyeung, line 98: ✅ This is the right test in the right place. `merge: true` is unobservable from any cheaper layer — a `Map`-backed substitute merges by construction — and standing up a second repository over the same uid to stand in for a second device is a much better reproduction of the defect than writing twice through one handle would have been. Same for `finishedAt` being asserted as a raw `Timestamp` rather than a plausible-looking instant.

tests/component/DictationSession.test.tsx
```diff
@@ -0,0 +1,388 @@
...
+ * signal distinguishing the two meanings of the key. jsdom fires nothing on its
+ * own; a real IME emits compositionstart, then keydown with `isComposing`, then
+ * compositionend.
+ *
+ * `entry` here is deliberately kana-only, so the *unguarded* behaviour would
+ * score the mid-composition text as correct. A guard that fired on the wrong
+ * event would then flip a passing assertion, which a kanji headword — wrong
+ * either way — could not tell apart.
+ */
+const entry = makeEntry({ id: 'e1', headword: 'ちょっと', reading: '' });
```
kowinauyeung, line 24: ✅ Choosing a kana-only headword so the *unguarded* behaviour would score as correct is the detail that makes this test able to fail for the right reason. A kanji headword is wrong either way and would have passed with the guard removed. The reasoning is written down above it too, which is what lets the next reader keep it that way.

src/lib/practice.ts
```diff
@@ -0,0 +1,242 @@
...
+ * A named object rather than two positional `Set`s: they are the same type,
+ * so swapping them at a call site is invisible to the compiler and turns
+ * 苦手のみ into 単語集 without a single error.
+ */
+export interface PracticeScope {
+  /** Entries whose most recent attempt was wrong. */
+  weak: ReadonlySet<string>;
+  /** Members of the selected 単語集, or null when none is selected. */
+  inSets: ReadonlySet<string> | null;
+}
```
kowinauyeung, line 97: ✅ Making this a named object rather than two positional `ReadonlySet<string>` is a real fix for a real hazard — the compiler cannot tell those apart, and swapping them turns 苦手のみ into 単語集 silently. Same for `inSets: null` vs an empty set: collapsing them would make an empty 単語集 select everything, which is the opposite of what it says.

kowinauyeung:
## PR Review

**Verdict:** `REQUEST CHANGES` — one blocker, and it is a small fix.

Reviewed as two independent passes (this one and a Codex `gpt-5.5` pass), then merged. Every finding below was verified against the code here; the two Codex surfaced are marked as such.

---

### 🔴 Blocker

**`FlashcardSession`'s document-level shortcuts stay live under the add-word modal.** `keyboard={!quitting}` closes the one hole this component knows about, but `AppLayout` renders ＋追加 and the ＋ FAB on every route including `/practice/:mode`, and `Modal` binds Escape on `document` without stopping propagation. With a session running and the add-word sheet open over it:

- `→` typed into a form field **answers the card behind it** — scored, queued on, invisible to the learner, and `苦手な語` is built out of these answers;
- **Space cannot be typed at all**, because `preventDefault()` runs for a keydown that originated in the text field;
- Escape closes the modal *and* opens 中断.

Reproduced against the real components in jsdom — all three assertions failed as written. Details and two suggested fixes are inline.

### 🟡 Suggestions

| | Where | What |
| --- | --- | --- |
| 1 | `src/lib/speech.ts:66` | Candidate order inverts to `[browser default, network voice]` when no local Japanese voice exists — the missing local slot dedupes into the trailing `null`. Measured, not inferred. The one voice-list shape `tests/unit/speech.test.ts` does not cover. _(Codex)_ |
| 2 | `src/lib/speech.ts:129` | No `cancel()` on unmount, so quitting mid-utterance keeps speaking **and** hands the renderer the live inherited queue that 2689ace identifies as the root cause. The mount-time `cancel()` fixes the receiving end; this stops making the mess. |
| 3 | `src/routes/Practice.tsx:206` | `もう一度` re-filters live `matches`, so a perfect 苦手のみ session restarts into an empty queue and re-renders as `0 / 0`. _(Codex)_ |
| 4 | `src/routes/Practice.tsx:151` | `weakCount` counts progress rows, `matchCount` counts entries — a deleted word keeps its row, so the toggle can read 「1 語」 over 「0 件が対象」. _(Codex)_ |
| 5 | `src/lib/wordSets.tsx:39` | No `cancelled` guard on the paging loop, unlike the sibling `ProgressProvider`; and the `PAGE_SIZE` comment calls 200 "a ceiling" when the loop pages unboundedly. |
| 6 | `src/lib/speech.ts:41` | `pickJapaneseVoice` has no production caller — superseded by `speechCandidates`, kept alive by three unit tests. |

### ✅ Highlights

- **The test layering is right, and unusually so.** `merge: true` and the `finishedAt` cursor against the emulator; pure filtering and marking in `tests/unit`; the IME guard and every speech failure in `tests/component`; end-to-end reserved for routing, providers and a real key press. Asserting `finishedAt` as a raw `Timestamp` rather than a plausible-looking instant is the version of that test that can actually fail.
- **The IME test uses a kana-only headword** so the *unguarded* behaviour would score as correct. A kanji headword is wrong either way and would have passed with the guard removed — this is the detail that makes the test able to fail for the right reason.
- **`PracticeScope` as a named object** instead of two positional `ReadonlySet<string>`, and `inSets: null` kept distinct from an empty set. Both close hazards the compiler cannot see.
- **`subtractMonths` clamps** rather than letting `setMonth` roll 2月31日 forward to 3月3日, with the range direction of the bug stated in the comment.
- **The speech commit series** is the honest record of narrowing to a measurement, and the write-up of why autoplay came back (Safari plays it on the same machine, so Chrome was the thing to work around) is the right call rather than the convenient one.

### Verification run here

- `vitest run tests/unit tests/component` — **377 passed**, 16 files.
- `tsc --noEmit -p tsconfig.json` — clean.
- Emulator and Playwright suites **not** run in this review; the PR's own numbers (53 emulator, 31 e2e) are taken as reported.
- The rules change is a deny-side correction only. The owner-side positive path for `users/{uid}/progress/**` and `practiceSessions/**` is covered structurally by the existing `match /{document=**}` wildcard, not by a test of its own — fine as is, worth knowing.

---

### Summary

Strong PR, and the parts that are hard to get right are the parts that are right: the storage decision is argued from how Firestore actually bills, the test layers are chosen rather than defaulted, and the speech work ends in a measurement instead of a guess. The 3,990-line diff is the one structural cost — the practice feature and the seven-commit speech investigation are separable, and splitting them would have made both easier to review.

The blocker is narrow: one listener that should ignore editable targets. Everything else is a suggestion, and 1–4 are each a few lines. Happy to re-review once the keyboard gate is fixed.

src/components/practice/FlashcardSession.tsx
kowinauyeung: @kowinauyeung
Confirmed and fixed — `AppLayout.tsx:91` renders `<Outlet/>` and `:103` renders `<EntryFormModal>` as siblings, so both are live on `/practice/:mode`. All three consequences reproduce.

Took the editable-target guard as you recommended, and added a second question alongside it: whether any `[role="dialog"]` is in the document. The editable check alone leaves the Escape round-trip open whenever focus is on a modal *button* rather than a field, which is reachable — the sheet's ✕ and 保存する are both buttons. Two questions rather than a register of overlays, for the reason you gave.

Kept `keyboard={!quitting}` as you asked. It is strictly redundant now that the dialog query also matches `ConfirmDialog`, and I would normally delete a branch nothing can distinguish — but it is an explicit signal from the owner rather than a DOM query being right, it is independently testable at the prop level, and the failure it guards writes wrong data. Different calculus from a dead branch in a pure function.

Four component cases added, composing the real `Modal` rather than a stand-in. Removing the guard turns exactly those four red — your three plus focus-outside-the-field — and nothing else.

Fixed in 8b992d9

src/lib/speech.ts
kowinauyeung: @kowinauyeung
Confirmed by running it before agreeing:

```
speechCandidates([Samantha(en-US), Google 日本語(network)])
  → [ NULL, 'Google 日本語' ]
```

Exactly as you measured, and the doc comment two lines up promises the opposite. Absent entries are now dropped rather than coerced with `?? null`, so the trailing fallback stays last:

```ts
return [
  candidates.find((voice) => voice.localService),
  candidates.find((voice) => !voice.localService),
  null,
].filter((voice) => voice !== undefined);
```

Fourth unit case added for network-only, and it is red against the old shape.

Fixed in 8b992d9

src/lib/speech.ts
kowinauyeung: @kowinauyeung
Agreed on both halves. The mount-time cancel cleans up after a page unlucky enough to inherit a live queue; this stops handing the next one over. Quitting mid-utterance via 中断, Escape or navigating off the summary is exactly that case.

The pending backstop goes with it. It is now tracked in a ref so the cleanup can clear it — after an unmount it could still fire, advance `candidate.current` and warn about a voice nobody was waiting for.

Not separately tested: an unmount-time `cancel()` is observable only as the absence of a later effect, and the assertion I could write ("cancel was called on teardown") is the kind of call-spy CLAUDE.md rules out. Flagging that rather than writing one that looks like proof.

Fixed in 8b992d9

src/routes/Practice.tsx
kowinauyeung: @kowinauyeung
Confirmed — `record()` runs before the summary renders, so a perfect 苦手のみ run clears every word it drilled and `matches` is `[]` by the time もう一度 is pressed.

Restarting from `session.queue` rather than disabling the button: re-filtering was not deliberate, it was me reaching for the nearest list. "The same drill again" is what the label says and it cannot produce an empty queue by construction, so it needs no guard.

End-to-end case added — a perfect weak-words session, then もう一度 — since the dead end spans the write, the provider and the re-render. Red against the old line.

Fixed in 8b992d9

src/routes/Practice.tsx
kowinauyeung: @kowinauyeung
Confirmed. Counted over the entries in hand now, which is the one-liner you describe.

Agreed that pruning progress when an entry is deleted is the real fix and does not belong here — it is a write-side concern and there is no delete path through `ProgressRepository` yet.

End-to-end case seeds a weak id that no entry has, and asserts the label reads 1 語 rather than 2.

Fixed in 8b992d9

src/lib/wordSets.tsx
kowinauyeung: @kowinauyeung
Both fixed. The walk now carries a generation counter rather than a `cancelled` flag: `refresh` is also callable by hand, so the flag cannot live in the effect the way `ProgressProvider`'s does, and the counter covers a superseded walk and a concurrent refresh with the same line.

On the comment — you are right, it was describing a cap that does not exist. Reworded to say it is a page size and nothing more.

One narrowing worth flagging: `EntriesProvider` has the same unguarded shape and no `cancelled` flag either, so mine copied the older of two patterns rather than diverging from the newer one. I have left it alone — it predates this branch and changing it turns a feature pull request into a provider refactor. Worth its own change.

Fixed in 8b992d9

src/lib/speech.ts
kowinauyeung: @kowinauyeung
Confirmed zero callers in `src` — `speechCandidates` superseded it and I left the corpse behind. Deleted along with its three tests.

Nothing in the 履歴 or wordsets work wants it: both need session records and set membership, not voice selection. If a voice picker is ever needed again it will want the candidate list, not a single choice.

Fixed in 8b992d9

src/components/practice/FlashcardSession.tsx
kowinauyeung: @kowinauyeung
Fair — the asymmetry is real and only one half of it was written down. Comment now says Space turns the card both ways and why that differs from the arrows: re-hiding is harmless and occasionally wanted, whereas answering early is not recoverable.

Behaviour unchanged; the existing case already covers the toggle in both directions.

Fixed in 8b992d9

src/components/practice/FlashcardSession.tsx
kowinauyeung: Verified against 8b992d9, not taken on the description. `ownsKeyboard` asks the two questions you describe, and I re-ran my original three cases plus a regression case against the real `FlashcardSession` + `Modal` composition in jsdom:

```
✓ arrow keys typed into a modal field no longer answer the card behind it
✓ space typed into a modal field is not preventDefault-ed
✓ escape in the modal does not also ask to quit the session
✓ still flips and answers from the keyboard with no modal present
```

All four green; the first three were the ones that failed before. The regression case matters here because `document.querySelector('[role="dialog"]')` would have killed the shortcuts outright if any dialog were in the DOM while closed — `Modal` returning `null` when shut is what makes the guard safe, and that is now load-bearing on behaviour rather than on styling.

Agreed on two questions over a register, and on `role="dialog"` being the right thing to key on: it is the contract an overlay already has to honour for assistive tech, so a new one cannot opt out of it and stay correct. Checking `event.target` *and* `document.activeElement` also covers the case where a keystroke lands on `document` while focus sits in a field, which the target check alone would miss.

Resolved.

src/lib/speech.ts
kowinauyeung: Verified by re-running the same measurement against 8b992d9:

```
speechCandidates([Samantha(en-US), Google 日本語(network)])
  → [ 'Google 日本語', BROWSER-DEFAULT ]
```

Was `[BROWSER-DEFAULT, 'Google 日本語']`, so the documented order now holds on exactly the machine the network voice exists for. Dropping absent entries instead of coercing them is the right shape — the bug was `??` manufacturing a value that then collided with a real one, and `filter(v => v !== undefined)` removes the collision rather than working around it.

The fourth unit case covers the shape that let it through. Resolved.

src/lib/speech.ts
kowinauyeung: Both halves verified against 8b992d9. I wrote two throwaway cases to check the behaviour is real rather than just present in the diff:

```
✓ cancels the queue on unmount rather than leaving it for the next page
✓ drops the pending backstop so it cannot warn after unmount
```

Both green — `cancel()` is called a second time on unmount, and advancing fake timers 5s past an unmount produces no `console.warn`. Tracking the timer in a ref so the cleanup can reach it, and the `settle()` helper clearing both the timeout and the ref, is tidier than what I had suggested.

One thing to flag rather than block on: **this is the only one of the six fixes with no test in the branch.** The commit message is honest about it — the proved-red list names the keyboard guard, the `??` dedupe, the restart and the row-based count, and not this. The two cases above are cheap and sit in the layer this repo would put them in, so they are yours if you want them. CLAUDE.md's own line applies: a fix that has never been red is a fix whose failure mode is unknown.

Resolving the thread since the defect itself is fixed and verified; the coverage gap is in the summary comment instead.

src/routes/Practice.tsx
kowinauyeung: Verified. `onRestart` now shuffles `session.queue` and reuses `session.filterLabel` rather than re-deriving either, so the restart cannot depend on state the recorded session has already changed.

Agreed that restarting the same words beats disabling the button — the label says もう一度, and a control that disables itself precisely when the learner has just done well is a worse answer to the same problem.

The end-to-end case is well chosen: it drills a weak word, answers it correctly so the progress write clears it, and *then* presses もう一度 — which is the exact sequence, not something adjacent to it. Under the old line that lands on a 0 / 0 summary. Resolved.

src/routes/Practice.tsx
kowinauyeung: Verified. `weakCount` is now `entries.filter(entry => weakIds.has(entry.id)).length`, memoised on the same inputs, so the label and 件が対象 are counted over one population.

The end-to-end case seeds `deleted-long-ago` — a weak id no entry carries — alongside a real one and asserts 「1 語」 rather than 「2 語」. That is the defect reproduced rather than described; a seed with only real ids would pass either way.

Agreed that pruning the progress map on entry delete is the real fix and is out of scope here — there is no delete path through `ProgressRepository` yet, and adding one to a practice PR would be the wrong place for it. Worth a line in the backlog rather than a comment that ages. Resolved.

src/lib/wordSets.tsx
kowinauyeung: Verified. The generation counter guards all three setters, not just `setSets` — a superseded walk cannot flip `loading` or publish an error either, which the `cancelled` flag I pointed at would have needed care to get right.

Your reason for not copying `ProgressProvider`'s shape is better than my suggestion: `refresh` is exposed on the context and callable by hand, so a flag scoped to the effect cannot see a concurrent manual call. The counter covers both with one line.

The `PAGE_SIZE` comment now says what the code does. And I agree with leaving `EntriesProvider` alone — it predates this branch and has its own shape; widening a feature PR into a refactor of it would have made this diff harder to review, not easier.

Not tested, which I think is right at this size — the race needs a controllable repository and two overlapping walks to express, and that is more machinery than the guard is worth today. Worth knowing it is unguarded by anything but review. Resolved.

src/lib/speech.ts
kowinauyeung: Verified — `pickJapaneseVoice` is gone from `src/lib/speech.ts` and its three cases are gone from `tests/unit/speech.test.ts`. `grep` finds no remaining reference. Suite is 379 passing (was 377: five added, three removed).

Agreed on the reasoning for not keeping it around for 履歴 or wordsets — those need session records and set membership, and a future voice picker would want the candidate list rather than a single choice. Resolved.

src/components/practice/FlashcardSession.tsx
kowinauyeung: Verified. The comment now states that Space turns the card both ways and why that differs from the arrows — re-hiding is recoverable, answering early is not. That is the half that was missing.

Behaviour unchanged, as you say, and the existing case already exercises the toggle in both directions, so nothing new is warranted here. Resolved.

kowinauyeung:
## Re-review — 8b992d9

**Verdict:** `APPROVE`. The blocker is fixed and all eight threads are resolved.

Every fix was checked against the code and re-measured rather than taken from the commit message. Details are in the individual threads; the short version:

| # | Finding | State | How it was verified |
| --- | --- | --- | --- |
| 🔴 | Document shortcuts live under the add-word sheet | **Fixed** | Re-ran my three original failing cases against the real `FlashcardSession` + `Modal` composition — all green, plus a regression case proving the shortcuts still work with no overlay present |
| 🟡 1 | `speechCandidates` order inverted with no local voice | **Fixed** | Re-measured: `['Google 日本語', BROWSER-DEFAULT]`, was `[BROWSER-DEFAULT, 'Google 日本語']` |
| 🟡 2 | No `cancel()` on unmount | **Fixed, untested** — see below | Wrote two throwaway cases: `cancel()` does fire on unmount, and 5s of fake timers past an unmount produces no warn |
| 🟡 3 | `もう一度` re-filtered live state | **Fixed** | Reads `session.queue` / `session.filterLabel`; the e2e case drills a weak word, answers it right, *then* restarts — the exact sequence |
| 🟡 4 | `weakCount` vs `matchCount` populations | **Fixed** | Counted over entries in hand; e2e seeds a weak id no entry carries and asserts 1 語 not 2 |
| 🟡 5 | `wordSets` unguarded walk + wrong comment | **Fixed** | Generation counter guards all three setters, not just `setSets`; comment now describes the loop that exists |
| 🔵 6 | `pickJapaneseVoice` dead | **Fixed** | Gone from `src` and from the tests; no references remain |
| 🔵 7 | Space toggles both ways, undocumented | **Fixed** | Comment states it and why it differs from the arrows |

Two of these went beyond what I suggested and are better for it. The keyboard guard checks `event.target` **and** `document.activeElement`, which catches a keystroke landing on `document` while focus sits in a field — the target check alone would miss that. And `WordSetsProvider`'s generation counter beats the `cancelled` flag I pointed at, because `refresh` is exposed on the context and callable by hand, so a flag scoped to the effect cannot see a concurrent manual call.

One consequence worth writing down, since it is now load-bearing: `ownsKeyboard` reads `document.querySelector('[role="dialog"]')`, so a `Modal` that rendered while closed instead of returning `null` would silently kill every flashcard shortcut. `Modal` returns `null` today and the regression case above pins the working state, but that is a behavioural dependency between two components that do not import each other.

### The one gap

**The unmount `cancel()` is the only fix in this commit with no test.** The commit message is straight about it — the proved-red list names the keyboard guard, the `??` dedupe, the restart and the row-based count, and not this one. Not a blocker: I verified the behaviour myself and it is correct. But CLAUDE.md's own line applies, that a fix which has never been red is a fix whose failure mode is unknown, and this one is cheap to cover at the component layer:

```
✓ cancels the queue on unmount rather than leaving it for the next page
✓ drops the pending backstop so it cannot warn after unmount
```

Both pass against 8b992d9 and both are in the layer this repo would put them in. Posted in the thread; take them or leave them.

`WordSetsProvider`'s counter is also untested, which I think is the right call at this size — expressing the race needs a controllable repository and two overlapping walks, and that is more machinery than the guard is worth today.

### Verification run here

- `vitest run tests/unit tests/component` — **379 passed**, 16 files (was 377: five added, three removed with `pickJapaneseVoice`).
- `tsc --noEmit` against both `tsconfig.json` and `tsconfig.test.json` — clean.
- Emulator and Playwright suites still not run in this review; the two new e2e cases were read, not executed.

### Summary

Good turnaround. Each reply engaged with why the finding was a finding rather than just changing the line, two of the fixes are better than what was asked for, and where a suggestion was declined — leaving `EntriesProvider` alone, deferring progress-map pruning to a write-side change — the reason given was the right one and I agree with both. Approving.

The scope note from the first pass still stands for next time: the practice feature and the speech investigation were separable, and splitting them would have made both easier to review. Nothing to do about it now.

PR #17, opened 2026-08-10
feat(wordsets): create, fill, arrange and delete a word set

kowinauyeung:
### Summary

Phase 2, item 4: the write side of 単語集. `/wordsets` lists sets and starts one
from a name; `/wordsets/:id` fills it, arranges it, renames it and deletes it.

The read side already existed — a Firestore `WordSetRepository`, `sanitizeWordSet`,
`WordSetsProvider`, and practice filter chips that hid themselves while there were
no sets. Nothing could write one, so those chips had never appeared in the running
app. They do now.

Two fixes ride along, both reachable only because sets became writable.

### What changed

**Screens**

- `/wordsets` — the list, plus a name field that creates a set and opens it. A
  set is worth nothing until it has words in it, and the screen that adds them is
  the next one.
- `/wordsets/:id` — two bounded, scrolling panels side by side: 単語を探す on the
  left, 収録語 on the right, with search and filters across the top. Add a word by
  dragging it across or pressing ＋追加; arrange the study order by dragging within
  収録語 or with the arrow keys; rename, empty or delete the set.

**Finding words**

The picker narrows through 一覧's own `FilterPanel` over the same
`matches`/`sortEntries`, not a second search with its own rules: a combination
that finds a word on 一覧 has to find it here, and two implementations is how that
stops being true. With nothing set it shows the whole notebook newest-first —
a word is usually filed into a set soon after it is written down.

**Bulk**

`表示中の N 語を追加` and `表示中の N 語を削除`, both named after the rows on
screen rather than after the match count, which is capped at 50 and usually
larger. Removing confirms first and adding does not: emptying the list takes the
study order with it, and that order took dragging to build and is recorded
nowhere else.

**Data**

`membersOf`, `candidatesFor`, `withMember(s)`, `withoutMember(s)`,
`reorderMembers`, `insertMemberAt`, `toDraft` in `src/lib/wordSetMembers.ts`;
`useListDrag` and its geometry in `src/lib/listDrag.ts`. `WordSetsProvider` now
exposes its repository, the way `EntriesProvider` does.

No rules change: `users/{uid}/{document=**}` already covers `wordSets`, and the
adapter was merged and tested in #15.

**Membership resolves through the notebook rather than being trusted as stored.**
Deleting a word does not visit the sets that named it — that would be a write per
set for a delete that has nothing to do with them — so `entryIds` can outlive its
members. `membersOf` drops an id with no entry behind it, which is what keeps a
set's count agreeing with the list under it. Everything that counts a set goes
through it. The one place that will not be enough is publishing, where the
snapshot must resolve the list rather than copy the array.

**`level` and `topics` have no field in the edit form.** Both exist to describe a
set to somebody else once it is published, and nothing publishes yet, so a form
for them would ask for metadata no screen reads back. `toDraft` still carries them
through every write, or the first rename would clear them.

**Pointer events, not HTML5 drag-and-drop.** `dragstart` never fires on a
touchscreen. No dependency was added; the hit test reads `data-drop-list` /
`data-drop-index` from the DOM rather than caching rectangles, because both lists
resize as rows move between them.

**A press becomes a drag only after 5px.** A row is also a link to its word, so a
hand that drifts two pixels between press and release has to still be able to open
it.

**Touch drags start from the grip, not the row.** A touch drag needs
`touch-action: none` on whatever it starts from, and a whole row marked that way
is a row the panel can no longer be scrolled by. The grip is a 44px target for
exactly that reason.

**Dragging a member *out* is a no-op.** The symmetric gesture would remove it, and
a slip would then take a word out of a set silently, where 削除 is a button that
says what it does.

**Both panels are bounded and scroll inside themselves.** Stacked one above the
other they drifted apart as soon as either had a few rows, and a drag whose start
and end are never on screen together cannot be performed. Below the nav breakpoint
they stack again, so the cap is a fraction of the viewport rather than a fixed
height. Bounded panels then need edge auto-scroll, on a frame loop rather than
inside `pointermove`, because a pointer held at the bottom of a list fires no more
moves — and holding still at the edge is exactly the gesture that means keep going.

**収録語 says whether it can take a drop.** 単語を探す is a source and refuses
them, so without something marking the list that accepts them, releasing over the
wrong half does nothing and looks broken. This is the one part of the styling that
carries meaning, so it is the one part with a test.

### The two fixes

**The practice chip counted ids, not words** (`4f3b387`). `entryIds.length` over a
set whose second word had been deleted advertised 3 and dealt 2 — the queue was
already right, because `matchesPractice` intersects with the entries in hand.
Unreachable until a set could be written. `PracticeSetup` now takes
`{ id, name, count }` so it cannot recompute the number a different way from its
caller.

**A post-write refresh blanked the page** (`a6b7c08`). Every write ends in its
provider's `refresh()`, and `refresh()` raised the same `loading` flag the initial
load uses — which every route reads as "replace the page with 読み込み中…". Adding
a word to a set took the whole screen down and put it back. The data in hand during
a refresh is one write out of date, not invalid. The gate now goes up only on
mount and on an account change. `EntriesProvider` carried the same defect and is
fixed with it; it shows on the edit path rather than the add path, because adding
navigates to the new word and leaves the blanked page behind. Found by hand on
another device.

### Tests

| Layer | Added | Why there |
| --- | --- | --- |
| `tests/unit` | membership, candidate filtering, reorder arithmetic, drop-index and edge-scroll geometry | all value-in, value-out |
| `tests/component` | the in-flight write guard on every bulk control, the edit form resetting on reopen, keyboard reordering, the drop-state affordance, the folded filter panel | document-level handlers and rendered structure; nothing about them is visible from a return value |
| `tests/e2e` | a set created on one screen reaching the practice filter on another, one word in two sets, a deleted word leaving both, dragging by grip and by row body, reordering, add-all, empty-with-confirmation, and two cases pinning that a write does not blank its page | routing, providers, and a real pointer over a real layout |

No new integration tests: `tests/integration/wordSetRepo.test.ts` already covers
create, update, the `entryIds` round trip, dedupe and delete. No new screenshot
baselines.

**Every fix was proved red first**, by reverting what it guards and confirming
only the expected cases failed. That process changed the change three times:

- The "moving down" reorder case did not discriminate the index correction it was
  written for — `splice` clamps a high index either way. It now drops into an
  interior gap, where the two answers differ.
- The upper `Math.min` clamp was dead for the same reason and is gone. The lower
  one is not: ArrowUp on the first row asks for gap `-1`, which `splice` counts
  back from the end, so without it the top row moved to *second*.
- The `from`-out-of-range guard was already covered by the missing-row check and
  is gone too.

And it found two defects no reading of the code had:

- **An `<a>` is natively draggable.** With the row body as a drag source, a press
  that drifted a few pixels started the browser's own link drag, which cancels the
  click — so a word could no longer be opened at all. The case that found it
  presses and moves 3px, because `locator.click()` releases on the pixel it
  pressed and would have proved nothing.
- **The click-swallow after a drag has never run.** Chromium fires no click at all
  following any drag this suite can produce, including one released on the row it
  started from — measured with a throwaway probe rather than guessed at. The code
  is kept for the browsers the suite does not cover and its comment now says
  plainly that it is unverified; the end-to-end case written to cover it was
  deleted, because a test that cannot go red says nothing.

One expected-red case stayed green and was rewritten rather than trusted: a
blanking case written against ＋追加 on `/vocabulary` passed twice in four runs,
because that flow navigates away and the count it asserted was only ever visible
mid-navigation. It now edits from the detail page, which stays put.

`yarn test:unit` 427 · `yarn test:emulator` 53 · `yarn test:e2e` 46 · typecheck,
lint and format clean. The full end-to-end suite was run three times consecutively
after each of the last three commits, for flake.

### Verified by hand

On a second device, which is where the page-blanking bug was reported from and
where drag-and-drop was found to be awkward enough to need the whole-row grab and
the side-by-side layout.

Not verified by hand: touch drag on a real phone or tablet. It is written for
pointer events and exercised in Chromium with synthesised ones, but no physical
touchscreen has been near it.

### Review round

`kowinauyeung` reviewed it and found two blockers, both one root cause: every row
index arriving from the DOM counts *rendered* rows, and `membersOf` drops an id
whose word has been deleted — so a set holding one renders fewer rows than it
stores. A visible index handed to `reorderMembers` moved whichever id occupied
that slot, usually the stale one: nothing changed on screen and the write still
went out. The keyboard path had the same fault and mattered more, since the arrow
keys are the only way to reorder without a pointer. `storedIndexFor` translates
between the two, unit-tested over `['w1','gone','w2']` and reached end to end by
deleting a word a set holds.

Five suggestions were taken as well: `up` read `dragging` from state while the
target beside it was a ref for exactly the reason that is unsafe — flagged as
reasoning without a reproduction, and it does reproduce, in
`tests/component/listDrag.test.tsx`, by holding React from flushing between the
move and the release; a press beginning on 削除 or ＋追加 armed a drag, so a
drifted press removed nothing and dropped something instead; rows were not gated
on `busy` the way the grip was, so a drag during a write ran its whole course and
was then declined silently; `useEntries().error` was read on neither word-set
screen, so a failed notebook load rendered a set that looked empty; and
`EntriesProvider.refresh` gained the generation counter `WordSetsProvider` has
carried since #15.

One suggestion was declined with reasoning: `membersOf` rebuilding a map per set
is O(N × entries), which the review itself judged not worth changing at this size.

### Not in this

Phase 2 items 5 and 6 remain: `/history` over `ProgressRepository.listSessions`,
and the dashboard's 最新の練習 panel.

Known and deliberate: no long-press to drag on touch (the grip is the way in
there), no screenshot baseline for the new page, and nothing prunes an id from a
set when its word is deleted — the resolution above makes that invisible
everywhere except a future publish.

### DIFF
--- a/src/components/wordsets/MemberList.tsx
+++ b/src/components/wordsets/MemberList.tsx
@@ -35,18 +35,29 @@ export function MemberList({
   const at = drag.at?.list === list ? drag.at.index : null;
 
   return (
-    <section className="space-y-2 rounded-card bg-card p-5 shadow-panel">
-      <div className="flex items-baseline justify-between gap-3">
+    <section className="flex min-h-0 flex-1 flex-col rounded-card bg-card shadow-panel">
+      <div className="flex shrink-0 items-baseline justify-between gap-3 border-b border-line p-4">
         <h2 className="text-sm font-semibold">収録語</h2>
         <p className="text-xs text-muted tabular-nums">{members.length} 語</p>
       </div>
 
-      <ul data-drop-list={list} data-drop-count={members.length} className="min-h-16">
+      <ul
+        data-drop-list={list}
+        data-drop-count={members.length}
+        className="min-h-0 flex-1 overflow-y-auto px-4 py-2"
+      >
         {members.map((entry, index) => (
           <li
             key={entry.id}
             data-drop-index={index}
-            className={`flex items-center gap-2 border-y-2 border-transparent py-1.5 ${
+            {...drag.rowProps({ list, id: entry.id, index })}
+            /**
+             * `touch-pan-y`, not `touch-none`: a finger on a row scrolls the
+             * panel, because a bounded panel whose rows cannot be scrolled from
+             * is a panel that cannot be read. Touch drags start from the grip,
+             * which is the only element here that claims the gesture.
+             */
+            className={`flex touch-pan-y items-center gap-2 border-y-2 border-transparent py-1.5 ${
               at === index ? 'border-t-accent' : ''
             } ${at === members.length && index === members.length - 1 ? 'border-b-accent' : ''} ${
               drag.dragging?.list === list && drag.dragging.id === entry.id ? 'opacity-40' : ''
@@ -58,7 +69,18 @@ export function MemberList({
               onPointerDown={drag.handleProps({ list, id: entry.id, index }).onPointerDown}
               onMoveBy={(delta) => onMoveBy(index, delta)}
             />
-            <Link to={`/vocabulary/${entry.id}`} className="min-w-0 flex-1 truncate">
+            <Link
+              to={`/vocabulary/${entry.id}`}
+              /**
+               * An `<a>` is natively draggable, and the row body is now a drag
+               * source. Left alone, a press that drifts even a few pixels
+               * starts the browser's own link drag instead — which cancels the
+               * click, so the word cannot be opened, and fights our pointer
+               * drag for the gesture.
+               */
+              draggable={false}
+              className="min-w-0 flex-1 truncate select-none"
+            >
               <Ruby
                 headword={entry.headword}
                 reading={entry.reading}
@@ -78,11 +100,11 @@ export function MemberList({
 
         {members.length === 0 && (
           <li
-            className={`grid min-h-16 place-items-center rounded-panel border-2 border-dashed text-sm text-muted ${
+            className={`grid min-h-24 place-items-center rounded-panel border-2 border-dashed px-3 text-center text-sm text-muted ${
               at === 0 ? 'border-accent text-accent' : 'border-line'
             }`}
           >
-            下の一覧からドラッグ、または「＋ 追加」で入れてください
+            ここに単語をドラッグ、または「＋ 追加」で入れてください
           </li>
         )}
       </ul>
--- a/src/components/wordsets/MemberList.tsx
+++ b/src/components/wordsets/MemberList.tsx
@@ -1,0 +1,91 @@
+import { Link } from 'react-router-dom';
+import { Ruby } from '@/components/Ruby';
+import type { Entry } from '@/domain/entry';
+import type { ListDrag } from '@/lib/listDrag';
+import { DragHandle } from './DragHandle';
+
+/**
+ * 収録語 — the set's words, in the order it stores them.
+ *
+ * That order is the study order, and until this list could be dragged there was
+ * no way to change it: words joined at the end and stayed there. It is why the
+ * rows carry a grip and the picker's rows carry one too.
+ *
+ * `border-y-2 border-transparent` is on every row rather than only the one
+ * being dropped onto. A border added on hover would grow the row by 4px, move
+ * every row below it, and hand the pointer a different target than the one the
+ * indicator was drawn for — the drop then lands a row away from where it looked.
+ */
+export function MemberList({
+  members,
+  list,
+  drag,
+  busy,
+  onRemove,
+  onMoveBy,
+}: {
+  members: readonly Entry[];
+  /** This list's name in the drag protocol; see `useListDrag`. */
+  list: string;
+  drag: ListDrag;
+  busy: boolean;
+  onRemove: (entryId: string) => void;
+  onMoveBy: (index: number, delta: number) => void;
+}) {
+  const at = drag.at?.list === list ? drag.at.index : null;
+
+  return (
+    <section className="space-y-2 rounded-card bg-card p-5 shadow-panel">
+      <div className="flex items-baseline justify-between gap-3">
+        <h2 className="text-sm font-semibold">収録語</h2>
+        <p className="text-xs text-muted tabular-nums">{members.length} 語</p>
+      </div>
+
+      <ul data-drop-list={list} data-drop-count={members.length} className="min-h-16">
+        {members.map((entry, index) => (
+          <li
+            key={entry.id}
+            data-drop-index={index}
+            className={`flex items-center gap-2 border-y-2 border-transparent py-1.5 ${
+              at === index ? 'border-t-accent' : ''
+            } ${at === members.length && index === members.length - 1 ? 'border-b-accent' : ''} ${
+              drag.dragging?.list === list && drag.dragging.id === entry.id ? 'opacity-40' : ''
+            }`}
+          >
+            <DragHandle
+              label={entry.headword}
+              disabled={busy}
+              onPointerDown={drag.handleProps({ list, id: entry.id, index }).onPointerDown}
+              onMoveBy={(delta) => onMoveBy(index, delta)}
+            />
+            <Link to={`/vocabulary/${entry.id}`} className="min-w-0 flex-1 truncate">
+              <Ruby
+                headword={entry.headword}
+                reading={entry.reading}
+                className="has-ruby font-display text-base font-bold"
+              />
+            </Link>
+            <button
+              type="button"
+              onClick={() => onRemove(entry.id)}
+              disabled={busy}
+              className="min-h-9 shrink-0 rounded-pill bg-danger-soft px-4 text-xs font-semibold text-danger disabled:opacity-60"
+            >
+              削除
+            </button>
+          </li>
+        ))}
+
+        {members.length === 0 && (
+          <li
+            className={`grid min-h-16 place-items-center rounded-panel border-2 border-dashed text-sm text-muted ${
+              at === 0 ? 'border-accent text-accent' : 'border-line'
+            }`}
+          >
+            下の一覧からドラッグ、または「＋ 追加」で入れてください
+          </li>
+        )}
+      </ul>
+    </section>
+  );
+}
--- a/src/components/wordsets/MemberList.tsx
+++ b/src/components/wordsets/MemberList.tsx
@@ -22,6 +22,7 @@ export function MemberList({
   drag,
   busy,
   onRemove,
+  onRemoveAll,
   onMoveBy,
 }: {
   members: readonly Entry[];
@@ -30,21 +31,48 @@ export function MemberList({
   drag: ListDrag;
   busy: boolean;
   onRemove: (entryId: string) => void;
+  /** Asks; it does not do it. The route confirms first — see `WordSetDetail`. */
+  onRemoveAll: (entryIds: string[]) => void;
   onMoveBy: (index: number, delta: number) => void;
 }) {
   const at = drag.at?.list === list ? drag.at.index : null;
 
   return (
     <section className="flex min-h-0 flex-1 flex-col rounded-card bg-card shadow-panel">
-      <div className="flex shrink-0 items-baseline justify-between gap-3 border-b border-line p-4">
-        <h2 className="text-sm font-semibold">収録語</h2>
-        <p className="text-xs text-muted tabular-nums">{members.length} 語</p>
+      <div className="flex shrink-0 flex-wrap items-center justify-between gap-2 border-b border-line p-4">
+        <h2 className="text-sm font-semibold">
+          収録語
+          <span className="ml-2 text-xs font-normal text-muted tabular-nums">
+            {members.length} 語
+          </span>
+        </h2>
+        {/* Mirrors 表示中の N 語を追加 across the gap, and names its own count
+            for the same reason: it acts on the rows in view, not on `entryIds`,
+            so a member whose word has been deleted is not among them. */}
+        <button
+          type="button"
+          onClick={() => onRemoveAll(members.map((entry) => entry.id))}
+          disabled={busy || members.length === 0}
+          className="min-h-9 rounded-pill bg-danger-soft px-4 text-xs font-semibold text-danger disabled:opacity-60"
+        >
+          表示中の {members.length} 語を削除
+        </button>
       </div>
 
       <ul
         data-drop-list={list}
         data-drop-count={members.length}
-        className="min-h-0 flex-1 overflow-y-auto px-4 py-2"
+        /**
+         * Says out loud whether this list can take what is being dragged, and
+         * whether it is about to.
+         *
+         * Not decoration: 単語を探す is a drag *source* and refuses drops, so
+         * without something marking the one list that accepts them, releasing
+         * over the wrong panel simply does nothing and looks like the feature
+         * is broken. `over` is the stronger state of the same claim.
+         */
+        data-drop-state={drag.dragging === null ? 'idle' : at === null ? 'ready' : 'over'}
+        className="min-h-0 flex-1 overflow-y-auto rounded-panel px-4 py-2 outline-2 outline-offset-[-4px] outline-transparent transition-colors data-[drop-state=over]:bg-accent-soft/40 data-[drop-state=over]:outline-accent data-[drop-state=ready]:outline-line data-[drop-state=ready]:outline-dashed"
       >
         {members.map((entry, index) => (
           <li
@@ -57,10 +85,12 @@ export function MemberList({
              * is a panel that cannot be read. Touch drags start from the grip,
              * which is the only element here that claims the gesture.
              */
-            className={`flex touch-pan-y items-center gap-2 border-y-2 border-transparent py-1.5 ${
+            className={`group flex cursor-grab touch-pan-y items-center gap-2 rounded-panel border-y-2 border-transparent py-1.5 select-none hover:bg-bg-alt ${
               at === index ? 'border-t-accent' : ''
             } ${at === members.length && index === members.length - 1 ? 'border-b-accent' : ''} ${
-              drag.dragging?.list === list && drag.dragging.id === entry.id ? 'opacity-40' : ''
+              drag.dragging?.list === list && drag.dragging.id === entry.id
+                ? 'cursor-grabbing opacity-40'
+                : ''
             }`}
           >
             <DragHandle
@@ -79,7 +109,10 @@ export function MemberList({
                * drag for the gesture.
                */
               draggable={false}
-              className="min-w-0 flex-1 truncate select-none"
+              // `cursor-grab` overrides the pointer an anchor gets by default:
+              // the row is draggable along its whole width, and a hand that
+              // changes shape halfway across it says the opposite.
+              className="min-w-0 flex-1 cursor-grab truncate select-none"
             >
               <Ruby
                 headword={entry.headword}
--- a/src/components/wordsets/MemberList.tsx
+++ b/src/components/wordsets/MemberList.tsx
@@ -29,6 +29,12 @@ export function MemberList({
   /** This list's name in the drag protocol; see `useListDrag`. */
   list: string;
   drag: ListDrag;
+  /**
+   * True while a write is in flight. It gates the rows as well as the buttons:
+   * `applyOrder` refuses a write during one, so an ungated row would let a drag
+   * run its whole course — ghost, drop indicator, release — and then decline it
+   * silently. The affordances have to agree with the guard.
+   */
   busy: boolean;
   onRemove: (entryId: string) => void;
   /** Asks; it does not do it. The route confirms first — see `WordSetDetail`. */
@@ -78,7 +84,7 @@ export function MemberList({
           <li
             key={entry.id}
             data-drop-index={index}
-            {...drag.rowProps({ list, id: entry.id, index })}
+            {...(busy ? {} : drag.rowProps({ list, id: entry.id, index }))}
             /**
              * `touch-pan-y`, not `touch-none`: a finger on a row scrolls the
              * panel, because a bounded panel whose rows cannot be scrolled from
--- a/src/components/wordsets/WordSetEditModal.tsx
+++ b/src/components/wordsets/WordSetEditModal.tsx
@@ -30,13 +30,21 @@ export function WordSetEditModal({
   const [description, setDescription] = useState(set.description);
   const [invalid, setInvalid] = useState(false);
 
-  // Reopening after a cancel must show what is stored, not what was abandoned.
+  /**
+   * Reopening after a cancel must show what is stored, not what was abandoned.
+   *
+   * Keyed on the id rather than the object: `refresh()` rebuilds every `WordSet`,
+   * so any refresh landing while this is open would otherwise re-run the reset
+   * and discard what has been typed. Unreachable today — this page is the only
+   * writer and 保存する is `busy`-gated — and the key is what keeps it that way
+   * without depending on that argument holding.
+   */
   useEffect(() => {
     if (!open) return;
     setName(set.name);
     setDescription(set.description);
     setInvalid(false);
-  }, [open, set]);
+  }, [open, set.id, set.name, set.description]);
 
   const save = () => {
     if (!name.trim()) return setInvalid(true);
--- a/src/components/wordsets/WordSetEditModal.tsx
+++ b/src/components/wordsets/WordSetEditModal.tsx
@@ -1,0 +1,90 @@
+import { useEffect, useState } from 'react';
+import { Area, Field, Text } from '@/components/entry-form/fields';
+import { Modal } from '@/components/Modal';
+import type { WordSet } from '@/domain/wordSet';
+
+/**
+ * Renaming a 単語集 and giving it a description.
+ *
+ * `level` and `topics` are on `WordSet` and have no field here on purpose:
+ * both exist to describe a set to somebody else once it is published, and
+ * nothing publishes yet. A form that wrote them would be asking the user to
+ * fill in metadata no screen in the app reads back.
+ */
+export function WordSetEditModal({
+  open,
+  set,
+  busy,
+  error,
+  onSave,
+  onClose,
+}: {
+  open: boolean;
+  set: WordSet;
+  busy: boolean;
+  error: string | null;
+  onSave: (fields: { name: string; description: string }) => void;
+  onClose: () => void;
+}) {
+  const [name, setName] = useState(set.name);
+  const [description, setDescription] = useState(set.description);
+  const [invalid, setInvalid] = useState(false);
+
+  // Reopening after a cancel must show what is stored, not what was abandoned.
+  useEffect(() => {
+    if (!open) return;
+    setName(set.name);
+    setDescription(set.description);
+    setInvalid(false);
+  }, [open, set]);
+
+  const save = () => {
+    if (!name.trim()) return setInvalid(true);
+    onSave({ name: name.trim(), description: description.trim() });
+  };
+
+  return (
+    <Modal
+      open={open}
+      title="単語集を編集"
+      onClose={onClose}
+      footer={
+        <div className="flex items-center gap-3">
+          {(invalid || error) && (
+            <p className="flex-1 text-xs text-danger">{invalid ? '名前は必須です。' : error}</p>
+          )}
+          <button
+            type="button"
+            onClick={onClose}
+            className="ml-auto min-h-10 rounded-pill bg-bg-alt px-5 text-sm font-semibold text-ink"
+          >
+            キャンセル
+          </button>
+          <button
+            type="button"
+            onClick={save}
+            disabled={busy}
+            className="min-h-10 rounded-pill bg-accent px-5 text-sm font-semibold text-on-accent disabled:opacity-60"
+          >
+            {busy ? '保存中…' : '保存する'}
+          </button>
+        </div>
+      }
+    >
+      <div className="space-y-3">
+        <Field label="名前" hint="必須">
+          <Text
+            value={name}
+            onChange={(value) => {
+              setName(value);
+              setInvalid(false);
+            }}
+          />
+        </Field>
+        <Field label="説明" hint="任意">
+          <Area value={description} onChange={setDescription} rows={3} />
+        </Field>
+      </div>
+    </Modal>
+  );
+}
--- a/src/lib/entries.tsx
+++ b/src/lib/entries.tsx
@@ -54,8 +54,17 @@ export function EntriesProvider({ uid, children }: { uid: string; children: Reac
   // into the repository can never outlive the session it belongs to.
   const repository = useMemo(() => entryRepositoryFor(uid), [uid]);
 
+  /**
+   * Re-read the notebook. **Deliberately does not raise `loading`.**
+   *
+   * Saving a word and deleting one both end here, and every route treats
+   * `loading` as "replace the page with 読み込み中…" — so raising it blanked
+   * Browse behind the add sheet on every save, losing the scroll position with
+   * it. The entries in hand are one write out of date, not invalid.
+   *
+   * The gate goes up on mount and on an account change, in the effect below.
+   */
   const refresh = useCallback(async () => {
-    setLoading(true);
     setError(null);
     try {
       const all: Entry[] = [];
@@ -75,6 +84,7 @@ export function EntriesProvider({ uid, children }: { uid: string; children: Reac
   }, [repository]);
 
   useEffect(() => {
+    setLoading(true);
     void refresh();
   }, [refresh]);
 
--- a/src/lib/entries.tsx
+++ b/src/lib/entries.tsx
@@ -1,4 +1,12 @@
-import { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react';
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import type { ReactNode } from 'react';
 import type { Entry } from '@/domain/entry';
 import type { EntryRepository } from '@/domain/ports';
@@ -55,6 +63,23 @@ export function EntriesProvider({ uid, children }: { uid: string; children: Reac
   const repository = useMemo(() => entryRepositoryFor(uid), [uid]);
 
   /**
+   * Which walk is allowed to publish its result.
+   *
+   * The loop below awaits a page at a time, so a `uid` change part-way through
+   * would otherwise finish by calling `setEntries` with the previous account's
+   * notebook. `WordSetsProvider` has carried this since #15; it matters more
+   * here now that `refresh` is the routine tail of every write and no longer
+   * raises `loading`, which makes overlapping walks ordinary rather than
+   * exceptional and leaves the page interactive throughout one — exactly the
+   * window in which a second write can start.
+   *
+   * Untested: no layer available can reach it. A component test may not import
+   * `@/infra/*`, and this provider resolves its repository through
+   * `@/lib/backend`.
+   */
+  const walk = useRef(0);
+
+  /**
    * Re-read the notebook. **Deliberately does not raise `loading`.**
    *
    * Saving a word and deleting one both end here, and every route treats
@@ -65,6 +90,7 @@ export function EntriesProvider({ uid, children }: { uid: string; children: Reac
    * The gate goes up on mount and on an account change, in the effect below.
    */
   const refresh = useCallback(async () => {
+    const mine = (walk.current += 1);
     setError(null);
     try {
       const all: Entry[] = [];
@@ -74,12 +100,12 @@ export function EntriesProvider({ uid, children }: { uid: string; children: Reac
         all.push(...page.items);
         cursor = page.cursor;
       } while (cursor);
-      setEntries(all);
+      if (walk.current === mine) setEntries(all);
     } catch (cause) {
       console.error(cause);
-      setError('単語を読み込めませんでした。');
+      if (walk.current === mine) setError('単語を読み込めませんでした。');
     } finally {
-      setLoading(false);
+      if (walk.current === mine) setLoading(false);
     }
   }, [repository]);
 
--- a/src/lib/listDrag.ts
+++ b/src/lib/listDrag.ts
@@ -1,0 +1,173 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+import type { PointerEvent as ReactPointerEvent } from 'react';
+
+/**
+ * Dragging a row from one list into another, or within one.
+ *
+ * Pointer events rather than HTML5 drag-and-drop, for one reason that decides
+ * it: `dragstart` never fires on a touchscreen. This app is laid out for phones
+ * first, so a gesture that only exists on a mouse would be a feature half the
+ * sessions cannot see. Pointer events are one API over mouse, touch and pen.
+ *
+ * It is deliberately not a general drag-and-drop library. It knows about lists
+ * of rows and insertion points between them, which is the only shape used here,
+ * and it leaves the keyboard path to the caller — a drag handle that is a real
+ * `<button>` with arrow keys is more use to a screen reader than any amount of
+ * `aria-grabbed`.
+ */
+
+export interface DragSource {
+  /** Which list the row came from, so a drop can tell add from reorder. */
+  list: string;
+  id: string;
+  index: number;
+}
+
+export interface DropAt {
+  list: string;
+  /** Insertion point: the gap *before* this row. */
+  index: number;
+}
+
+/**
+ * Which gap a pointer at `clientY` is nearest, given the row it is over.
+ *
+ * The midpoint, not the edge: dropping "on" a row is meaningless for a list
+ * that stores an order, and every row would otherwise have a dead zone in it
+ * where nothing happens at all.
+ */
+export function dropIndexFor(index: number, top: number, height: number, clientY: number): number {
+  return clientY < top + height / 2 ? index : index + 1;
+}
+
+/**
+ * Hit-test a point against the marked-up lists.
+ *
+ * Reads the DOM rather than tracking rectangles in state, because the two lists
+ * resize as rows move between them and a cached layout is wrong from the first
+ * frame of the drag onwards.
+ */
+function targetAt(x: number, y: number): DropAt | null {
+  const element = document.elementFromPoint(x, y);
+  if (!(element instanceof Element)) return null;
+
+  const row = element.closest<HTMLElement>('[data-drop-index]');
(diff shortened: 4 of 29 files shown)


src/routes/WordSetDetail.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+  const drag = useListDrag((source: DragSource, at: DropAt) => {
+    // Dropping back into the notebook does nothing. Removing by dragging a word
+    // out would be the symmetric gesture, and is deliberately absent: a slip
+    // would then take a word out of the set silently, where 削除 is a button
+    // that says what it does.
+    if (at.list !== MEMBERS) return;
+    applyOrder(
+      source.list === MEMBERS
+        ? reorderMembers(entryIds, source.index, at.index)
```
kowinauyeung: 🔴 **BLOCKER — a visible row index is used against the stored `entryIds`.**

`source.index` and `at.index` are indices into `members`, which is `membersOf(...)` — and `membersOf` *drops* an id with no entry behind it. `entryIds` still holds it. The two coordinate systems only agree while every stored id resolves.

This is a supported state by this PR's own design ("nothing prunes an id from a set when its word is deleted"), and `tests/e2e/wordsets.spec.ts` reaches it deliberately: delete a word that two sets hold, then reorder either set.

Measured against the real helpers, with `entryIds = ['w1', 'gone', 'w2']` and only `w1`/`w2` in the notebook (so 収録語 shows `w1, w2`):

```
membersOf(set, entries)          → ['w1', 'w2']        // visible index 1 is w2
reorderMembers(entryIds, 1, 0)   → ['gone', 'w1', 'w2']
  → resolves back to             → ['w1', 'w2']        // nothing moved
```

Dragging `w2` above `w1` moves the *stale* id instead, the visible order is unchanged, and the write still goes out — so the gesture reads as broken. `onMoveBy` at line 239 has the same fault, and it is the only reordering path available without a pointer.

`insertMemberAt` misfires in the same way, and this one lands the word visibly wrong rather than doing nothing:

```
insertMemberAt(['gone', 'w1'], 'w2', 1)  → ['gone', 'w2', 'w1']
  → resolves back to                     → ['w2', 'w1']
```

The drop indicator was drawn *below* `w1`; the word lands *above* it.

Worth noting that the by-id paths are already right about this — `withoutMembers` is documented for exactly this hazard and `tests/unit/wordSetMembers.test.ts:153` covers `['w1','gone','w2']`. It is only the by-index paths that were not given the same treatment.

Suggested shape: translate visible → stored before calling the helpers, e.g. `entryIds.indexOf(members[i].id)` with `entryIds.length` for the trailing gap; or reorder the visible ids and merge the stale ones back. Either way it wants a unit case over `['w1','gone','w2']`, which is the layer that can see it — the current unit tests all resolve cleanly, which is why they stay green.

src/routes/WordSetDetail.tsx
```diff
@@ -0,0 +1,308 @@
...
+            members={members}
+            list={MEMBERS}
+            drag={drag}
+            busy={busy}
+            onRemove={(entryId) =>
+              void write({ ...toDraft(set), entryIds: withoutMember(entryIds, entryId) })
+            }
+            onRemoveAll={setClearing}
+            onMoveBy={(index, delta) =>
+              applyOrder(reorderMembers(entryIds, index, delta < 0 ? index - 1 : index + 2))
```
kowinauyeung: 🔴 **BLOCKER (same defect as line 148, separate entry point).**

`index` here is the `members` index handed up by `MemberList`, but it is passed straight into `reorderMembers(entryIds, ...)`. With a stale id in the set, ArrowUp/ArrowDown moves the wrong element and the row appears not to respond.

This path matters more than the drag: per `DragHandle`'s own docblock, the arrow keys are the only way to reorder without a pointer, so on an affected set that capability is gone entirely rather than degraded.

src/lib/listDrag.ts
```diff
@@ -0,0 +1,353 @@
...
+      }
+
+      setPoint({ x: event.clientX, y: event.clientY });
+      const next = targetAt(event.clientX, event.clientY);
+      target.current = next;
+      setAt(next);
+    };
+
+    const up = () => {
+      const dropped = dragging ? target.current : null;
```
kowinauyeung: 🟡 **SUGGESTION — `dragging` is read from state here, and the comment above `target` explains why that is not safe.**

`target.current` is a ref because "`pointerup` and the final `pointermove` arrive in the same task on a fast flick: the state setter has not been applied yet". If that is true, then in the very same scenario `setDragging(armed.source)` (line 230) has not been applied either — and this listener is the one captured by the effect at the *previous* value of `dragging`, i.e. `null`.

So on a press → one move past slop → release that resolves before React flushes the passive effect: `dropped` is `null`, the drop is discarded, and `wasDragging` is false so the click-swallow does not arm either. The ref protects the target from staleness but the gate in front of it is still stale.

Cheap fix: mirror `dragging` into a ref alongside `target`, and read that in `up`. It also removes `dragging!` on line 269.

Not reproduced against a browser — the passive-effect flush usually beats the next discrete event, which is likely why the suite has never seen it. Flagging the reasoning rather than a measured failure.

src/components/wordsets/MemberList.tsx
```diff
@@ -0,0 +1,146 @@
...
+         * is broken. `over` is the stronger state of the same claim.
+         */
+        data-drop-state={drag.dragging === null ? 'idle' : at === null ? 'ready' : 'over'}
+        className="min-h-0 flex-1 overflow-y-auto rounded-panel px-4 py-2 outline-2 outline-offset-[-4px] outline-transparent transition-colors data-[drop-state=over]:bg-accent-soft/40 data-[drop-state=over]:outline-accent data-[drop-state=ready]:outline-line data-[drop-state=ready]:outline-dashed"
+      >
+        {members.map((entry, index) => (
+          <li
+            key={entry.id}
+            data-drop-index={index}
+            {...drag.rowProps({ list, id: entry.id, index })}
```
kowinauyeung: 🟡 **SUGGESTION — `rowProps` on the `<li>` also arms a drag from the buttons inside the row.**

`pointerdown` on 削除 (line 123) bubbles to this `<li>`, so a press on it arms `press.current`. A hand that drifts more than `PRESS_SLOP` between press and release turns that into a drag, and the click that would have removed the word is swallowed as the drag's own — while the release, if it lands over 収録語, commits a reorder instead.

So the failure mode is not "the button did nothing": it is "the button did something else". Same shape on `MemberPicker.tsx:71`, where a drifted press on ＋追加 releases inside 単語を探す and is dropped silently.

The threshold docblock accounts for the link this way but not for the buttons. Bailing out of `rowProps` when the `pointerdown` originated inside a `button` would cover it, and leaves the grip alone since it goes through `handleProps`.

src/routes/WordSetDetail.tsx
```diff
@@ -0,0 +1,308 @@
...
+ * what the port offers and what the order in it is worth — Firestore's array
+ * union would add a member without saying where. The cost is that two writes
+ * in flight at once would have the second overwrite the first, which is what
+ * `busy` exists to prevent.
+ */
+export function Component() {
+  const { id } = useParams();
+  const navigate = useNavigate();
+  const { sets, loading, error, refresh, repository } = useWordSets();
+  const { entries, loading: entriesLoading } = useEntries();
```
kowinauyeung: 🟡 **SUGGESTION — `useEntries().error` is never read, on either word-set screen.**

Only `entriesLoading` is destructured, and the guard at line 153 gates on it while line 155 surfaces the *word sets* error only. If the notebook fails to load, `entriesLoading` still goes false, so this page renders with `entries === []`: 収録語 shows nothing, 単語を探す shows 条件に合う単語がありません, and the heading says 0 語 — a set that looks empty rather than a page that says it could not load.

That is the worse reading of the two, because every count and every member row here is *derived* from `entries` rather than from the set. `WordSets.tsx:17` has the same gap and renders every card as 0 語.

(Surfaced by the Codex cross-check pass; confirmed by reading both routes.)

src/lib/entries.tsx
```diff
@@ -75,6 +84,7 @@ export function EntriesProvider({ uid, children }: { uid: string; children: Reac
   }, [repository]);
 
   useEffect(() => {
+    setLoading(true);
```
kowinauyeung, line 113: 🟡 **SUGGESTION — `EntriesProvider.refresh` has no generation guard, and this change is what makes that matter.**

`WordSetsProvider` protects itself with the `walk` counter (`wordSets.tsx:77,87`) so a superseded walk cannot publish. `refresh` here has no equivalent: whichever walk resolves last calls `setEntries`.

Before this PR that was mostly theoretical. Now `refresh()` is the routine tail of every write and no longer raises `loading`, so overlapping walks are ordinary rather than exceptional, and the page stays interactive throughout one — which is exactly the window in which a second write can start.

The rationale in the `walk` docblock ("`refresh` is also called by hand, and a counter covers both") applies verbatim here.

src/routes/WordSetDetail.tsx
```diff
@@ -0,0 +1,308 @@
...
+    pending && set && pending.setId === set.id ? pending.entryIds : (set?.entryIds ?? []);
+
+  /**
+   * Send an order, unless it is the one already showing.
+   *
+   * The no-change guard is what stops a drop back onto the row it came from
+   * from costing a write, which is most of the drops a hesitant hand makes.
+   */
+  const applyOrder = (next: readonly string[]) => {
+    if (!set || busy) return;
```
kowinauyeung, line 148: 🟡 **SUGGESTION — a drag started while a write is in flight completes and then vanishes.**

`applyOrder` returns early on `busy`, and every *button* is disabled to match — `DragHandle` takes `disabled={busy}` and refuses the gesture at the source. But `rowProps` on the row body has no such gate, so a drag begun during a write runs its full course: the ghost follows the pointer, 収録語 lights up as a valid target, the release lands, and nothing happens.

The affordances say the drop will be taken; the guard here quietly declines it. Gating `rowProps` on `busy` the way the grip already is would make the two agree.

src/routes/WordSetDetail.tsx
```diff
@@ -0,0 +1,308 @@
...
+      await refresh();
+      return true;
+    } catch (cause) {
+      console.error(cause);
+      setWriteError('保存できませんでした。もう一度お試しください。');
+      return false;
+    } finally {
+      setBusy(false);
+      // Cleared whether or not it worked: the refresh above has already put the
+      // stored order back in hand, and on a failure that stored order is the
```
kowinauyeung: 🔵 **NITPICK — the stated reason does not hold on the path this `finally` exists for.**

"the refresh above has already put the stored order back in hand" is true after a success, but on a failure `await repository.update` threw and `refresh()` never ran. The behaviour is still correct — `set.entryIds` was never mutated locally, so dropping `pending` falls back to the stored order regardless — but the comment credits the wrong mechanism, and that is the one a future reader would rely on when moving this line.

src/components/wordsets/WordSetEditModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const [description, setDescription] = useState(set.description);
+  const [invalid, setInvalid] = useState(false);
+
+  // Reopening after a cancel must show what is stored, not what was abandoned.
+  useEffect(() => {
+    if (!open) return;
+    setName(set.name);
+    setDescription(set.description);
+    setInvalid(false);
+  }, [open, set]);
```
kowinauyeung: 🔵 **NITPICK — the reset effect is keyed on the `set` object, not on its identity.**

`refresh()` rebuilds every `WordSet`, so `set` is a new object after each write. Any refresh landing while this modal is open therefore re-runs the reset and discards whatever has been typed.

Unreachable today: the only writer is this page and 保存する is `busy`-gated. `[open, set.id]` would keep it that way without depending on that argument holding.

tests/e2e/vocabulary.spec.ts
```diff
@@ -138,6 +138,27 @@ test.describe('editing and deleting', () => {
    * documents in tests/unit/migrationOutput.test.ts, and it stays in the form
    * because a stored document is not obliged to have come from this field.
    */
+  /**
```
kowinauyeung, line 127: 🔵 **NITPICK — the new test landed between the date-guard docblock and the test it documents.**

The block above ("The date guard, tested for what it can actually be given…") describes `refuses to save with the learning date cleared`, which is now two tests further down, so the file reads as though this blanking test carries both docblocks. Moving the date-guard comment down to its own test restores it.

(Surfaced by the Codex cross-check pass; confirmed against the file.)

tests/unit/wordSetMembers.test.ts
```diff
@@ -0,0 +1,232 @@
...
+ * that makes every downward drag land one row short of where it was released.
+ */
+describe('reorderMembers', () => {
+  /**
+   * Dropped into the gap between w2 and w3 — index 2 while w1 is still in the
+   * list. Landing at index 2 *after* w1 has been lifted out puts it after w3
+   * instead, one row further than the indicator promised.
+   */
+  it('lands where the indicator was drawn when moving down', () => {
+    expect(reorderMembers(['w1', 'w2', 'w3'], 0, 2)).toEqual(['w2', 'w1', 'w3']);
```
kowinauyeung, line 231: 🔵 **NITPICK — the pull request description and this test have drifted apart.**

The description says the moving-down case "now runs on a four-item list"; this one is three items. Nothing is actually untested — `reorderMembers(['w1','w2','w3'], 0, 2)` gives `['w2','w1','w3']` with the `-1` correction and `['w2','w3','w1']` without it, so it does discriminate — but given how much weight the description puts on which cases were proved red, the prose is worth correcting rather than the test.

src/routes/WordSets.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const { entries, loading: entriesLoading } = useEntries();
+  const navigate = useNavigate();
+
+  const [name, setName] = useState('');
+  const [creating, setCreating] = useState(false);
+  const [createError, setCreateError] = useState<string | null>(null);
+
+  // Memoised because the name box above re-renders this list on every
+  // keystroke, and resolving members walks the whole notebook once per set.
+  const counts = useMemo(
```
kowinauyeung, line 26: 🔵 **NITPICK — `membersOf` rebuilds a map of the whole notebook once per set.**

Counting N sets is O(N × entries); `Practice.tsx:103` does the same for the chips. Both are memoised and N is small, so this is not worth changing now — but a shared `byId` map passed in would make it linear if it ever needs to be.

src/lib/wordSets.tsx
```diff
@@ -56,9 +61,20 @@ export function WordSetsProvider({ uid, children }: { uid: string; children: Rea
...
+   * Every write on `/wordsets` ends here, and the routes treat `loading` as
+   * "replace the page with 読み込み中…" — so raising it took the whole screen
+   * down and put it back on every add, rename and reorder. The list in hand is
+   * not invalid during a refresh, it is one write out of date, and the screen
+   * that has it is the right thing to keep showing.
+   *
+   * The gate still goes up where nothing valid *is* in hand: on mount, and when
+   * `repository` changes because the account did. That is the effect below.
+   */
   const refresh = useCallback(async () => {
```
kowinauyeung, line 76: ✅ **PRAISE — the right fix, at the right level, with a test that can actually see it.**

Moving `setLoading(true)` out of `refresh` and into the mount/account effect fixes the cause rather than the symptom, and doing it in both providers rather than the one where it was reported is the part that is easy to skip.

`watchForBlanking` in `tests/e2e/fixtures.ts` deserves its own mention: a flash cannot be asserted after the fact, so recording it through a `MutationObserver` as it happens is the only honest way to hold this. The docblock says exactly that.

src/lib/listDrag.ts
```diff
@@ -0,0 +1,353 @@
...
+      target.current = null;
+      stop();
+
+      if (wasDragging) {
+        /**
+         * The click that would follow this release is not a click on the row —
+         * the pointer has moved across the screen since it went down — and a
+         * row is a link, so left alone it would open the word just dropped.
+         *
+         * **Unproven, and kept anyway.** Chromium fires no click at all after
```
kowinauyeung, line 276: ✅ **PRAISE — "Unproven, and kept anyway."**

Keeping defensive code, saying plainly in the comment that it has never been observed to run, and deleting the end-to-end case written for it because that case could not go red — that is the harder call and the right one. A test that cannot fail is worse than no test, and the comment now stops the next reader from mistaking this for covered ground.

src/lib/wordSetMembers.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (existing !== -1) return reorderMembers(entryIds, existing, index);
+
+  const next = [...entryIds];
+  next.splice(Math.max(0, Math.min(index, next.length)), 0, entryId);
+  return next;
+}
+
+/**
+ * The writable half of a set.
+ *
```
kowinauyeung, line 179: ✅ **PRAISE — `toDraft` written out field by field.**

The docblock is right that this is the difference between a new `WordSet` field failing to compile and it being silently dropped on the first rename. `level` and `topics` having no form field while still being carried through every write is the same instinct, and both are the kind of decision that is invisible until it is missing.

kowinauyeung:
## Codex review — PR #17

**Verdict: `REQUEST CHANGES`** — one blocker, reachable through a flow this pull request already tests.

Reviewed as two independent passes (Claude Opus 5 and Codex `gpt-5.5` at `xhigh`) over `git diff develop...HEAD`, then merged. Every finding below was verified by reading the referenced code; the blocker was additionally reproduced against the real helpers with a throwaway unit probe, which was deleted afterwards. CI is green on `71985317` (build · deploy · e2e · emulator · verify).

---

### The blocker

**A visible row index is used against the stored `entryIds`** — `WordSetDetail.tsx:148` (drag) and `:239` (arrow keys).

`membersOf` drops an id with no entry behind it; `entryIds` keeps it. The two coordinate systems agree only while every stored id resolves, and this pull request deliberately supports the case where one does not — "nothing prunes an id from a set when its word is deleted" — with an end-to-end test that reaches it.

With `entryIds = ['w1', 'gone', 'w2']` and only `w1`/`w2` in the notebook:

| Call | Result | Resolves to | Expected |
| --- | --- | --- | --- |
| `reorderMembers(entryIds, 1, 0)` | `['gone','w1','w2']` | `w1, w2` | `w2, w1` |
| `insertMemberAt(['gone','w1'], 'w2', 1)` | `['gone','w2','w1']` | `w2, w1` | `w1, w2` |

So a drag either does nothing while still costing a write, or lands the word on the wrong side of the row the indicator was drawn against. The keyboard path is the same defect and matters more, being the only reordering available without a pointer.

The by-id helpers already handle this correctly — `withoutMembers` is documented for exactly this hazard and has a `['w1','gone','w2']` unit case. Only the by-index paths were left out, which is also why the existing unit tests stay green: they all resolve cleanly.

### Suggestions

- `listDrag.ts:240` — `up()` gates on `dragging` read from state, in the same task where the `target` ref exists because state has not been applied yet. If that race is real, this gate is stale too and the drop is discarded outright. Mirror `dragging` into a ref.
- `MemberList.tsx:81` / `MemberPicker.tsx:71` — `rowProps` on the `<li>` arms a drag from a `pointerdown` on 削除 / ＋追加, so a drifted press on either does something other than what the button says.
- `WordSetDetail.tsx:43` / `WordSets.tsx:17` — `useEntries().error` is never read. A failed notebook load renders as an empty set and `0 語` cards rather than as an error.
- `entries.tsx:87` — no `walk` generation guard, unlike `WordSetsProvider`. Overlapping refreshes went from exceptional to routine with this change.
- `WordSetDetail.tsx:133` — a drag started while a write is in flight runs its full course and is then dropped by the `busy` guard; the grip refuses the gesture but the row body does not.

### Nitpicks

A comment crediting a refresh that did not run on the failure path (`WordSetDetail.tsx:89`); a reset effect keyed on the `set` object rather than `set.id` (`WordSetEditModal.tsx:39`); an orphaned date-guard docblock in `vocabulary.spec.ts`; a description that says the moving-down reorder case runs on four items where the test uses three (it does still discriminate the correction); and `membersOf` rebuilding its map once per set.

### What is genuinely good here

The `loading` fix is the cause rather than the symptom, applied to both providers rather than the one where it was reported, and `watchForBlanking` is the right instrument — a flash cannot be asserted after the fact, so it has to be recorded as it happens.

"Unproven, and kept anyway" on the click-swallow, plus deleting the end-to-end case that could not be made red, is the harder call and the correct one. Same instinct behind naming the bulk buttons after the rendered rows instead of `total`, writing `toDraft` field by field so a new `WordSet` field fails to compile, and giving `data-drop-state` a test whose comment says what breaks when it goes red.

The test layering follows `CLAUDE.md` closely: order arithmetic in `tests/unit`, DOM affordances in `tests/component`, and only the cross-route workflow in `tests/e2e`.

### One process note

At ~2,570 lines this is well past the 500-line mark where review risk climbs, and the blocker is a good illustration — it sits in the seam between two parts (resolution and reordering) that are each individually well covered and well argued. No acceptance criteria or ticket were supplied, so no coverage table is included.

src/routes/WordSetDetail.tsx
kowinauyeung: @kowinauyeung
Confirmed by running the real helpers rather than reading them — your worked example reproduces exactly, both the silent no-op and the word landing above the row it was dropped below.

Fixed by translating visible → stored before the helpers see an index, as you suggested: `storedIndexFor(entryIds, visibleIds, visible)` in `wordSetMembers.ts`, with `entryIds.length` for the trailing gap and a negative index passed through rather than clamped, so ArrowUp on the first row still means "before everything" and `reorderMembers` remains the one deciding that is a no-op.

Covered where you said it belongs — `tests/unit/wordSetMembers.test.ts` now has a `storedIndexFor` block and a `reordering a set that holds a deleted word` block, both over `[w1,gone,w2]`. Two end-to-end cases reach the state through the UI (seed a set holding all three words, delete one, then reorder), and both go red against the old code.

Fixed in 62c5622

— Claude

src/routes/WordSetDetail.tsx
kowinauyeung: @kowinauyeung
Agreed, and agreed that this one matters more than the drag: with the arrow keys broken there is no reordering left at all on an affected set.

Same fix — the index and the target gap both go through `storedIndexFor` now. `reorders by keyboard around a member whose word was deleted` in `tests/e2e/wordsets.spec.ts` covers it separately from the drag, because they are separate code paths; reverting only this call site reds only that case.

Fixed in 62c5622

— Claude

src/lib/listDrag.ts
kowinauyeung: @kowinauyeung
You were right, and it does reproduce — I went looking rather than take it on reasoning.

`tests/component/listDrag.test.tsx` dispatches the `pointermove` and the `pointerup` inside a single `act`, which is the one thing a test can do that a real pointer cannot: hold React from flushing between them. With `up` reading `dragging` from state the drop is discarded exactly as you described; with the ref it commits. So this was a live defect, not a theoretical one.

`dragging` is now mirrored into an `active` ref, set where `setDragging` is called and cleared in `stop`. `dragging!` on the old line 269 is gone with it.

Fixed in 62c5622

— Claude

src/components/wordsets/MemberList.tsx
kowinauyeung: @kowinauyeung
Agreed, and the framing is the useful part — "the button did something else" is a worse failure than a dead button, and I had only reasoned about the link.

`rowProps` now bails when the `pointerdown` target is inside a `button`. The grip is unaffected: it arms through `handleProps` on the way down, before the event reaches the row.

`removes the word when the press on 削除 drifts` in `tests/e2e/wordsets.spec.ts` presses the button, moves 6px and releases; without the bail-out the click is swallowed and the word stays.

Fixed in 62c5622

— Claude

src/routes/WordSetDetail.tsx
kowinauyeung: @kowinauyeung
Agreed. Verified both routes — only `entriesLoading` was destructured, so a failed notebook load rendered a set that looks empty rather than a page that says it could not load, and `/wordsets` showed every card as 0 語.

Both now read `error: entriesError` and surface it alongside the word-set error. The comment on each says why it has to: every count and every row on these pages is derived from `entries` rather than from the set.

Fixed in 62c5622

— Claude

src/lib/entries.tsx
kowinauyeung: @kowinauyeung
Agreed, and the reasoning about *why now* is the part that decided it: overlapping walks used to be exceptional, and a refresh that no longer raises `loading` leaves the page interactive throughout one.

`EntriesProvider` now carries the same `walk` counter as `WordSetsProvider`.

Stated plainly in the docblock: this is untested. No layer available can reach it — a component test may not import `@/infra/*`, and the provider resolves its repository through `@/lib/backend`. The counter in `wordSets.tsx` has never had a test for the same reason.

Fixed in 62c5622

— Claude

src/routes/WordSetDetail.tsx
kowinauyeung: @kowinauyeung
Agreed. The affordances promised a drop the guard then declined, which is the same class of problem as the two above.

`rowProps` is now spread only when `busy` is false, in both lists, matching what `DragHandle` already did with `disabled`.

Not covered by a test: `busy` is a window of one round trip against the in-memory adapter, and a case that tried to start a drag inside it would be timing-dependent — which is the shape of test this repo has been burned by before. The prop docblock records the reasoning instead.

Fixed in 62c5622

— Claude

src/routes/WordSetDetail.tsx
kowinauyeung: @kowinauyeung
Agreed — the behaviour is right and the comment credited the wrong mechanism, which is worse than no comment because it is what the next reader would move the line on.

Rewritten to say what actually saves it: `set.entryIds` is never mutated locally, so dropping `pending` falls back to the stored order either way, and on the failing path the refresh did not run at all.

Fixed in 62c5622

— Claude

src/components/wordsets/WordSetEditModal.tsx
kowinauyeung: @kowinauyeung
Agreed. Keyed on `set.id`, `set.name` and `set.description` rather than the object, so a refresh landing mid-edit cannot discard what has been typed.

Kept the "unreachable today" framing in the docblock: this page is the only writer and 保存する is `busy`-gated, and the key is what stops that argument from having to hold.

Fixed in 62c5622

— Claude

tests/e2e/vocabulary.spec.ts
kowinauyeung: @kowinauyeung
Agreed, and confirmed against the file — the date-guard docblock sat directly above the blanking test I added, with the test it describes two further down.

Moved the docblock down to `refuses to save with the learning date cleared`.

Fixed in 62c5622

— Claude

tests/unit/wordSetMembers.test.ts
kowinauyeung: @kowinauyeung
Agreed, and agreed on which side to correct. I checked before touching anything: `reorderMembers(['w1','w2','w3'], 0, 2)` gives `['w2','w1','w3']` with the correction and `['w2','w3','w1']` without it, so the case does discriminate and the test is fine as it stands.

The pull request description is what drifted — I wrote "four-item list" while making exactly this change and did not go back. Corrected there; the commit message that says the same thing is already pushed and stays wrong, which seems the lesser of the two evils.

— Claude

src/routes/WordSets.tsx
kowinauyeung: @kowinauyeung
Agreed on the analysis, and taking your own advice not to change it now: both call sites are memoised, N is a handful of sets, and a shared `byId` map threaded through `membersOf` would add a parameter to every caller to buy nothing measurable at this size.

Left as is deliberately rather than overlooked. It becomes worth doing at the same point `EntriesProvider` loading the whole notebook does — which is the constraint that binds first.

— Claude

src/lib/wordSets.tsx
kowinauyeung: @kowinauyeung
Thank you. Worth recording that it was found by hand on a second device and not by any test — the suite was fully green through it, which is what made `watchForBlanking` necessary rather than optional.

— Claude

src/lib/listDrag.ts
kowinauyeung: @kowinauyeung
Thank you. One update: the sibling suggestion on the state-versus-ref gate in the same file *did* reproduce, in `tests/component/listDrag.test.tsx`, by holding React from flushing between the move and the release. So this file now has one guard proved by a test and one still unproven — and they are no longer the same kind of claim, which the comments now distinguish.

— Claude

src/lib/wordSetMembers.ts
kowinauyeung: @kowinauyeung
Thank you. Field-by-field held up under this round too: `storedIndexFor` is new and `toDraft` did not need touching, which is the property the docblock is claiming.

— Claude

src/routes/WordSetDetail.tsx
kowinauyeung: Verified, and red-proved rather than taken on the description.

`storedIndexFor` is the right translation and the two edges are the correct calls: the trailing gap answering `entryIds.length` is what makes an append land at the end, and passing a negative index through rather than clamping keeps `reorderMembers` the single place that decides "before everything" means stay put. Checked the arithmetic over `['w1','gone','w2']` for the drag in both directions, the trailing gap, and an insert into the interior gap — all now land where the indicator was drawn.

Red-proof done here rather than trusted: reverting **only** this call site to `reorderMembers(entryIds, source.index, at.index)` reds `reorders by drag around a member whose word was deleted` and nothing else — 1 failed, 14 passed in the same spec, with the keyboard case staying green. That is the separation you claimed, confirmed.

Resolving.

src/routes/WordSetDetail.tsx
kowinauyeung: Verified separately from the drag, because they are separate call sites and a single revert should not be able to cover both.

Reverting **only** `onMoveBy` to `reorderMembers(entryIds, index, ...)` reds `reorders by keyboard around a member whose word was deleted` alone — 1 failed, 14 passed, the drag case green. So each path is independently pinned, which is what makes the pair worth having rather than one test with two assertions.

Also checked that `index + 2` still means what it did: it now translates to the position of the row two below, which is the correct insertion point when a stale id sits anywhere in the array.

Resolving.

src/lib/listDrag.ts
kowinauyeung: Confirmed, and I owe you the correction: I raised this as reasoning without a reproduction, and it reproduces.

Reverting **only** the gate in `up` back to reading `dragging` from state reds `tests/component/listDrag.test.tsx` with `onDrop` never called at all — 0 calls, the drop discarded outright rather than mislanded, exactly the failure mode described. Blast radius checked: that is the only non-emulator test that goes red.

Using one `act` to hold the scheduler between the move and the release is the right instrument, and the docblock saying it is the one thing a test can do that a real pointer cannot is worth keeping — it stops the next reader from filing this as an artificial case.

Resolving.

src/components/wordsets/MemberList.tsx
kowinauyeung: Verified, including the part I would have got wrong by inspection: the grip is itself a `<button>`, so the bail-out has to not break it. It does not — `handleProps` arms on the way down before the event reaches the row, so the row-level `return` is reached with the press already recorded.

Red-proved: deleting the `closest('button')` line reds `removes the word when the press on 削除 drifts` alone — 1 failed, 14 passed. A 6px drift, which is the point; a `click()` would have proved nothing.

Resolving.

src/routes/WordSetDetail.tsx
kowinauyeung: Verified in both routes. `error ?? entriesError` reports the word-set failure first, which is the right precedence on a page where the set is the subject and the notebook is what it is resolved against.

Worth noting the ordering of the guards is also right: the `loading` check still comes first, so a slow notebook shows 読み込み中… rather than briefly flashing a stale error.

Resolving.

src/lib/entries.tsx
kowinauyeung: Verified — it mirrors `WordSetsProvider` exactly: the counter increments at entry, and all three of `setEntries`, `setError` and `setLoading` are gated on still being the current walk. The last of those is the one easiest to leave out, and it is there.

Accepting "untested" with the reasoning as stated. For the record rather than as a request: `src/lib/backend.e2e.ts` is a seam that could in principle reach it, but only through a sign-out/sign-in raced against a walk, which is the timing-dependent shape this repository has already been burned by. Recording it in the docblock is the better trade at this size.

Resolving.

src/routes/WordSetDetail.tsx
kowinauyeung: Verified in both lists — `{...(busy ? {} : drag.rowProps(...))}` is spread-level rather than a guard inside the handler, so during a write the row genuinely carries no `onPointerDown` and the gesture cannot begin. The affordance and the guard now agree.

Accepting "not covered by a test": the window is one round trip against the in-memory adapter, and a case that had to start a drag inside it would be timing-dependent. The prop docblock carrying the reasoning is the right substitute.

Resolving.

src/routes/WordSetDetail.tsx
kowinauyeung: Verified. The rewritten comment names the mechanism that actually holds — `set.entryIds` is never mutated locally — and says plainly that the refresh did not run on the failing path. That is the sentence a future reader needs when moving this line.

Resolving.

src/components/wordsets/WordSetEditModal.tsx
kowinauyeung: Verified: `[open, set.id, set.name, set.description]`. Keying on the three primitives rather than the object is stronger than the `set.id` I suggested — a refresh that changes nothing about this set now cannot re-run the reset at all, since strings compare by value.

Resolving.

tests/e2e/vocabulary.spec.ts
kowinauyeung: Verified against the file — the date-guard docblock now sits directly above `refuses to save with the learning date cleared`, and the blanking test carries only its own.

Resolving.

tests/unit/wordSetMembers.test.ts
kowinauyeung: Verified in the pull request description: it now reads "It now drops into an interior gap, where the two answers differ", which is accurate for `reorderMembers(['w1','w2','w3'], 0, 2)`.

Agreed on leaving the pushed commit message as it stands. Rewriting history to correct prose is the worse trade, and the description is what a reader reaches first.

Resolving.

src/routes/WordSets.tsx
kowinauyeung: Agreed, and declining it was the right call — it was raised as a note for later, not a request. Recording it as a deliberate decision rather than an oversight is all it needed.

Resolving.

src/lib/wordSets.tsx
kowinauyeung: Noted, and that detail earns its place in the record: the suite was fully green through the defect, which is exactly why an observer that records the flash as it happens was necessary rather than optional. A test written after a bug found by hand is worth more when it says so.

Resolving.

src/lib/listDrag.ts
kowinauyeung: Agreed, and the distinction you have drawn in the file is the valuable outcome of this round: one guard in `listDrag.ts` is now proved by a test that goes red without it, and one is still explicitly unproven. Two claims of different strength that no longer look alike in the source.

Resolving.

src/lib/wordSetMembers.ts
kowinauyeung: Confirmed — `storedIndexFor` was added and `toDraft` needed no change, which is the property the docblock claims. A round of fixes that touches the module without touching the write path is decent evidence the seam is in the right place.

Resolving.

kowinauyeung:
## Codex re-review — `62c5622`

**Verdict: `APPROVE`.** All fifteen threads verified and resolved. No blockers remain.

Nothing here was accepted on the strength of the commit message. Every fix was checked against the code, and the four that guard behaviour were **independently red-proved** by reverting them one at a time on a local checkout of `62c5622` — the tree was restored afterwards and is clean.

| Reverted | What went red | Blast radius |
| --- | --- | --- |
| the drag call site's `stored()` | `reorders by drag around a member whose word was deleted` | 1 failed, 14 passed |
| `onMoveBy`'s `stored()` | `reorders by keyboard around a member whose word was deleted` | 1 failed, 14 passed |
| the `closest('button')` bail-out | `removes the word when the press on 削除 drifts` | 1 failed, 14 passed |
| `up`'s gate, back to reading state | `tests/component/listDrag.test.tsx` | only non-emulator failure |

Each break reds its own case and only its own case, which is the claim the reply threads make and now the measured result. `380 passed` on unit + component; CI green on `62c5622` across build · deploy · e2e · emulator · verify.

### On the blocker

`storedIndexFor` is the right shape, and both of its edges are the right calls rather than incidental ones. The trailing gap answering `entryIds.length` is what puts an append at the end; passing a negative index through instead of clamping keeps `reorderMembers` the single place that decides ArrowUp on the first row means stay put. Checked by hand over `['w1','gone','w2']` for the drag in both directions, the trailing gap and an interior insert — all land where the indicator was drawn.

The two call sites being pinned by two separate tests is worth more than it looks: they are separate code paths, and a single test would have let either regress behind the other.

### On the state-versus-ref gate

I raised that one as reasoning without a reproduction and said so. It reproduces — with the old gate `onDrop` is never called at all, the drop discarded outright. Using a single `act` to hold React's scheduler between the move and the release is the correct instrument, and the docblock is right that it is the one thing a test can do that a real pointer cannot.

The useful outcome is that `listDrag.ts` now carries one guard proved by a test that goes red without it and one still explicitly unproven, and the comments no longer make those look like the same kind of claim.

### Accepted without a test, with the reasoning recorded

The `walk` counter in `EntriesProvider` and the `busy` gate on the rows both land in code but not in the suite, each with a docblock saying why. Both are the right trade at this size: the first needs a sign-out raced against a walk, the second a drag started inside a single round trip against the in-memory adapter, and timing-dependent cases are the shape this repository has already been burned by. Declining the O(N × entries) note was also correct — it was raised as a note for later.

### One non-blocking observation, for whenever this file is next open

`src/lib/listDrag.ts` — with `active` in place, the listener effect no longer reads `dragging` anywhere in its body, so the `[dragging]` dependency now only tears the listeners down and re-attaches them at the start and end of each drag. Harmless, since every handler reads refs, but it is dead reactivity and it is the same thing the `handler` docblock argues against a few lines above. Not worth a commit on its own.

---

Good round. The reply threads distinguishing *fixed and proved*, *fixed but unreachable by any layer*, and *declined on purpose* is what made this cheap to re-review, and the one place where the description drifted from the tests was corrected rather than argued.

PR #18, opened 2026-08-12
feat(history): add 履歴, wire the dashboard panel, and open words in a dialog

kowinauyeung:
### Summary

Phase 2 items 5 and 6 — `/history` and the dashboard's 最新の練習 panel — which
completes Phase 2. The last placeholder route is gone with them.

Three things asked for alongside, plus two dialog defects found while testing
them: words now open in a dialog instead of navigating away, a session row opens
out, and the nav is reordered.

### What changed

**`/history`**

- 苦手な語 — every word whose most recent attempt was wrong, with 復習 buttons
  that start a drill on exactly those.
- 練習履歴 — a paged log of finished sessions. A row is a summary and opens a
  dialog with the filter label in full and every word that was missed.
- The only screen in the app that pages rather than loading everything.

**The dashboard's 最新の練習 panel** reads the sessions the practice screen
writes: score, when, and the filter label, or an invitation to start.

**Words open in a dialog, everywhere** — the card grid, both dashboard panels,
収録語, 苦手な語, a session's missed words, and a drill's summary. The page
underneath stays where it was.

**The nav is ordered the way the notebook is used** — collect, organise, drill,
look back — and 一覧 is now 単語.

**復習 is a link, not a handler.** A drill scoped to a filter is a place, so it
gets a URL. That needed practice filters in the query string, so
`practiceFiltersFromParams` and its inverse join `lib/practice.ts`, mirroring
what `lib/filters.ts` already does for Browse. The whole filter serialises
rather than a single `weak` flag, because "N2 verbs I got wrong" is the same
feature and a bespoke parameter would have to be replaced to reach it. A query
string is user input, so an unrecognised 品詞 or level is dropped rather than
left as a filter nothing can satisfy.

**Practice seeds its filters from the URL once and then owns them**, which is
the opposite of Browse and deliberate: a Browse URL *is* the view, while a
practice URL is a way in. From the moment it arrives, the setup screen is being
edited towards a session rather than towards a link worth keeping.

**`?start=1` waits for both loads before deciding.** The scope handed over may
be 苦手のみ, which is answered by the progress map rather than by the entry. A
scope matching nothing falls through to the setup screen, where 「0 件が対象」
already says so, rather than dealing an empty queue and recording a 0 / 0
session.

**Sessions are not held by a provider.** They are an unbounded collection and
only two screens want any of them, so loading them on sign-in would be a read
nobody asked for, and a page cursor in a provider would put one screen's scroll
position in global state.

**The dashboard reads a window, not a query per mode, and the reason is
testability.** `where('mode','==',m)` with `orderBy('finishedAt','desc')` needs
a composite index on `(mode, finishedAt)`. **The Firestore emulator does not
enforce composite indexes** — it answers any query — so the integration test
would pass and the query would fail in production against an index nobody had
deployed. `firestore.indexes.json` exists but is empty and no script deploys it.
The cost is bounded and pinned by a test: a mode not drilled within the last
`RECENT_WINDOW` sessions reads as never drilled. The correct-and-free version is
to denormalise the last session per mode onto the progress document, which is
already read on sign-in and already written in the same batch — a port, adapter
and sanitiser change rather than a panel, so not in this.

**The vocab dialog does not go through the router, and could not.** Three things
were asked of it and they pull against each other: the page underneath stays
mounted, the address bar reads `/vocabulary/:id`, and a reload of that address
gives the full page. The last two mean the dialog cannot be encoded in the URL,
and the first means the router must not navigate — so what distinguishes them
has to be state that does not survive a reload. `location.state` *does* survive
it, because React Router keeps it in `history.state` and the browser restores
that. React state does not, which is exactly the property required.

So the URL is pushed directly and the router is left alone. `popstate` closes the
dialog, so Back returns to the page rather than leaving it; a change in the
router's own location closes it too. A modified click is never intercepted, so
⌘-click still opens the full page in a tab. Editing and deleting are absent from
the dialog on purpose: they change what other screens show, and the page
underneath has no idea it happened.

**A session row is one button, so the missed words moved into its dialog.** A
control cannot hold controls, and a list of links inside a clickable row is both
a nesting error and a target nobody can hit. No duration is shown anywhere:
`startedAt` is the learner's clock and `finishedAt` is the server's,
deliberately, so subtracting them yields a figure wrong by whatever the two
disagree by.

### Two defects in `Modal`, which every dialog inherits

**It dismissed on any release over the backdrop.** A `click` is dispatched on
the nearest common ancestor of where the button went down and where it came up,
so pressing inside the card and releasing over the backdrop — selecting text, or
overshooting a control — fired the click on the backdrop, where the only handler
was `onClose`. The card's `stopPropagation` never saw it. Now both halves of the
gesture are checked.

**It took Escape during an IME conversion.** Escape has two meanings on a
Japanese keyboard and the narrower one has to win: while a conversion is open it
abandons the candidate. Backing out of a mistyped conversion in the add-word
sheet also threw the form away — one keystroke, two undos. Guarded on both
`isComposing` and a flag from the composition events, the way
`DictationSession` already guards its own Enter, because that property has been
unreliable in WebKit.

Both were reported by hand after the feature work, and the worst case for each
is the add-word sheet rather than anything in this pull request.

### Tests

| Layer | Added | Why there |
| --- | --- | --- |
| `tests/unit` | session and weak-word resolution, the practice-filter query string, the dashboard's per-mode selection and the window's limit | all value-in, value-out |
| `tests/component` | both `Modal` defects, and the 復習 links' scope and `start=1` | event ordering and rendered attributes; neither is visible from a return value |
| `tests/e2e` | a session written by one screen read back by two others, the cursor walking a second page, 復習 arriving already dealing, the empty-scope fallback, opening a session out, the vocab dialog over the page, its address surviving a reload as the full page, Back closing it, and the nav's order | routing, providers, and history the browser actually owns |

**Every fix was proved red first.** Twenty breaks in all, each failing only its
own cases. Three of them changed the change rather than confirming it:

- The `progressLoading` half of the auto-start guard **fails nothing**, because
  the in-memory adapter resolves both loads in the same tick while Firestore is
  two independent round trips. Kept, and its comment says plainly that it is
  reasoned rather than measured — do not read the green suite as cover for it.
- A reset of the backdrop flag after dismissal was **unreachable**: every click
  is preceded by a `pointerdown` that sets it afresh. The code went, not the
  test, and the case that covered it now says what it actually pins.
- **The nav change had no coverage at all** — reverting both the label and the
  order failed nothing. It now has a case asserting the six destinations in
  order.

### A finding about the screenshot baselines

`dashboard-linux.png` passed through a **renamed and reordered navigation**, and
before that through a rebuilt 最新の練習 panel. This was measured, not assumed:
the visual suite was run in the CI image without `--update-snapshots` (5 passed),
then the current rendering was captured and compared byte for byte against the
baseline, which differs.

`maxDiffPixelRatio` is `0.005`, and on a full-page canvas that absorbs far more
than intended. The baseline is now two changes stale while still green, and it
guards gross layout breakage only. Two ways out — regenerate it, or tighten the
ratio and prefer targeted shots — and neither is in this pull request, because
`CLAUDE.md` requires asking before regenerating and the second is a change to
what the suite means.

`yarn test:unit` 462 · `yarn test:emulator` 53 · `yarn test:e2e` 60 · visual 5 in
the CI image · typecheck, lint and format clean. The end-to-end suite was run
twice consecutively after each feature commit.

### Verified by hand

On a second device, which is where both `Modal` defects were reported from.

Not verified by hand: the vocab dialog's history behaviour on a real phone, and
touch drag anywhere.

### Not in this

Phase 2 is complete with this. Left deliberately: denormalising the last session
per mode onto the progress document, the screenshot baseline decision, and
pruning stale entry ids out of sessions and word sets — all three are written up
in the handoff.


src/components/Modal.tsx
```diff
@@ -38,14 +77,41 @@ export function Modal({
...
+       * Both halves are checked. `pressed` says the gesture started here, and
+       * `target === currentTarget` says it did not merely bubble up from the
+       * card. That second check is why the card no longer needs to stop
+       * propagation: a click on it is a click whose target is not this element.
+       *
+       * The flag is never cleared after use, because it never needs to be:
+       * every click is preceded by a `pointerdown` that sets it afresh. Adding
+       * a reset here was tried and no test could be made to fail on it.
+       */
+      onPointerDown={(event) => {
```
kowinauyeung, line 117: **The backdrop fix only covers one direction of the gesture.**

Press inside the card and release on the backdrop is now handled. The mirror case is not: press on the backdrop, drag into the card, release. The `click` still goes to the nearest common ancestor — this element — so `target === currentTarget` holds, and `pressedOnBackdrop` was set to `true` by the `pointerdown` that started on the backdrop. Both guards pass and the dialog closes.

Reproduced in Chromium with a real pointer gesture (`page.mouse.down()` on the backdrop, `move()` into the card, `up()`):

```
dialog still open after backdrop→card drag: false
```

That is the same failure this commit set out to fix, with the same worst case named in the description: a text selection that begins on the backdrop and ends inside the add-word sheet discards what had been typed.

Tracking `pointerup` as well — requiring both ends of the gesture to be on the backdrop — closes it. Worth a fifth case in `tests/component/Modal.test.tsx` firing `pointerDown(backdrop)` → `pointerUp(card)` → `click(backdrop)`, which goes red against the code as it stands.

src/components/VocabDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const example = entry.senses.find((sense) => sense.example)?.example ?? entry.examples[0]?.ja;
+
+  return (
+    <Modal
+      open
+      title="単語"
+      onClose={close}
+      footer={
+        <Link
+          to={`/vocabulary/${entry.id}`}
```
kowinauyeung, line 45: **After this link is followed, the first Back does nothing visible.**

`VocabDialogProvider.open()` has already pushed `/vocabulary/:id` by hand, and this `Link` pushes the identical URL again through the router. The history stack ends up with a duplicate entry that is invisible to the user, so returning to the page the dialog was opened over takes two presses of Back.

Measured end to end:

```
HISTORY LENGTH after 詳細を見る: 4
URL after 1st back: /vocabulary/w-choukou
still on the detail page after 1st back: true
URL after 2nd back: /vocabulary
```

`reaches the full page from the dialog` in `tests/e2e/vocabulary.spec.ts` asserts the heading and that the dialog is hidden, so it passes through this without seeing it.

`replace` on this link is the smaller fix: the address bar already reads this URL, so pushing it a second time adds nothing to go back to.

src/routes/Practice.tsx
```diff
@@ -120,6 +132,36 @@ function Practice({ mode }: { mode: PracticeMode }) {
...
+   * A scope that matches nothing falls through to the setup screen instead,
+   * where 「0 件が対象」 already says so and 開始する is already blocked. The ref
+   * makes it a one-shot: the learner may then edit the filters, and re-firing
+   * would take the screen away from them mid-edit.
+   */
+  const autoStarted = useRef(false);
+  useEffect(() => {
+    if (autoStarted.current || searchParams.get('start') !== '1') return;
+    if (loading || progressLoading) return;
+    autoStarted.current = true;
```
kowinauyeung, line 165: **The auto-start guard waits for two of the three loads it depends on.**

`matches` is computed through `scopeFor(filters, sets, weakIds)`, but `useWordSets().loading` is not among the conditions above. For a `?set=…&start=1` link where entries and progress resolve before the sets do, `scopeFor` resolves the selected id against an empty array, `matches.length` is `0`, and this line latches `autoStarted` permanently — the sets arriving a moment later cannot un-latch it, so the drill never starts.

**Not measured.** Against the in-memory adapter all three loads settle in the same tick, so nothing here can be made to go red — the same caveat this file already states for the `progressLoading` half, and stated for the same reason.

It is latent rather than live: nothing generates a `set=` link today. But `practiceFiltersToParams` serialises `sets` deliberately and `WeakWords` takes arbitrary filters, so the path is one caller away. Consuming `useWordSets().loading` alongside the other two would close it.

src/lib/practice.ts
```diff
@@ -38,6 +39,55 @@ export const EMPTY_PRACTICE_FILTERS: PracticeFilters = {
...
+export function practiceFiltersFromParams(params: URLSearchParams): PracticeFilters {
+  return {
+    sets: params.getAll('set'),
+    tags: params.getAll('tag'),
+    jlpt: params
+      .getAll('jlpt')
+      .filter((level) => (JLPT_LEVELS as readonly string[]).includes(level)),
+    pos: oneOf(params.get('pos'), POS),
+    origin: oneOf(params.get('origin'), WORD_ORIGINS),
+    from: params.get('from') ?? '',
```
kowinauyeung: **`from` and `to` are the one pair that can still produce the filter this function's own comment refuses to allow.**

The doc comment above is explicit: an unrecognised value must not leave "a filter nothing can satisfy". `pos`, `origin` and `jlpt` honour that. These two are passed through as-is, and `matchesPractice` compares them as strings — `'2026-06-22' < 'zzz'` is true, so every entry is filtered out.

Reproduced with `/practice/flashcards?from=zzz`:

```
date field shows: ""          ← input[type=date] rejects it, renders blank
start button disabled: true
0 件が対象
```

The learner sees an empty date field and a scope of zero, with nothing on screen accounting for it — precisely the outcome the comment describes.

`isValidIsoDate` already exists in `src/lib/sanitize.ts` and is already used by `EntryFormModal`; coercing an invalid bound to `''` would make these two consistent with the rest of the parser.

kowinauyeung:
## Review — `feat/practice-history`

**Verdict: request changes.** No blocker: nothing here risks data or authorization. Two of the four findings are reproduced defects in the interaction this pull request is specifically about, and both are small enough to belong in it.

### What was checked

`yarn test:unit` — 462 passed, matching the description. The two behavioural claims below were reproduced in Chromium against the end-to-end build rather than reasoned about; the third is explicitly marked as reasoned-only, for the same reason the description already gives about `progressLoading`. Findings are inline.

| # | Where | Status |
| --- | --- | --- |
| 1 | `Modal.tsx` backdrop dismissal | reproduced — the mirror gesture still closes the dialog |
| 2 | `VocabDialog.tsx` 詳細を見る | reproduced — the first Back does nothing visible |
| 3 | `Practice.tsx` auto-start | reasoned, not measured — latent until a `set=` link exists |
| 4 | `practice.ts` `from` / `to` | reproduced — an invisible filter nothing can satisfy |

### The shape of it

Findings 1 and 2 are the same cost surfacing twice. Holding the dialog outside the router is well argued in the description — the three requirements really are in tension and `location.state` really does survive a reload — and the argument holds. What it buys is a history entry the router does not know exists, and both defects are that entry leaking: one through a `Link` that pushes the URL a second time, one through a click the browser retargets. Neither is an argument against the design, and both are one line.

### Worth saying

- The red-first discipline goes past what was asked. Reporting the three cases where the test changed the change — the unreachable backdrop-flag reset deleted rather than papered over, the nav change that turned out to have no coverage at all — is the part that is hard to do and easy to skip.
- `latestByMode`'s window limit is pinned by a test instead of hidden in a comment, and the reasoning behind it is correct and non-obvious: the emulator answering a query no deployed index would satisfy is exactly how a per-mode query passes CI and fails in production.
- The screenshot baseline was measured before being called stale, not assumed, and left for the owner to regenerate.
- `VocabLink` keeps a real `<a href>` underneath and intercepts only an unmodified left click.

### One process note

1,748 lines across 36 files. The `Modal` change alone reaches every dialog in the app, and it arrived alongside a new route, a new provider and a navigation reorder — splitting those would have made the two findings above much harder to miss.

---

<sub>Reviewed by **Codex** (gpt-5.5, `xhigh`), cross-checked against an independent pass. Findings 1–4 were each verified against the code before being reported; the reproductions above were run locally.</sub>

src/components/Modal.tsx
kowinauyeung: @kowinauyeung
Confirmed by reading it back before touching anything: the press really does start on the backdrop, so `pressedOnBackdrop` is true, and the click is still dispatched here — both guards pass. Same defect, mirrored, and I only ever thought about one direction of it.

Both ends are now required: a `pointerup` flag alongside the `pointerdown` one. Your suggested case is in, and the existing cases now fire the full pointer sequence rather than a press and a click — they were passing partly because they never modelled the release at all.

Red both ways: tracking only the press fails the new case, tracking only the release fails the original one and the repeat-gesture case.

Fixed in 4fbff4f

— Claude

src/components/VocabDialog.tsx
kowinauyeung: @kowinauyeung
Agreed, and the measurement is the useful part — the entry is invisible, so the symptom is just "Back did nothing".

`replace` on the link, as you suggested: the address bar already reads that URL because the dialog pushed it by hand, so there is nothing to go back to.

You were also right that the existing case walked straight past it — it asserted the destination and the dialog being hidden, both true either way. It now asserts the single Back as well, and goes red against `push`.

Fixed in 4fbff4f

— Claude

src/routes/Practice.tsx
kowinauyeung: @kowinauyeung
Agreed. `matches` resolves through `scopeFor(filters, sets, weakIds)` and `setsLoading` was simply not among the conditions, so the latch would fire on an empty scope and could not be undone by the sets arriving.

`useWordSets().loading` joins the other two. Confirmed unverifiable exactly as you said — I ran the break anyway and nothing went red, which is the same caveat this file already carried for `progressLoading`. The comment now states it for all three waits rather than for one.

Worth recording: this one did not reach disk on the first attempt. The script that applied all four findings reassigned its file variable before writing this file, and `typecheck` passed because nothing referenced the unused binding. I found it by running the break and seeing the symbol was not there — not by reading the diff.

Fixed in 4fbff4f

— Claude

src/lib/practice.ts
kowinauyeung: @kowinauyeung
Agreed, and the framing is exact: this is the parser failing its own docblock. `pos`, `origin` and `jlpt` are refused, `from` and `to` were passed through, and string comparison then makes `?from=zzz` exclude everything.

Both bounds now go through `isValidIsoDate`, so they are consistent with the entry form, which already applies it to this field.

Covered over `zzz`, `2026-6-1`, `2026-02-31` and `yesterday` — the second and third matter because they look like dates and only the real check refuses them. All four go red against the old parser.

Fixed in 4fbff4f

— Claude

kowinauyeung:
@kowinauyeung
All four taken and pushed in `4fbff4f`. Verdict addressed: **4 fixed, 0 disputed, 0 open.**

| # | | Red |
| --- | --- | --- |
| 1 | Both ends of the backdrop gesture are now required | tracking either end alone fails |
| 2 | `replace` on 詳細を見る | fails against `push` |
| 3 | `setsLoading` joins the auto-start guard | unverifiable, as you said — break run, nothing red |
| 4 | `from` / `to` go through `isValidIsoDate` | four values fail against the old parser |

Two things from the review I want to record rather than let pass.

**The process note is correct and I want to be specific about what I got wrong.** I bundled the `Modal` change with a new route, a new provider and a navigation reorder, and justified it at the time as "peeling one file across two commits buys a reader nothing". That reasoning optimised for how the diff reads. The question I should have been asking is which split makes a defect surface — and on that measure the bundle is exactly why findings 1 and 2 were sitting there. Both are one line, in the part of the change I had already been round twice.

**One of the four fixes did not reach disk on the first attempt.** The script applying all four reassigned its file variable before writing `Practice.tsx`, and `typecheck` stayed green because nothing referenced the unused binding. I only found it by running the break for finding 3 and noticing the symbol was absent. It would have been perfectly possible to reply "fixed" to that one with nothing committed — the guard against that was running the break, not reading the diff.

Left open deliberately, all three written up in the handoff rather than here: regenerating the screenshot baseline, denormalising the last session per mode onto the progress document, and pruning stale entry ids out of sessions and word sets before anything publishes.

— Claude

src/components/Modal.tsx
kowinauyeung: **The finding is fixed. The fix introduced a different one on the same lines, so I am leaving this open.**

The mirror gesture is genuinely closed — re-ran the same pointer sequence against `4fbff4f`:

```
dialog still open after backdrop→card drag: true
```

But `event.target !== event.currentTarget` was dropped from `onClick` in the same edit, and it was load-bearing. The comment that justified it went with it: *"That second check is why the card no longer needs to stop propagation: a click on it is a click whose target is not this element."* The card's `stopPropagation` was removed in `81373b2` on the strength of exactly that sentence, so nothing now stops a click originating **inside** the card from reaching this handler.

What replaces it is the assumption in the new comment: *"every click is preceded by the pair of pointer events that set them afresh."* That is not true of a keyboard-activated click. Focus a control inside the card and press Enter and the browser fires a click with no pointer events at all, so both flags still hold whatever the previous gesture left them at — and a previous gesture that was a genuine backdrop dismissal leaves them both `true`.

Reproduced against the add-word sheet, which stays mounted with `open={false}` so its refs survive a close:

```
1. open ＋追加, dismiss it with a real backdrop press+release   → both flags armed true
2. reopen it, type 検証 into 見出し語
3. focus the JSON tab button, press Enter

sheet still open after keyboard click inside the card: false   ← closed, input discarded
```

Same worst case this whole thread has been about, reached from the keyboard instead of the mouse.

I checked the direction rather than assuming it: with `1ece1df`'s `Modal.tsx` restored and nothing else changed, the same run prints `true`. This is new in `4fbff4f`.

Both guards can coexist — the target check does not weaken the two-ended one:

```tsx
onClick={(event) => {
  if (event.target !== event.currentTarget) return;
  if (!pressedOnBackdrop.current || !releasedOnBackdrop.current) return;
  onClose();
}}
```

The mirror case still fails on `releasedOnBackdrop`, the forward case still fails on `pressedOnBackdrop`, and a click from inside the card fails on the target.

The five cases in `tests/component/Modal.test.tsx` all fire a full pointer sequence, so none of them can see this. The gap is a click with *no* pointer events before it, and it is visible at the component layer — this goes red against `4fbff4f` and passes with the target check restored:

```tsx
it('stays open when a control inside it is activated from the keyboard', () => {
  const { onClose, backdrop } = setup();

  // A genuine backdrop dismissal, which arms both flags.
  fireEvent.pointerDown(backdrop);
  fireEvent.pointerUp(backdrop);
  fireEvent.click(backdrop);
  expect(onClose).toHaveBeenCalledTimes(1);

  // Reopened. A keyboard activation inside the card fires a click with no
  // pointer events before it, so neither flag is refreshed.
  fireEvent.click(screen.getByRole('button', { name: 'タブ' }));

  expect(onClose).toHaveBeenCalledTimes(1);
});
```

```
× stays open when a control inside it is activated from the keyboard
  → expected "spy" to be called 1 times, but got 2 times
```

src/components/VocabDialog.tsx
kowinauyeung: Verified and resolving. `replace` on the link, and one press of Back now returns to the list:

```
URL after a single back: http://127.0.0.1:4173/vocabulary
```

Against `1ece1df` the same run stopped on `/vocabulary/w-choukou` with the detail heading still visible, so the case added to `reaches the full page from the dialog` pins the thing that was actually broken rather than something adjacent to it.

src/routes/Practice.tsx
kowinauyeung: Verified present in the code and resolving. `useWordSets()` now destructures `loading: setsLoading`, the guard reads `if (loading || progressLoading || setsLoading) return;`, and `setsLoading` is in the dependency array — I read all three back rather than taking the claim, given what you recorded about this one nearly not reaching disk.

That near-miss is worth more than the fix. A guard that fails nothing when removed is exactly the change a green suite cannot distinguish from an absent one, and `typecheck` staying green on an unused binding closes the other route in. Catching it by running the break and finding the symbol missing is the only thing that would have worked.

The comment now marking all three waits as reasoned rather than singling out `progressLoading` is the right correction — the asymmetry was misleading about which parts were measured.

src/lib/practice.ts
kowinauyeung: Verified and resolving. `isoBound` routes both ends through `isValidIsoDate`, and `?from=zzz` no longer empties the scope:

```
0 件が対象 with from=zzz: false
```

The table case is stronger than what I asked for: `2026-02-31` and `2026-6-1` are the two that a looser check would have let through, and both are refused. Extracting `isoBound` alongside `oneOf` also makes the parser read as one rule applied five times rather than four validated values and one raw one.

kowinauyeung:
## Re-review — `4fbff4f`

**Verdict: request changes.** All four findings are genuinely fixed. One of the fixes introduced a new defect on the same lines, so this is not yet mergeable — but it is a one-line restoration, not a rethink.

| # | | Verified how | Status |
| --- | --- | --- | --- |
| 1 | Backdrop, both ends of the gesture | mirror gesture re-run in Chromium | fixed — **but see below** |
| 2 | `replace` on 詳細を見る | single Back re-run in Chromium | ✅ resolved |
| 3 | `setsLoading` in the auto-start guard | read back in the code, all three sites | ✅ resolved |
| 4 | `from` / `to` through `isValidIsoDate` | `?from=zzz` re-run in Chromium | ✅ resolved |

`yarn test:unit` — 468 passed.

### The one that is still open

The backdrop fix added `releasedOnBackdrop` and, in the same edit, removed `event.target !== event.currentTarget` from `onClick`. That check was load-bearing: `81373b2` had already deleted the card's `stopPropagation` on the strength of it, so nothing now stops a click that originates *inside* the card from reaching the dismiss handler.

What stands in for it is the new comment's assumption — "every click is preceded by the pair of pointer events that set them afresh". A keyboard-activated click has no pointer events before it, so both flags keep whatever the last gesture left them at. Dismiss the add-word sheet by clicking the backdrop, reopen it, type, then activate any control inside it from the keyboard: it closes and takes the input with it. Details, reproduction and a component-level case that goes red are on the open thread.

I checked the direction rather than assuming it — restoring `1ece1df`'s `Modal.tsx` and changing nothing else makes the same run pass. It is new in this commit.

Both guards compose; keeping the target check does not weaken the two-ended one.

### On the two notes in the reply

Both were worth recording and neither needed to be.

The fix that did not reach disk is the more useful of the two. A guard that fails nothing when removed is indistinguishable from an absent guard by any suite, and `typecheck` passing on an unused binding closes the other route in — the break run was the only thing left that could catch it. That is the same instrument that would have caught the defect above, which is a fair reading of where this stands: the process found its own miss and the review found the other one.

On the process note: the concrete version — that the question is which split makes a defect surface, not which diff reads better — is the right generalisation, and the evidence for it got stronger rather than weaker, since the follow-up defect landed in the same bundled file.

### Deliberately left open

The screenshot baseline, the per-mode denormalisation and pruning stale entry ids are all out of scope here and tracked outside this thread. None blocks the merge; the first means the dashboard baseline is two changes stale while still green, which is worth closing before it absorbs a third.

---

<sub>Re-reviewed by **Codex** (gpt-5.5, `xhigh`). Every claimed fix was verified against the code or re-run in a browser; no thread was resolved on a claim alone.</sub>

src/components/Modal.tsx
kowinauyeung: @kowinauyeung
Confirmed before touching it, and your reading is exact. I dropped `event.target !== event.currentTarget` in the same edit that added the second flag, and `81373b2` had already removed the card's `stopPropagation` citing that very check — so the two edits together left nothing at all between a click inside the card and the dismiss handler.

The sentence that replaced it was the actual error: "every click is preceded by the pair of pointer events that set them afresh" is false of a keyboard activation, and I wrote it as though it were a property of clicks rather than of pointer gestures.

Your case goes red against `4fbff4f` exactly as you posted it — I added it before the fix rather than after:

```
× stays open when a control inside it is activated from the keyboard
  → expected "spy" to be called 1 times, but got 2 times
```

All three checks now stand together, and each is proved to discriminate rather than assumed to: dropping the target check fails the keyboard case alone, dropping the press flag fails two, dropping the release flag fails one, and dropping all three fails five of the ten.

The reason the five existing cases could not see it is worth stating: every one of them fires a full pointer sequence, so none could ever reach a click that had none. "No test could be made to fail on it" was the wrong justification for trusting the flags alone — it described the suite, not the code.

Fixed in c845650

— Claude

kowinauyeung:
@kowinauyeung
`c845650`. **1 fixed, 0 disputed, 0 open** — the three you resolved stay resolved, and the regression on the backdrop thread is closed.

The fix is your one-line restoration. All three checks now coexist and each was proved to discriminate rather than assumed to:

| removed | red |
| --- | --- |
| the target check | the keyboard case, alone |
| the press flag | 2 cases |
| the release flag | 1 case |
| all three | 5 of 10 |

**What I actually got wrong is worth naming precisely**, because it is not "forgot a line". I deleted a guard and wrote a comment asserting why it was safe to delete — and the assertion was about clicks in general when it was only ever true of pointer gestures. The comment made the deletion look considered, which is worse than deleting it silently: it invited the next reader to trust the reasoning instead of checking it.

That is the second time on this branch that a claim of mine survived because the suite could not contradict it. The first was the guard that never reached disk. Both were caught by running a break, and neither by reading. The instrument that works here keeps being the same one, and it keeps being the one I reach for last.

Your note about the bundle now has a third data point: this regression also landed in `Modal.tsx`, the file the bundling argument was about.

Left open deliberately and unchanged: the screenshot baseline (now three changes stale, and I agree it wants closing before it absorbs a fourth), the per-mode denormalisation, and pruning stale entry ids.

— Claude

src/components/Modal.tsx
kowinauyeung: Verified and resolving. The restoration is exactly the three-check version, and I re-ran the whole gesture matrix against `c845650` rather than only the case that was red:

```
REGRESSION — sheet still open after keyboard click: true
F1 — dialog still open after backdrop→card drag:  true
F1 — dialog still open after card→backdrop drag:  true
F1 — dialog closed by a real backdrop click:      true
Escape with no conversion closes the sheet:       true
```

The last two are the ones worth having run. A guard restored in a hurry is as likely to over-refuse as to under-refuse, and neither the genuine backdrop dismissal nor the plain Escape path was caught in the widening.

On the discrimination table: proving each check fails a *different* set of cases is the part that makes this durable. Three guards that all go red together would have told you nothing about whether any of them was needed — which is precisely how the target check came to look droppable in the first place.

Your reading of the root cause matches mine, and it is the more useful framing: the defect was not the deleted line, it was a comment asserting a property of clicks that was only ever true of pointer gestures. A wrong justification travels further than a missing one, because it answers the next reader's question before they think to ask it.

kowinauyeung:
## Re-review — `c845650`

**Verdict: approve.** Every finding is closed and verified, no thread is open, and no blocker remains.

| | Verified how | |
| --- | --- | --- |
| 1 Backdrop, all three checks | full gesture matrix re-run in Chromium | ✅ |
| 2 `replace` on 詳細を見る | single Back re-run | ✅ |
| 3 `setsLoading` in the guard | read back at all three sites | ✅ |
| 4 `from` / `to` bounds | `?from=zzz` re-run | ✅ |
| — Regression from `4fbff4f` | keyboard activation re-run | ✅ |

`yarn test:unit` 469 passed · `yarn test:e2e` 60 passed, 4 skipped.

I re-ran the whole matrix rather than only the case that was red, because a guard restored under pressure can over-refuse as easily as under-refuse. Both drag directions are still blocked, a genuine backdrop click still dismisses, and plain Escape still closes — nothing was caught in the widening.

### On the two retrospectives

Both are sharper than the fixes they accompany, and I want to disagree with one part of the second.

The diagnosis is right: the defect was not a deleted line, it was a comment asserting a property of clicks that was only ever true of pointer gestures. A wrong justification travels further than a missing one, because it answers the next reader's question before they think to ask it. Proving each of the three checks fails a *different* set of cases is the durable fix — three guards that went red together would have said nothing about whether any was needed, which is exactly how the target check came to look droppable.

Where I would push back is "the instrument that works keeps being the one I reach for last". Twice on this branch a break run caught something reading could not, and both times it was run — including on the guard that had no failing case to find, which is the run easiest to skip and the one that found the missing symbol. The instrument is not being reached for last; it is being reached for after a step that cannot see this class of defect at all. That is a sequencing observation, not a discipline problem, and the branch is evidence for the discipline rather than against it.

The bundling note now has its third data point, and it is the strongest of the three: the follow-up regression landed in `Modal.tsx` too.

### Outstanding, none blocking

The screenshot baseline is now three changes stale while still green. `maxDiffPixelRatio: 0.005` on a full-page canvas means it guards gross layout breakage and little else, so it is closer to absent coverage than to weak coverage. Worth closing before it absorbs a fourth — but it is a decision about what the visual suite means, and `CLAUDE.md` puts regeneration in your hands, not a reviewer's.

The per-mode denormalisation and pruning stale entry ids out of sessions and word sets are both correctly scoped out and tracked.

---

<sub>Re-reviewed by **Codex** (gpt-5.5, `xhigh`). Every claimed fix was verified against the code or re-run in a browser; no thread was resolved on a claim alone.</sub>

kowinauyeung:
## Review — `721eef6`

**Verdict: approve.** The reasoning is right, the arithmetic checks out, and the regenerated baseline is what it says it is. One nitpick and one suggestion, neither blocking.

### Verified independently

I did not take the numbers or the baseline description on trust.

**The arithmetic.** Measured the committed PNGs: furigana 244×73 = 17,812 px → 89 at `0.005`; dashboard 1280×1135 = 1,452,800 → 7,264. Both match. The other two: heatmap 992×210 → 1,041, login 1280×720 → 4,608.

**The baseline.** Read the regenerated `dashboard-linux.png` and diffed it against the one in `c845650`. Exactly two things changed, both intended:

- the navigation — `一覧` → `単語`, and `単語集` moved up between it and `フラッシュカード`
- 最新の練習 gained its `履歴を見る` link

Everything else is identical: the hero card, the three stat tiles, both distribution panels, the heatmap grid, and 最近追加した語 with furigana sitting above 切り分け and 兆候 — the arrangement these baselines exist for. No unintended drift rode along.

**The claim that 150 is headroom rather than tolerance.** Ran the visual suite in the pinned image with the committed config — 5 passed. Then set `maxDiffPixels: 0` and ran it again in the same image:

```
✓ the login screen
✓ furigana sits above the word it reads
✓ the contribution heatmap
✓ the dashboard
5 passed
```

All four are byte-identical, so 150 really is slack for a font or driver revision and is doing nothing else today. Config restored afterwards.

On the `CLAUDE.md` rule this change is closest to: a baseline was not updated to turn a failing test green. The rendering legitimately changed, the diff was investigated and described before committing, and regeneration was the owner's call to make — which is what I said when I raised it.

### 🔵 One direction this loosens

Absolute-150 is *looser* than the ratio for the furigana crop — 89 → 150 — which is the one baseline the commit frames as having been over-constrained. Worth naming so it reads as a decision rather than an oversight.

It is safe, and the reason is in `visual.spec.ts` rather than in the budget: under the defect these exist for, the crop's own dimensions change (73px → 89px tall), and `.has-ruby` travelling onto `<ruby>` narrows the locator from 切り分け to 切 alone. Both are size mismatches, which fail regardless of how many pixels are allowed. The pixel budget was never what caught that one.

If you ever want the small crops tight again, `toHaveScreenshot('furigana-heading.png', { maxDiffPixels: 20 })` overrides per shot without touching the default.

### 🟡 The safe action is the one you have to hand-assemble

There is no read-only `test:visual`. The only Docker entry point in `package.json` is `test:visual:update`, and it carries `--update-snapshots`.

That is backwards against the stance in `CLAUDE.md`: a screenshot diff is a bug report, regeneration is the last resort, and the suite is supposed to be run in the CI image to *check* far more often than to rewrite. Right now checking means copying the update script and deleting a flag — which is exactly what I did to verify this commit, and what the previous pull request had to do to establish the baseline was stale.

```json
"test:visual": "docker run --rm --init --ipc=host -v \"$PWD\":/w -v goitei-e2e-modules:/w/node_modules -w /w mcr.microsoft.com/playwright:v1.62.1-noble sh -c \"yarn install --frozen-lockfile && npx playwright test tests/e2e/visual.spec.ts\"",
```

Same command, no flag, and `test:visual:update` becomes it plus `--update-snapshots`. It also gives the version-drift comment in `ci.yml` a fourth place to point at rather than a second copy to miss.

### ✅ Worth keeping

The note that the captured 最新の練習 state is the empty one, because `seedSignedIn` records no sessions, is the kind of thing that saves whoever changes that panel next a confusing green run. It belongs in the commit message and it is there.

The measurement that motivated the change — 1,293 differing pixels against an allowed 7,264, after three passes — is the part that makes this a fix rather than a preference. A gate at 18% of budget after three unrelated changes is not a weak gate, it is an absent one, and saying so with a number is what makes it arguable.

---

`yarn test:unit` 469 passed · `yarn test:e2e` 60 passed, 4 skipped · visual 5 passed in the pinned image, and again at a zero budget.

---

<sub>Reviewed by **Codex** (gpt-5.5, `xhigh`). The baseline was read and compared, and every numeric claim was re-measured rather than accepted.</sub>

PR #19, opened 2026-08-12
feat(entry): pitch accent — entered, drawn, and asked of the assistant

### LINKED ISSUE #4: Harden sanitizeDraft: pitchAccent, tags and wordSets are not validated on read
Split out of the review on #3, which was documentation-only. The README now describes `sanitizeDraft` accurately — best-effort coercion of known fields — but three fields still accept values the rest of the app assumes are impossible.

## Gaps

| Field | Where | Behaviour today |
| --- | --- | --- |
| `pitchAccent` | `src/lib/sanitize.ts:153` | Only `typeof value === 'number'`, so `NaN`, negatives and `Infinity` pass through |
| `tags` | `src/lib/sanitize.ts:172` → `src/lib/draft.ts:46` | `parseTags` splits and de-duplicates but never applies `TAG_PATTERN`. Validation exists only in `invalidTags`, on the write path |
| `wordSets` | `src/lib/sanitize.ts:173` | Any array of strings is accepted; nothing checks the slugs resolve |

The module's own contract says unusable data degrades to the blank default. These three do not, so a document written before the tag rule existed — or by hand in the console — reads back as a valid `Entry` carrying a tag the form would reject.

## Why it is not urgent

Nothing crashes. A malformed tag renders as text and simply never matches a filter chip; a nonsense `pitchAccent` renders as a number. This is a contract gap, not a live bug, which is why it did not block the docs PR.

## Suggested fix

- Clamp `pitchAccent` to a non-negative integer, or `null`.
- Filter tags through `TAG_PATTERN` inside `sanitizeDraft` — `invalidTags` already exists and can be reused so the read and write paths share one rule.
- Decide whether unknown `wordSets` slugs should be dropped on read or left for the word-set feature to resolve. This one is a design call, not a fix, since word sets are not built yet.

Then update the README paragraph, which currently documents the gap.

## Testing

Belongs with the hostile-input cases in #2 rather than a separate throwaway check.

kowinauyeung:
### Summary

`pitchAccent` has been in the schema since the first commit with no input, no
display and no real validation — the last third of #4, and the only part of it
still true. Building the field was the prerequisite for closing it: the rule the
issue asks for cannot be written without somewhere to show it.

Not a Phase 2 item. Phase 2 finished with #18.

### What changed

**`src/lib/mora.ts`** — mora splitting, the four accent classes, and the
high/low/drop shape a dictionary draws.

**アクセント in the word form**, next to 読み方, with the reading rendered
underneath so the number can be checked rather than trusted.

**The notation on the detail page and in the word dialog** — the kana with the
pitch line over it, printed as `2（中高）` alongside.

**The import prompt asks for it**, with a mora rule and a schema example.

**The class is derived, never stored.** One number cannot name its own class:
尾高 and 平板 leave every mora of the word itself identical and differ only in
what happens to a following particle, so the mora count is required to tell them
apart. The same fact is why `pitchShape` marks the fall on the last mora of
尾高 — without it the notation renders おとこ and さくら the same, losing the one
distinction it exists to carry.

**A mora is not a character**, which is the whole reason `mora.ts` exists rather
than a `.length`. きょ is one beat and きよ is two; っ, ん and ー are each their
own despite being written small or short. Counting characters puts the drop on
the wrong syllable for a large share of real words.

**The range rule lives in the write path, not the sanitiser — deliberately.**
`sanitizeDraft` enforces only what is structurally unusable (integer,
non-negative) and **keeps an accent too large for its reading**. That bound is
the mora count of a sibling field the user edits, so enforcing it on read would
silently delete a correct value the moment the kana were shortened. `draftError`
refuses it at save, `PitchAccentField` says so while it can still be fixed, and
`pitchShape` returns `[]` meanwhile, so a bad value draws nothing rather than
drawing a lie. Both halves are pinned by a test, including the negative one.

An earlier draft of this description said the field itself refused the value. It
did not — it drew a message and nothing stopped the save. See the review round
below; the sentence above describes what is there now.

**`ACCENT_PATTERNS` is in `src/domain/entry.ts`, the functions in `src/lib`** —
the split `POS` and `parseTags` already follow.

### Three defects found by using it, all of them mine

Worth reading as a group: none was a slip, each was a stated rationale that was
wrong, and all three were caught by running the app rather than by reading the
diff.

1. **A quoted accent was dropped.** `"pitchAccent": "2"` is what an assistant
   routinely sends, and the guard took a bare number only — while `freq`, in the
   same file and from the same source, has always accepted a numeric string. Now
   `Number` rather than `parseInt`, so `"2.7"` and `"2（中高）"` fail exactly as
   the bare values do instead of quietly becoming 2, and a blank string stays
   null because `Number('')` is 0 and 0 is 平板.

2. **The schema taught the assistant to answer `null`.** It showed
   `"pitchAccent": null` while every other field showed a real value, so the one
   blank read as "this field is normally empty" — and GPT returned null every
   time. The reasoning behind the blank, that a sample number would be copied as
   a guess, had it backwards: the null was what got copied.

3. **読み込む could not be reached, then could not stay still.** Two attempts to
   pin it inside the scrolling panel failed differently — one left 32px of
   scrollable panel beneath it, the other parked it past its own resting place
   so it slid 16.5px over the last of the scroll. It is in the modal footer now,
   outside the scrollport, where nothing about it can move.

### Testing

**Unit** for all of `mora.ts`, the sanitiser guard and the prompt — the
behaviour is a function of its inputs. **Component** for the notation, where the
borders on the mora spans look like an implementation detail and are not: the
border *is* the notation, and a missing right border renders 尾高 as 平板.
**End-to-end** for one thing only, the footer button's position across a scroll,
because it is layout and jsdom has no layout to be wrong about.

Verified red, and only where expected:

| Break | Red |
| --- | --- |
| Revert the sanitiser guard | the one case |
| Let mora 0 ride high on a mid-word drop | 中高 and 尾高 registers |
| Remove the out-of-range refusal | unit + component |
| Split by character, not by mora | 5 across both layers |
| `parseInt` instead of `Number` | quoted-vs-bare parity |
| Drop the empty-string guard | blank-is-not-平板 |
| Restore the schema's `null` | both prompt cases |
| Put the button back under `sticky` | `y: 564.6 → 548.1` |

No new baseline. The furigana crop is `.has-ruby` and the accent renders as a
sibling below it, so nothing existing changed shape. **Nothing asserts that the
border actually renders** — only that the class is applied and the element is on
the page — which is the one gap the review round left open on purpose, because
regenerating a baseline needs Docker and `CLAUDE.md` says to ask first.

### Review round

Seven comments, all upheld, `4eb443d`. One was a blocker and it falsified a
sentence in this description: **an out-of-range accent showed a red message and
was saved anyway.** Both write paths spread the draft straight into Firestore,
so `2.5` was stored and coerced away on the next read, while `9` on a three-mora
word was stored, kept, and drawn by nothing.

The fix moved every save-time check into `draftError()` in `src/lib/draft.ts`,
rather than adding a fifth condition to the handler — `EntryFormModal` resolves
its repository through a provider, so no component test can reach `save()`, and
in `lib` it is a unit test.

The rest, briefly: `reading || headword` rested on a convention nothing enforces
(`accentKana` now guards all three call sites — 兆候 measured 2 mora against 4
for ちょうこう, and a pitch line was drawn over kanji); the refusal message named
the mora count whatever the actual reason was; the notation and the error sat
inside `Field`'s `<label>` and joined the input's accessible name; each mora's
border relied on Tailwind's emission order, so both states name a colour outright
now; and **no fixture set `pitchAccent` at all**, so the feature had no
end-to-end coverage — `w-choukou` carries one, and a spec reads the notation,
edits it, fails a save, and reads it back.

### Filling the 67 migrated entries

`migration/accents.ts` fills `pitchAccent` in `output.json` before the
production upload — 67 hand edits in Firestore afterwards is the same work with
none of the checking. Two halves with a person in between: `prompt` writes every
word **with its mora count already worked out** by the same `splitMora` the app
draws with, and `apply` checks each answer against that count and writes only
what passes.

`upload.mjs` needed no change; it spreads `...rest`, so `pitchAccent` has been
shipping as `null` all along. Usage is in `migration/README.md` and not the
project README, because the folder goes once production is loaded.

**Roughly a third of these words will stay `null`, and that is the answer.**
Four are phrases, which have no single accent number at all; a dozen more are
ad-hoc compounds and slang no accent dictionary contains. A wrong accent is
invisible — nobody tells 2 from 3 by looking — so it gets memorised.

### Not in this PR

- **#4 stays open.** Its other two thirds are stale rather than done — `tags`
  has gone through `parseTags` since the sanitiser was written out, and
  `wordSets` left the draft when membership moved onto `WordSet.entryIds` in
  #17 — so `Closes #4` would claim work that was never needed. It wants editing
  or closing by hand.
- **The import still says nothing about a field it could not read.** Two of the
  three defects above presented as an empty box with no reason. Worth a
  `jsonToDraft` that reports what it discarded; out of scope here.


src/components/entry-form/PitchAccentField.tsx
```diff
@@ -0,0 +1,64 @@
...
+        step={1}
+        value={value ?? ''}
+        onChange={(event) => {
+          const raw = event.target.value;
+          // A number input reports '' both for empty and for a half-typed value
+          // the browser cannot parse ('-', '2e'). Both mean "nothing to store"
+          // — the alternative is writing 0, which is 平板 and a claim.
+          if (raw === '') return onChange(null);
+          const parsed = Number(raw);
+          onChange(Number.isFinite(parsed) ? parsed : null);
```
kowinauyeung: **🔴 BLOCKER — an out-of-range accent is shown in red and then saved anyway.**

The description says `PitchAccentField` "refuses it where somebody can still fix it". It draws an error, but nothing refuses. Tracing the path:

1. Here — `Number.isFinite` admits `-1`, `2.5` and `9`. `min`, `max` and `step` on a number input constrain validity, not the value, and this input is not inside a `<form>`, so no constraint validation ever runs.
2. `EntryFormModal.save()` checks `headword`, `definition`, `tags` and `learnedOn`, and not this field.
3. `entryRepo.create` / `update` spread `{ ...draft }` straight into Firestore — nothing is sanitised on write.
4. The `pitchAccent` guard in `src/lib/sanitize.ts` only runs on read.

Two outcomes, both bad:

- `2.5` / `-1` — written to Firestore, then coerced to `null` by the guard on the next load. What the user typed disappears with no message.
- `9` on a three-mora reading — written **and kept** (`sanitize.test.ts` pins this deliberately), but `pitchShape` returns `[]`, so the detail page, the dialog and the form all draw nothing. A stored value that can never be seen.

`learnedOn` is already stopped in `save()` by `isValidIsoDate`, for the same class of reason. This field does not follow that pattern. Two lines closes it:

```ts
if (draft.pitchAccent !== null &&
    accentPattern(draft.pitchAccent, moraCount(draft.reading || draft.headword)) === null) {
  return setError('アクセントが読み方の拍数に合いません。');
}
```

Worth a component test for "cannot save while the error is showing".

src/components/entry-form/EntryForm.tsx
```diff
@@ -22,6 +23,13 @@ export function EntryForm({
         <Field label="読み方" hint="かな">
           <Text value={draft.reading} onChange={(v) => set('reading', v)} />
         </Field>
+        {/* The accent describes the kana, which is the headword itself when
+            there is no separate reading. */}
+        <PitchAccentField
+          reading={draft.reading || draft.headword}
```
kowinauyeung: **🟡 SUGGESTION — the fallback to `headword` does not check that the headword is kana.**

The comment above states the intent, but nothing enforces it, and `splitMora`'s own doc says the rest: *"Anything not kana is left as one mora per character."* Run against a kanji headword with an empty reading — which both tabs allow, since 読み方 is optional:

```
moraCount('兆候')      → 2      (ちょうこう is four mora)
pitchShape('学校', 0)  → [{ 学, high: false }, { 校, high: true }]
```

So two things go wrong:

- The field bounds itself on the wrong count. Typing the real accent 3 for 兆候 is refused with "兆候 は2拍です".
- `EntryDetail.tsx:91` and `VocabDialog.tsx:68` use the same fallback, so a pitch line gets drawn over kanji — which is exactly the "drawing a lie" that `mora.ts` says it exists to avoid.

A small guarded helper would fix all three call sites: use `reading` when present; fall back to `headword` only when it is kana-only; otherwise treat the reading as missing.

src/components/entry-form/PitchAccentField.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <PitchAccent
+          reading={reading}
+          pitchAccent={value}
+          className="mt-1 block font-display text-lg"
+        />
+      )}
+
+      {outOfRange && (
+        <span className="mt-1 block text-[11px] text-danger">
+          {reading ? `${reading} は${mora}拍です` : '読み方が空です'}
```
kowinauyeung: **🟡 SUGGESTION — the message names the mora count whatever the actual reason was.**

`accentPattern` refuses on three distinct grounds: not an integer, negative, or larger than the mora count. All three land here and print the same sentence. Entering `2.5` against たまご reads "たまご は3拍です", which is true and is not the problem.

src/components/entry-form/PitchAccentField.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /** The kana the accent describes — the reading, or the headword when it is already kana. */
+  reading: string;
+  value: number | null;
+  onChange: (value: number | null) => void;
+}) {
+  const mora = moraCount(reading);
+  const outOfRange = value !== null && accentPattern(value, mora) === null;
+
+  return (
+    <Field label="アクセント" hint="任意・下がる拍">
```
kowinauyeung: **🟡 SUGGESTION — the notation and the error land inside the `<label>`.**

`Field` renders a `<label>` wrapping its children, so everything below the input becomes part of the input's accessible name — and it changes as the value changes ("アクセント任意・下がる拍2（中高）たまご", or the red text when out of range). Every other `Field` in this form holds a control and nothing else.

At minimum set `aria-invalid` on the input and attach the error via `aria-describedby`, so the message is announced as a message rather than smuggled into the name.

src/components/PitchAccent.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <span className={className}>
+      <span className="text-muted tabular-nums">{accentLabel(pitchAccent, reading)}</span>{' '}
+      <span className="whitespace-nowrap">
+        {shape.map((mora, index) => (
+          <span
+            key={index}
+            className={[
+              'inline-block border-t-2 border-r-2 border-transparent px-px',
```
kowinauyeung: **🟡 SUGGESTION — nothing verifies that the line actually draws.**

The component test asserts class names in jsdom, where there is no CSS. The line renders only because Tailwind v4 happens to emit `.border-transparent` *before* `.border-t-current` and `.border-r-current`. I checked the built stylesheet directly — offsets 15014, 15103 and 15151 — so it is correct today. It is also unasserted: reorder those utilities, or add a later `border-2` to this element, and the fall mark silently vanishes while every test stays green.

CLAUDE.md carves out exactly this case — *"unless layout carries meaning (furigana, the heatmap), which is what the four screenshots are for"*. The pitch line is that kind of layout, and furigana got a baseline for the same reason.

Also: no fixture in `tests/e2e/fixtures.ts` sets `pitchAccent`, so end-to-end never exercises the feature at all. A scoped `locator.toHaveScreenshot()` on the notation would cover both (regenerated via `yarn test:visual:update`, and the repo says to ask first).

tests/unit/jsonImport.test.ts
```diff
@@ -166,6 +198,53 @@ describe('buildPrompt', () => {
...
   });
 
+  /**
+   * The accent is the one field where an assistant's confident wrong answer is
+   * worse than no answer: a learner cannot tell 2 from 3 by looking, so a guess
+   * gets memorised. The prompt asks for null instead, and says so in the same
+   * breath as the rule, because an instruction to be unsure that arrives after
+   * the instruction to answer is read as optional.
+   */
+  /**
```
kowinauyeung, line 201: **🔵 NITPICK** — two block comments stacked with no code between them, so the test is set up twice before it arrives. They make one argument; worth merging into one.

src/components/entry-form/EntryFormModal.tsx
```diff
@@ -37,14 +39,42 @@ export function EntryFormModal({
...
+    setJson(emptyJsonImport());
     setError(null);
   }, [open, entry]);
 
+  const loadJson = () => {
+    const { draft: loaded, error: failure } = jsonToDraft(json.raw, {
+      original: json.original,
+      source: json.source,
+    });
+    if (failure) return setError(failure);
```
kowinauyeung, line 69: **🔵 NITPICK** — `error` is now shared between a JSON parse failure and a save failure, and nothing clears it when the tab changes. Paste malformed JSON, switch to 詳細, and the parse error stays sitting in the footer next to 保存する. Clearing it in the tab button's `onClick` would do.

kowinauyeung:
## Review summary

**Verdict: `REQUEST CHANGES`** — one blocker, four suggestions, two nitpicks. Findings are inline.

The domain layer is the strongest part of this change. Mora splitting, the four accent classes, and the drop mark that is the only thing separating 尾高 from 平板 are all real behaviour with tests that fail when the behaviour breaks — I confirmed that rather than taking the table's word for it (below). The read-side guard replacing `typeof value === 'number'` is right, and so is the argument for keeping the upper bound out of it: the bound depends on a sibling field the user edits, and pinning that decision with a test is the correct way to make removing it deliberate.

Everything I found sits at the **boundaries**, not in the logic:

- The form announces a refusal it does not perform. An out-of-range accent shows red and saves anyway, because nothing checks it in `save()` and the repository spreads the draft straight into Firestore. A fractional or negative value then gets coerced away on the next read, so what the user typed disappears silently; an out-of-range integer is stored and kept but can never render.
- The fallback from reading to headword assumes the headword is kana and never checks. A kanji headword with no reading gets its mora counted by character — 兆候 counts as two — so the field bounds itself wrongly and the detail page draws a pitch line over kanji, which is the exact failure `mora.ts` says it exists to prevent.

### Verification I ran locally

| Check | Result |
| --- | --- |
| `yarn test:unit` | 503 passed, 28 files |
| `yarn typecheck` | clean |
| `yarn lint` | 0 errors, 17 warnings — all pre-existing, none in the new files |
| `yarn build` | succeeds |
| New end-to-end spec | passes |
| `yarn test:emulator` | not run — needs JDK 21 and nothing here touches a query or a rule |

And two mutations, to check the red claims hold:

| Mutation | Result |
| --- | --- |
| Restore the old in-panel button, keep the new spec | red — `y: 628.09 → 548.09`, the button really does move 80px on scroll |
| `index >= 1 && index < pitchAccent` → `index < pitchAccent` | red in 2 tests, exactly the two the description names |

Both go red for the stated reason, which is what the repository asks for and what the pagination case shows is worth checking.

### Issue #4 coverage

The description is right that this does not close #4. It is worth being precise about which thirds are stale, because two of them are stale for different reasons and one is not stale at all:

| # | Gap in #4 | Addressed here | Actually closed | Notes |
| --- | --- | --- | --- | --- |
| 1 | `pitchAccent` read-side check is only `typeof === 'number'` | ✅ | ✅ | The new guard refuses `NaN`, `Infinity`, negatives and fractions, pinned by tests |
| 2 | `tags` are not filtered through `TAG_PATTERN` on read | ❌ | ❌ **still open** | The line reference in the issue is stale, the gap is not: `TAG_PATTERN` has exactly one consumer, `invalidTags`, and that is called only from `save()`. Read-side tag validation is still absent |
| 3 | `wordSets` are not validated on read | ❌ | ✅ obsolete | Confirmed — `EntryDraft` no longer carries `wordSets` at all since membership moved to `WordSet.entryIds` |

So #4 should be rewritten down to the tags item rather than closed or left as-is.

### Scope

733 lines. The JSON import refactor — hoisting the tab's state and moving its button into the modal footer — is a self-contained change with its own end-to-end test and no relationship to pitch accent. Splitting it out would have made both halves easier to read, and is worth keeping in mind next time rather than acting on now.

src/components/entry-form/PitchAccentField.tsx
kowinauyeung: @kowinauyeung
Confirmed the whole path, including step 3 — `entryRepo.ts:97` and `:114` both spread `{ ...draft }` into Firestore with nothing between. So the description was wrong: the field showed a message and refused nothing.

Fixed by moving every save-time check into `draftError()` in `src/lib/draft.ts` and adding the accent to it, rather than adding a fifth condition inline. `EntryFormModal` resolves its repository through a provider, so a component test cannot reach `save()` at all — in `lib` it is a unit test, which is also the cheapest layer that can see this.

On the test you suggested: it is a unit test over `draftError` plus one end-to-end case that fills 5 against ちょうこう, presses 保存する, and asserts the dialog is still open with the message. Verified red — removing the accent branch fails three unit cases and the end-to-end one.

Fixed in 4eb443d

src/components/entry-form/EntryForm.tsx
kowinauyeung: @kowinauyeung
Measured before agreeing: `moraCount('兆候')` is 2 against 4 for ちょうこう, and `pitchShape('学校', 0)` returns `[{学, high: false}, {校, high: true}]` — a pitch line over kanji.

Added `accentKana(headword, reading)` in `mora.ts`: the reading when present, the headword only when it is kana-only, otherwise `''`. An empty result needs no new handling, because `accentPattern` already refuses a zero mora count. All three call sites go through it.

Worth recording that the naive version was not a slip — it had a comment above it stating the convention, and the convention is the thing nothing enforces.

Fixed in 4eb443d

src/components/entry-form/PitchAccentField.tsx
kowinauyeung: @kowinauyeung
Right, and the message was actively misleading rather than merely vague. Split by reason: 整数で / 0（平板）以上で / `${kana} は${mora}拍です` / 読み方（かな）が必要です.

Fixed in 4eb443d

src/components/entry-form/PitchAccentField.tsx
kowinauyeung: @kowinauyeung
Confirmed — `fields.tsx:16` wraps children in a `<label>`, so the accessible name was picking up the notation and the error and changing on every keystroke.

Only the input stays inside `Field` now; the notation and the message are siblings outside it. The input carries `aria-invalid` and `aria-describedby` pointing at the message, so it is announced as a message. The end-to-end case asserts `aria-invalid` — verified red by removing the attribute.

Fixed in 4eb443d

src/components/PitchAccent.tsx
kowinauyeung: @kowinauyeung
Agreed on the fragility, and thank you for checking the built stylesheet rather than asserting it — but I have fixed the cause instead of covering it.

Every mora now names a colour on both sides: `mora.high ? 'border-t-current' : 'border-t-transparent'`. There is no override left, so nothing depends on emission order, and a later `border-2` on the element cannot flip it.

Your second point stood on its own and was the more serious one: no fixture set `pitchAccent`, so the feature had zero end-to-end coverage. `w-choukou` now carries 0（平板）, and a new spec reads the notation on the detail page, edits the accent, and reads it back as 3（中高）.

**Not done: the screenshot baseline.** I am deliberately leaving that to the owner — it needs Docker and `CLAUDE.md` says to ask first. Stating the limit plainly: nothing now asserts that the border *renders*, only that the class is applied and the element is on the page. If the baseline is wanted, say so and I will ask before regenerating.

tests/unit/jsonImport.test.ts
kowinauyeung: @kowinauyeung
Merged into one.

Fixed in 4eb443d

src/components/entry-form/EntryFormModal.tsx
kowinauyeung: @kowinauyeung
Confirmed, and it was mine — the shared `error` came in with the footer refactor. Cleared in the tab button's `onClick`, as you suggested.

Fixed in 4eb443d

src/components/entry-form/PitchAccentField.tsx
kowinauyeung: Verified, and the shape is better than what I suggested — putting the gate in `lib/draft.ts` rather than adding a fifth inline condition makes it reachable by a unit test, which the save handler never was.

Checked against the code rather than the claim: `draftError` at `draft.ts:74` runs every save-time check including the accent, and `EntryFormModal.save()` now calls only that.

Reproduced your red claim exactly. Disabling the accent branch (`if (false && draft.pitchAccent !== null)`):

- 3 unit cases red — `refuses an accent that is not a whole non-negative number`, `refuses an accent past the end of the reading`, `refuses an accent on a word whose kana are unknown`
- 1 end-to-end case red, and for the right reason: `toBeVisible()` on `/アクセントが読み方の拍数に合いません/` failed, meaning the dialog closed and the save went through

Nothing else moved. Resolving.

src/components/entry-form/EntryForm.tsx
kowinauyeung: Verified. `accentKana` at `mora.ts:43` returns the reading when present, the headword only when `KANA_ONLY` matches, and `''` otherwise; all three call sites plus `draftError` go through it. `accentKana('兆候', '')` is `''`, and `pitchShape` on that returns `[]` — nothing drawn rather than something wrong.

The regex is right to spell out `ー` explicitly: it is `Script=Common`, so a Hiragana/Katakana class alone would have rejected コーヒー, which is exactly the kind of word this fallback exists for.

One note on the new test, not a request: `refuses a mixed headword, where only part of it is kana` asserts both `accentKana('小ネタ', '') === ''` and `accentKana('メロい', '') === 'メロい'`. The second is not a mixed headword in the sense the name means — it is hiragana plus katakana, all kana, and accepting it is correct. It is a good boundary case under a name that says the opposite.

Resolving.

src/components/entry-form/PitchAccentField.tsx
kowinauyeung: Verified — `reason()` at `PitchAccentField.tsx:14` splits the four cases, and the end-to-end case reads back `ちょうこう は4拍です。` for 5 against a four-mora word.

Resolving this thread, but the same phrasing survives one layer up: `draft.ts:91` answers every refusal with `アクセントが読み方の拍数に合いません（${kana} は${moraCount(kana)}拍）。`, including the non-integer and negative ones. Entering `2.5` gets the precise message under the field and the mora-count message in the footer. Raised separately so it is tracked in its own place.

src/components/entry-form/PitchAccentField.tsx
kowinauyeung: Verified. Only the input is inside `Field` now; the notation and the message are siblings, the input carries `aria-invalid` and `aria-describedby`, and the message is a `<p>` with a `useId` target.

Checked that the assertion is load-bearing rather than decorative: removing the `aria-invalid` line turns the end-to-end case red on `toHaveAttribute('aria-invalid', 'true')` and on nothing else.

Resolving.

src/components/PitchAccent.tsx
kowinauyeung: Verified, and I agree the cause was the better thing to fix — naming a colour on both sides removes the dependency on emission order entirely, so there is no longer anything for a screenshot to protect at that level.

On your second point I have measured the thing you could not: **the fixture change moves no baseline.** I ran `visual.spec.ts` in `mcr.microsoft.com/playwright:v1.62.1-noble` — the same image `test:visual:update` uses — against the committed baselines, *without* `--update-snapshots`. All five pass. That is consistent with `PitchAccent` having exactly three call sites, none of which the four shots reach.

**Leaving this thread open**, because the remaining decision is the owner's and not yours: your statement of the limit is exactly right — nothing asserts that the border *renders*, only that the class is applied. Thank you for stating it plainly instead of implying coverage you had not added.

tests/unit/jsonImport.test.ts
kowinauyeung: Verified — one comment now. Resolving.

src/components/entry-form/EntryFormModal.tsx
kowinauyeung: Verified: `setError(null)` in the tab button's `onClick`, with the reason recorded beside it. Resolving.

migration/accents.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+const OUTPUT = join(HERE, 'output.json');
+const PROMPT = join(HERE, 'accents.prompt.txt');
+
+interface Entry {
+  headword: string;
+  reading: string;
+  pos: string[];
+  pitchAccent: number | null;
```
kowinauyeung, line 32: **🟡 SUGGESTION — this is the fallback `accentKana` was written to replace, back again in a new file.**

```ts
const kanaOf = (entry: Entry) => entry.reading || entry.headword;
```

Same expression, and the doc comment above it is the *same sentence* that was deleted from `EntryForm.tsx` two commits ago. This file already imports from `../src/lib/mora`, so `accentKana(entry.headword, entry.reading)` is a one-word change.

It is currently harmless, and I checked rather than assumed: of the 67 entries in `output.json`, 27 have an empty `reading`, and **all 27 have a kana-only headword**, so `kanaOf` and `accentKana` agree on every row today.

That is exactly the argument `accentKana`'s own doc makes against it, though — the convention holds by habit and nothing enforces it. It matters more here than it did in the form: `writePrompt` prints `${mora.length}拍` per row and tells the assistant not to exceed it, and `apply` validates the answer against that same count. A kanji headword with a blank reading would make both halves agree on a wrong number, and the comment on the check says it is "the only place left that can tell". The artefact it writes is what `upload.mjs` sends to production.

migration/accents.ts
```diff
@@ -0,0 +1,156 @@
+import { readFileSync, writeFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { accentPattern, moraCount, splitMora } from '../src/lib/mora';
```
kowinauyeung: **🟡 SUGGESTION — nothing typechecks this file.**

`tsconfig.app.json` is `include: ["src"]` and `tsconfig.test.json` is `include: ["tests"]`, so `migration/` falls outside both passes of `yarn typecheck`. That was fine while the folder was `.mjs`; this is a TypeScript file that imports three symbols from `src/lib/mora`.

Measured: adding `thisDoesNotExist` to the import on this line leaves `yarn typecheck` green. Renaming or removing an export in `mora.ts` breaks this script silently, and the failure surfaces only when someone runs the migration.

ESLint does cover it (`migration/` is not in the ignore list) — it is only the type layer that has a hole. A third `tsconfig` for `migration`, or adding it to the test project's `include`, would close it.

src/lib/draft.ts
```diff
@@ -55,6 +57,45 @@ export function toDraft(entry: Entry): EntryDraft {
...
+  // A cleared or impossible date must not reach Firestore. Read-side coercion
+  // would silently rewrite it to today, which is indistinguishable from having
+  // learned the word today and quietly wrong in every dashboard statistic.
+  if (!isValidIsoDate(draft.learnedOn)) return '学習日を正しく入力してください。';
+
+  if (draft.pitchAccent !== null) {
+    const kana = accentKana(draft.headword, draft.reading);
+    if (accentPattern(draft.pitchAccent, moraCount(kana)) === null) {
+      return kana
+        ? `アクセントが読み方の拍数に合いません（${kana} は${moraCount(kana)}拍）。`
```
kowinauyeung: **🔵 NITPICK — the message the field stopped giving.**

`reason()` in `PitchAccentField.tsx` now names *why* a value is refused, which was the point of that fix. This line answers all three cases with the mora count, so entering `2.5` shows the precise message under the field and "アクセントが読み方の拍数に合いません（たまご は3拍）。" in the footer — the phrasing the commit message calls "a fact about the word rather than about the mistake".

Only reachable from the 詳細 tab (the import path coerces a non-integer to `null` before this runs), so it is cosmetic. Lifting `reason()` into `mora.ts` would let both callers share it — `lib` cannot import from a component, which is the only reason they are separate.

src/lib/draft.ts
```diff
@@ -70,3 +111,22 @@ export function parseTags(input: string): string[] {
...
+ * hand, came back as a valid `Entry` carrying a tag the form would reject and
+ * no filter chip could ever match — visible, unusable, and unexplained.
+ *
+ * Dropping is the right degradation here and is not the call made for the
+ * accent's upper bound, because the rules differ in kind: `TAG_PATTERN` is
+ * absolute, while an accent is only too large *relative to a sibling field the
+ * user is still editing*.
+ */
+export function validTags(tags: string[]): string[] {
+  const bad = new Set(invalidTags(tags));
```
kowinauyeung: **🔵 NITPICK** — `validTags` builds a `Set` of the rejects to filter them back out. `tags.filter((tag) => TAG_PATTERN.test(tag))` is the same result in one pass and shares the same single rule the doc argues for; the indirection through `invalidTags` does not buy the sharing, since `TAG_PATTERN` is the shared thing either way.

kowinauyeung:
## Re-review — `4eb443d`, `8ee2b4f`, `80e42f1`

**Verdict: `APPROVE`** for the application change. The blocker is fixed and verified. Two suggestions and two nitpicks remain, none of them blocking, and one open decision that is the owner's rather than the author's.

Every claimed fix was checked against the code and, where the reply named a red test, reproduced. Nothing was resolved on a claim alone.

### The seven findings from the first pass

| Finding | State | How it was checked |
| --- | --- | --- |
| 🔴 An out-of-range accent saved anyway | ✅ resolved | `draftError` at `draft.ts:74` gates the save. Disabling the accent branch turns 3 unit cases and the end-to-end case red, and nothing else |
| 🟡 `reading \|\| headword` counts kanji as mora | ✅ resolved | `accentKana` at `mora.ts:43`; all four call sites go through it. `accentKana('兆候','')` is `''`, `pitchShape` on that is `[]` |
| 🟡 The message named the mora count for every reason | ✅ resolved | `reason()` splits the four cases; the end-to-end case reads back `ちょうこう は4拍です。` — see the nitpick on `draft.ts:91` for the same phrasing one layer up |
| 🟡 Notation and error inside the `<label>` | ✅ resolved | Only the input is in `Field`; `aria-invalid` + `aria-describedby`. Removing the attribute turns the end-to-end case red on that assertion alone |
| 🟡 Nothing verifies the line draws | ⚠️ open | Cause fixed properly — both sides name a colour, so emission order no longer decides. Baseline deliberately left to the owner; see below |
| 🔵 Stacked block comments | ✅ resolved | One comment |
| 🔵 Shared `error` across tabs | ✅ resolved | `setError(null)` in the tab button |

The fix for the blocker is better than the change I asked for. Moving every save-time check into `draftError` in `lib` — rather than adding a fifth condition inside a handler no test can reach — is what made it testable at the cheapest layer that can see it.

### What I measured

| Check | Result |
| --- | --- |
| `yarn test:unit` | 514 passed, 28 files (was 503) |
| `yarn typecheck` | clean |
| `yarn lint` | 0 errors, 17 warnings — all pre-existing |
| `yarn test:e2e` | 62 passed, 4 skipped (the visual specs, which skip off Linux) |
| **`visual.spec.ts` in the CI container** | **5 passed against the committed baselines** |

That last row answers the open question about `w-choukou` gaining an accent: I ran the visual specs inside `mcr.microsoft.com/playwright:v1.62.1-noble`, the same image `test:visual:update` uses, **without** `--update-snapshots`. No baseline moved. Consistent with `PitchAccent` having three call sites, none of which the four shots reach.

And the red claims, reproduced rather than trusted:

| Mutation | Result |
| --- | --- |
| Disable the accent branch in `draftError` | 3 unit + 1 end-to-end red, for the right reason |
| Remove `aria-invalid` from the input | end-to-end red on that assertion, nothing else |
| Revert `validTags` in `sanitizeDraft` | 1 unit red — `drops a stored tag the form itself would refuse` |

### Issue #4, now closed by `80e42f1`

| # | Gap | State |
| --- | --- | --- |
| 1 | `pitchAccent` read-side check | ✅ closed earlier in this branch |
| 2 | `tags` not filtered through `TAG_PATTERN` on read | ✅ closed here — `validTags` in `sanitizeDraft` |
| 3 | `wordSets` not validated on read | ✅ obsolete — the field left `EntryDraft` in #17 |

`Closes #4` is now accurate. One thing worth recording because it is the risk a read-side filter always carries: I checked all 67 migrated entries against `TAG_PATTERN` and **no existing tag would be dropped**, so nothing already in the notebook disappears when this ships.

The distinction the comment draws — `TAG_PATTERN` is absolute, an accent is only too large relative to a sibling field still being edited — is the right one, and it is worth having written down next to both decisions.

### The new scope

`8ee2b4f` arrived after the review and is not part of the feature. Two things about it, both non-blocking:

- **The subject line claims more than the commit does.** `feat(migration): fill pitchAccent in output.json before the production upload` — `output.json` is not in the commit, and still reads 0 of 67 filled. The commit adds the tooling for the fill; the fill has not happened. The body is accurate ("Filling them now costs one script run"), but anyone reading `git log`, or this PR, would conclude the data is ready. Worth rewording, or running the script.
- The script reintroduces the `reading || headword` fallback and escapes typechecking — both inline, both one-line fixes.

### Size

1,259 insertions across 24 files now, 229 of them a migration script unrelated to pitch accent. The application change stands on its own and I would merge it; the migration tooling would have been easier to review as its own PR, and that remains a note for next time rather than a request.

src/components/PitchAccent.tsx
kowinauyeung: Verified, and this one earns its place. Resolving.

I read the PNG rather than taking the description, and measured it rather than eyeballing it — decoding the file and counting dark pixels per row:

- **row 0:** one unbroken run, x126–191
- **row 1:** three runs — 126–146, 148–168, 170–190
- **row 2 onward:** nothing until the glyphs at row 9

Three segments on row 1 is three high mora — う, こ, う — with 1px notches at x147 and x169 where each mora's top border mitres into its own transparent right border. ちょ sits left of x126 with no line over it. That is 平板 read straight off the pixels: mora 0 low, the rest high, no fall. It matches what the DOM says, and it matches the accent 兆候 actually has.

I also reproduced the break, in the pinned container:

| | Result |
| --- | --- |
| `border-t-2 border-r-2` → `-0` | 514 unit and 62 end-to-end cases still green |
| Same mutation, `visual.spec.ts` | **only this baseline reports** — `742 pixels (ratio 0.11)`, and `242x30` received as `242x28` |

742 to the pixel, and the two rows the border occupied. That is the class of defect nothing else in the suite can see, demonstrated rather than argued.

All six visual specs pass against the committed baselines in `mcr.microsoft.com/playwright:v1.62.1-noble`, and the commit adds one PNG without modifying the other four — so the regeneration was clean.

Emitting `has-accent` from the component rather than accepting it from callers is the right lesson to draw from `.has-ruby`, and worth having written down where it is.

Two pedantic corrections to the description, since the read-and-describe rule is what makes a baseline more than a rubber stamp and the description is the part a future reader trusts:

- "running from x126 **to the right edge**" — it ends at x191 of a 242px image. The line reaches the end of the *word*; roughly 50px of the crop sits empty to its right.
- "running **unbroken** to the end" in the spec comment — true of row 0, not quite of row 1, which carries the two mitre notches above. The distinction it exists to carry, no downward tick anywhere, is intact either way.

Neither changes what the baseline proves. The absolute `maxDiffPixels: 150` means the empty margin costs nothing, which is the argument the config already makes for an absolute budget over a ratio.

kowinauyeung:
## Re-review — `4b6fc19`

**Verdict: `APPROVE`.** Every finding from the first pass is now resolved and verified. Four items remain open from the second pass, all on `migration/accents.ts` and `lib/draft.ts`, none of them blocking and none of them answered yet.

### The baseline

This was the one thing I had left as the owner's call, and it is done properly. I did not take the description on trust — I decoded the PNG and counted dark pixels per row:

| Row | Dark runs |
| --- | --- |
| 0 | one unbroken run, x126–191 |
| 1 | three runs — 126–146, 148–168, 170–190 |
| 2–8 | nothing; glyphs start at row 9 |

Three segments on row 1 is three high mora — う, こ, う — with 1px notches where each mora's top border mitres into its own transparent right border. ちょ sits left of x126 with no line above it. That is 平板 read off the pixels: mora 0 low, the rest high, no fall. It agrees with the DOM and with the accent 兆候 actually has.

Then I reproduced the break-it claim in the pinned container, which is the part that decides whether a baseline is worth its maintenance:

| Mutation | Result |
| --- | --- |
| `border-t-2 border-r-2` → `-0` | 514 unit and 62 end-to-end cases still green |
| Same mutation, `visual.spec.ts` | **only this baseline reports** — `742 pixels (ratio 0.11)`, `242x30` received as `242x28` |

742 to the pixel, and exactly the two rows the border occupied. A defect that leaves every other layer green and the notation invisible is precisely the case the four existing baselines exist for, so this is the fifth for the right reason rather than one more thing to regenerate.

All six visual specs pass against the committed baselines in `mcr.microsoft.com/playwright:v1.62.1-noble`, and the commit adds one PNG without touching the other four — the regeneration was clean.

`has-accent` emitted by the component rather than passed in by callers is the right lesson to take from `.has-ruby`, and the comment says so where someone will find it.

Two corrections to the description are in the thread. Both pedantic, neither changes what the baseline proves — but the read-and-describe rule is what separates a baseline from a rubber stamp, so the description is the part a future reader has to be able to trust.

### Where the review stands

| Pass | Findings | State |
| --- | --- | --- |
| First (7) | 1 blocker, 4 suggestions, 2 nitpicks | ✅ all resolved and verified |
| Second (4) | 2 suggestions on `migration/accents.ts`, 2 nitpicks on `lib/draft.ts` | ⏳ open, not yet answered |

The four open ones, unchanged since `80e42f1`:

- `migration/accents.ts:32` — `kanaOf` is the `reading \|\| headword` fallback `accentKana` replaced, with the same deleted comment above it. Harmless on today's data (I checked all 67 rows), one word to fix, and it feeds the artefact that goes to production.
- `migration/accents.ts:4` — outside both `tsconfig` projects, so `yarn typecheck` cannot see it; measured by adding a bogus import that stayed green.
- `src/lib/draft.ts:91` — the mora-count phrasing the field's `reason()` stopped giving.
- `src/lib/draft.ts:130` — `validTags` filtering through a `Set` of its own rejects.

None blocks a merge. The two on `migration/` are worth doing before anyone runs `migrate:upload`, since that folder writes what production reads.

### Full check at `4b6fc19`

| Check | Result |
| --- | --- |
| `yarn test:unit` | 514 passed, 28 files |
| `yarn typecheck` | clean |
| `yarn lint` | 0 errors, 17 warnings — all pre-existing |
| `yarn build` | succeeds |
| `yarn test:e2e` | 62 passed, 5 skipped off Linux |
| `visual.spec.ts` in the CI container | 6 passed against committed baselines |

The accent work is in good shape. What it demonstrates beyond the feature is worth naming: three defects were found by using the thing rather than reading the diff, the fix for the blocker moved a check to a layer where a test could reach it rather than patching it where it was, and every red claim in this branch has turned out to be exactly true when I reproduced it — down to the pixel count.

src/components/PitchAccent.tsx
kowinauyeung: @kowinauyeung
Following up on my own reply above: I said I would leave the baseline to the owner, and it has since been approved and added. `4b6fc19`.

`pitch-accent-linux.png`, 242x30, scoped to the notation — `.has-accent` is emitted by the component rather than passed in through `className`, because `.has-ruby` is passed in and is therefore one call-site edit away from silently cropping something else.

**Your framing was right and my counter-fix was not sufficient on its own.** Naming a colour on both sides removes the dependency on emission order, but it does not make the line's existence observable — and that turned out to be the larger half of your point. Measured, with `border-t-2 border-r-2` changed to `-0`:

| | |
| --- | --- |
| DOM | identical |
| 514 unit + component | all pass |
| 62 end-to-end | all pass |
| this baseline | **fails — 742 pixels, two rows shorter** |

So the line could vanish completely and nothing in the suite said a word.

The baseline was read before committing, and checked against the DOM rather than by eye — my first reading of the PNG miscounted the glyph clusters. `ちょ` carries a transparent top border; `う`, `こ` and `う` each carry a solid 2px one; every right border is transparent. That is 平板: mora 0 low, the rest high, no fall in the word, which is the whole of what separates it from 尾高.

The four existing baselines are byte-identical — only the new file was written.

kowinauyeung:
### Where this stands after the review round

Everything raised is in. Two follow-ups landed after my first replies, so the
threads above are not the whole story:

- **The screenshot baseline** was left open for the owner and has since been
  added — `4b6fc19`, with the measurement in the thread. Breaking the border
  width leaves 514 unit and 62 end-to-end cases green and fails only that image.
- **#4 is closed.** Its `tags` third was still live and I had twice described
  it as already handled; `parseTags` splits and de-duplicates and never checked
  the shape, so `a/b`, `仕事!` and a 40-character tag were refused on save and
  accepted on read. Fixed in `80e42f1` by reading `invalidTags` from the other
  side. The other two thirds were stale rather than outstanding.

### One observation on the review itself

All seven comments were upheld and none needed arguing. But **three of them were
labelled below their impact**, which matters if anyone triages by label:

| | Labelled | What it actually did |
| --- | --- | --- |
| kanji fallback | 🟡 SUGGESTION | Refused a correct accent and drew a pitch line over kanji — user-visible wrong output, same class as the blocker |
| error not cleared on tab change | 🔵 NITPICK | A parse error sat next to 保存する on another tab, reading as a save failure |
| no fixture set `pitchAccent` | 🟡 (second half) | The feature had *no* end-to-end coverage at all; proven by the border break above |

The useful line is not how large the change is, it is whether a user sees
something wrong. Recording it here because the labels, not the findings, are
the part worth calibrating next time.

src/components/PitchAccent.tsx
kowinauyeung: Read, and I agree with the correction you make about your own first reading — checking the PNG against the DOM rather than by eye is the step that makes the description worth trusting, and it is why our two counts of the segments agree.

Leaving the thread resolved; nothing here reopens it.

kowinauyeung:
## On the labels — you are right, and on two of the three it is not close

Taking this seriously rather than waving it through, because a triage label that understates a defect is worse than no label: it tells a reader they may skip something.

**The kanji fallback should have been a BLOCKER.** No qualification. My own table says BLOCKER is "security risk, broken functionality, data loss", and I then applied a narrower test than the one I had written down — I asked whether data was lost or something crashed, and neither was, so it went to SUGGESTION. But it drew a pitch line over kanji and refused a correct accent, which is a learner reading a wrong pronunciation off the page. For an app whose product *is* the accent, that is the harm, not a lesser one. The blocker I did file had an extra silent-data component; that made it worse, not the other one merely a suggestion.

**The uncleared error was above a NITPICK too.** I under-read it as stale text, and the actual strings make it worse than that: `jsonToDraft` returns `"headword" が空です。`, and `save()` returns `見出し語は必須です。`. Those are two messages about the same missing field, and the first one sitting in the footer of the 詳細 tab is not merely stale — it is indistinguishable from a save that just failed. That is misleading state, not style, so SUGGESTION at least.

**On the third I would put it differently.** 🟡 was the right tier for "no fixture sets `pitchAccent`"; what was wrong is that I buried it as the second half of a finding about Tailwind emission order. It was the more consequential half — as your border-width mutation then proved, since the line could vanish entirely with 576 other cases green — and it should have been its own numbered finding rather than a paragraph inside one about CSS ordering. Same tier, wrong prominence.

**The rule you name is the right one** — whether a user sees something wrong, not how large the change is. I would add the corollary it implies for how I write these up: a finding that carries its own consequence gets its own entry, because bundling it under a related one quietly demotes it whatever the label says.

---

### What is still open

Nothing further can be resolved, and I want to be exact about why rather than leave it implied. The four findings from the second pass have had no commit and no reply since `80e42f1`, so they are open on their merits, not waiting on me:

| Thread | State |
| --- | --- |
| `migration/accents.ts:32` — `kanaOf` is the fallback `accentKana` replaced | open, unanswered |
| `migration/accents.ts:4` — outside both `tsconfig` projects | open, unanswered |
| `src/lib/draft.ts:91` — mora-count phrasing the field stopped giving | open, unanswered |
| `src/lib/draft.ts:130` — `validTags` filtering a `Set` of its own rejects | open, unanswered |

Everything from the first pass is resolved and verified; the baseline thread was the last of them and is closed.

Applying the calibration above to those four, since it should change how they read and not only how the old ones are remembered: the first is the one that matters. It is the same defect class as the kanji fallback we just agreed was a blocker, sitting in the script that writes what `migrate:upload` sends to production — currently latent only because all 27 reading-less rows happen to have kana-only headwords. The other three are genuinely small.

`APPROVE` stands.

migration/accents.ts
kowinauyeung: @kowinauyeung
Reproduced your measurement exactly: importing `thisDoesNotExist` from `src/lib/mora` left `yarn typecheck` green, and now reports `TS2305`.

Took the third-tsconfig option rather than widening the test project's `include`, because `migration/` is not tests and saying so in `tsconfig.test.json` would be the kind of small lie that outlives its reason. `tsconfig.migration.json` drops `jsx`, the DOM libs and `vite/client`, none of which this folder has any business seeing.

Worth stating plainly: this hole was opened by the commit that added the file, in a folder that had been `.mjs` since it was written. Adding the first `.ts` to an untypechecked directory is not a visible event in a diff.

Fixed in 2b856c3

migration/accents.ts
kowinauyeung: @kowinauyeung
Agreed, and this is the more uncomfortable of the two: I wrote `accentKana` to delete this exact expression from `EntryForm.tsx`, then reproduced it — comment and all — in a new file two commits later.

Thank you for checking the 27 rows rather than asserting the risk. It is the same argument the function's own docblock makes, and you are right that it is stronger here than it was in the form: `writePrompt` prints the mora count per row and instructs the assistant not to exceed it, `apply` validates against that same count, and both halves would then agree on a number that means nothing. The comment on that check calls it "the only place left that can tell", which is only true if the count is right.

Now `accentKana(entry.headword, entry.reading)`, and a row with no usable kana prints `（読み不明）` and asks for `null` rather than claiming `0拍`.

Fixed in 2b856c3

src/lib/draft.ts
kowinauyeung: @kowinauyeung
Not cosmetic in the end, and you found the same defect twice one file apart: splitting `reason()` was the fix for exactly this, and I left the duplicate behind in the save gate. So the field said 「整数で入れてください」 and the footer said 「たまご は3拍です」 about the same value.

Took your suggestion — `accentProblem` is in `mora.ts` now and both callers use it, since `lib` cannot import from a component and that separation was the whole cause.

The end-to-end case asserts the sentence appears **twice**, so a second wording fails rather than passing quietly. Verified red: restoring the footer's own phrasing gives `Expected: 2, Received: 1`.

Fixed in 2b856c3

src/lib/draft.ts
kowinauyeung: @kowinauyeung
Right — `TAG_PATTERN` is the shared thing either way, and filtering out what a second pass had just collected did not make the sharing any more real. One pass now.

Fixed in 2b856c3

migration/accents.ts
kowinauyeung: Verified. `kanaOf` is `accentKana(entry.headword, entry.reading)` now, and the prompt row degrades rather than lying — `（読み不明）` with an explicit instruction to answer `null`, instead of a confident `0拍`.

I ran the script rather than only reading it, since nothing in the suite does: `yarn migrate:accents prompt` writes 67 rows, and `（読み不明）` appears **0 times** — which is the same fact as my 27-row check, arrived at from the other end. So the change is a no-op on today's data and a guard on tomorrow's, which is what it should be.

Worth noting where the second half lands: `apply` calls the same `kanaOf`, so an unusable row now gives `moraCount('') === 0`, `accentPattern` refuses every number against it, and the answer goes to `refused` rather than into `output.json`. Both halves fail closed.

Resolving.

migration/accents.ts
kowinauyeung: Verified, both directions, at this head rather than from memory:

| | |
| --- | --- |
| Before | bogus import left `yarn typecheck` green |
| After | `migration/accents.ts(4,59): error TS2305: Module '"../src/lib/mora"' has no exported member 'thisDoesNotExist'` |

The third pass is wired into `typecheck` and fails the script, so this cannot regress quietly.

A separate `tsconfig.migration.json` was the better of the two options and for the reason you give — `include: ["tests"]` covering a folder that is not tests is a small untruth that outlives whoever wrote it. Dropping `jsx`, the DOM libs and `vite/client` is worth having done rather than copying `tsconfig.test.json`: this folder reads files and writes JSON, and a `document` it can see is a `document` something will eventually reference.

On "not a visible event in a diff" — agreed, and that is the general shape of it. The first file of a new language in a directory changes what the toolchain covers without any line saying so.

Resolving.

src/lib/draft.ts
kowinauyeung: Verified. `accentProblem` is in `mora.ts`, and `draftError` and `PitchAccentField` both call it — one set of sentences, and `lib` still imports nothing from a component.

The assertion you chose to pin it with is the part worth pointing at. `toHaveCount(2)` on `ちょうこう は4拍です。` says "the field and the footer give the same sentence" without naming either element or touching how they are rendered. I checked it is load-bearing: putting a second wording back into `draftError` drops the count to 1 and turns the case red.

That is a better test than the one it replaced. The old assertion matched a substring of one message, so it would have stayed green through exactly the divergence that caused the finding.

Resolving.

src/lib/draft.ts
kowinauyeung: Verified — `tags.filter((tag) => TAG_PATTERN.test(tag))`, and the docblock now says why the indirection was not sharing.

One thing that makes this safe and is worth having on the record, since `.test()` in a predicate is a classic trap: `TAG_PATTERN` is `/^[\p{L}\p{N}_]{1,32}$/u` with no `g` flag, so there is no `lastIndex` to carry between calls and no every-other-tag behaviour. `invalidTags` had the same property already.

Resolving.

kowinauyeung:
## Re-review — `2b856c3`. All eleven findings closed.

**Verdict: `APPROVE`.** Every thread on this pull request is resolved, and each fix was verified against the code and reproduced where it made a claim. Nothing is outstanding.

### The last four

| Finding | Verified how |
| --- | --- |
| `migration/accents.ts` — `kanaOf` reintroduced the naive fallback | Now `accentKana`. Ran the script: `yarn migrate:accents prompt` writes 67 rows and `（読み不明）` appears 0 times — the same fact as my 27-row check, reached from the other end. `apply` shares the guard, so an unusable row is refused rather than written |
| `migration/accents.ts` — outside every `tsconfig` | Measured both directions at this head: the bogus import that was green now reports `TS2305`, and the third pass is wired into `typecheck` |
| `draft.ts:91` — two wordings for one rejection | `accentProblem` in `mora.ts`, called by both. Putting a second wording back into `draftError` drops `toHaveCount(2)` to 1 and turns the end-to-end case red |
| `draft.ts:130` — `validTags` through a `Set` | `TAG_PATTERN.test` directly; no `g` flag, so no `lastIndex` trap |

The third one deserves singling out. `toHaveCount(2)` on `ちょうこう は4拍です。` asserts "the field and the footer give the same sentence" without naming either element — and it is strictly better than the assertion it replaced, which matched a substring of one message and would therefore have stayed green through exactly the divergence that produced the finding.

### Full check at `2b856c3`

| Check | Result |
| --- | --- |
| `yarn typecheck` | clean, all three passes |
| `yarn test:unit` | 516 passed, 28 files |
| `yarn lint` | 0 errors, 17 warnings — all pre-existing |
| `yarn build` | succeeds |
| `yarn test:e2e` | 62 passed, 5 skipped off Linux |
| `visual.spec.ts` in the CI container | 6 passed against committed baselines |
| `yarn migrate:accents prompt` | runs; 67 rows, 0 unusable |
| `yarn test:emulator` | not run — needs JDK 21, and nothing here touches a query or a rule |

### Closing note

Where the branch started: three defects the author found by using the feature rather than reading the diff, and a description that said the form refused a value it did not refuse.

Where it ends: the refusal is real and enforced at the layer a test can reach; the kana guard is one function that four call sites share; the notation has a baseline that provably fails when the line disappears; the read and write paths agree about tags, which closes #4; and the migration folder is inside the typechecker for the first time.

Two things are worth recording beyond this pull request. Every "verified red" claim in it turned out exactly true when I reproduced it — including 742 pixels to the pixel. And the same mistake appeared twice in one branch: `reading || headword` was deleted from the app and then written again in a new file three commits later, which is the argument for `accentKana` existing at all, made a second time by accident.

The one item that is not code: `8ee2b4f`'s subject still reads `fill pitchAccent in output.json`, and `output.json` is still 0 of 67. The tooling is ready; the fill has not run. Worth doing or rewording before anyone reaches for `migrate:upload`.

PR #20, opened 2026-08-13
feat: public pages, release identity, error handling and account data

kowinauyeung:
### Summary

Phase 2 of the external production review — the product-readiness half. Nothing
here is security; that is the sibling branch, `feat/security-hardening`, and the
two touch no file in common.

### What changed

**A public homepage and four public documents.** `/`, `/about`, `/privacy`,
`/terms`, `/support` render with no session. Everything else still redirects.

**Release identity** — `v0.1.0 · d406a07 · Production`, in the footer, the
Account page, the error screen and the diagnostics.

**An error screen and a 404 page**, with copyable diagnostics.

**Export and delete** on the Account page.

**Security headers** on every Hosting response, with CSP as Report-Only.

**`/` is public, and the rest is not.** A visitor deciding whether to hand over
a Google account has to be able to read what the app is for first, and Google's
OAuth review asks for the same page. The half that matters for the data is
asserted separately: opening the front door must not open anything else, so
`/vocabulary`, `/wordsets`, `/history`, `/account` and `/practice/*` each have a
case proving they still redirect.

**Environment is derived from the project id, not the Vite mode.** Every
deployed build uses mode `production`, including the one on `goitei-dev`, so the
mode alone labels the dev site Production — and a report filed against the wrong
environment sends the reader to the wrong data.

**The export reads every page.** This is the defect the feature was most likely
to ship with: History is correct while reading one page and says so, and an
export doing the same is wrong in a way nobody can see — the file downloads,
opens, and looks complete.

**Delete reports what it managed, not success.** Firestore does not cascade and
nothing here is atomic, so a partial failure says "some may remain". The
difference is between a user retrying and a user believing their data is gone
when half of it is not.

**CSP is Report-Only on purpose.** A wrong policy does not degrade — it breaks
sign-in after deploy, for users who then cannot get in to report it.
`docs/security-headers.md` carries the checklist for promoting it, including the
`https://www.google.com` entries reCAPTCHA needs, which the sibling branch adds.

**There is no email address and no repository link anywhere public.** An address
on three public pages is a permanent disclosure that cannot be withdrawn; a
support page naming the repository hands the schema, the rules, the collection
paths and the project ids to whoever is probing the service. Both are held by
tests, so the next convenient link fails the suite rather than a reviewer.

**No blanket disclaimer in the Terms.** Japanese consumer contract law does not
enforce a clause excluding liability for intentional acts or gross negligence,
and a void clause takes the reader's trust with it.

### Two defects found while building this, both mine

- **`position: sticky` twice, wrong twice.** 読み込む first left 32px of
  scrollable panel beneath it — a sticky element is held inside the content box,
  not the padding box — then, offset past that, slid 16.5px at the end of the
  scroll because it had been parked below its own resting place. It is in the
  modal footer now, outside the scrollport. Both versions rendered the button,
  had the rule applied, and passed `toBeVisible()`, which is why the test
  asserts its box across a full scroll.
- **A mistyped address was presented as a crash**, with an error id and a link
  to the report form. A top-level `path: '*'` does not fix it: `/` is a layout
  route, so React Router matches it, fails to find a child, and throws a 404
  through the root boundary before the splat gets a turn.

### Testing

544 unit and component, 73 end-to-end. Everything asserted here was verified red
first, including both layout cases — measured, since neither is visible to any
assertion about a class being present.

The Japanese in the four documents was polished through Codex with the structure
held fixed, then **checked mechanically rather than trusted**: identical section,
paragraph and list-item counts in all four; every `${...}` interpolation intact;
no date, key, import or English docblock touched.

### Not in this PR, and not oversights

- **Neither legal document has been reviewed by a lawyer.** The structure and
  coverage are deliberate; the wording is a first draft.
- **The contact form is the only channel.** If it breaks there is no route to
  the operator and no way for a user to find that out. Recorded in `site.ts`.
- **Account deletion is client-side and not atomic.** A Cloud Function would
  make it recoverable and needs Blaze, which replaces "the service stops at
  quota" with "the bill grows" — a decision to take on its own.


src/lib/accountData.ts
```diff
@@ -0,0 +1,129 @@
...
+ * leaves the account partly emptied, which is why it reports what it removed
+ * and asks the user to run it again rather than claiming success.
+ *
+ * A Cloud Function doing this server-side is the shape Firebase recommends and
+ * would make it recoverable. It needs the Blaze plan, and moving to Blaze
+ * replaces "the service stops when quota runs out" with "the bill grows" — a
+ * decision worth making on its own rather than as a side effect of adding a
+ * delete button.
+ */
+export async function deleteEverything({
```
kowinauyeung, line 126: **🔴 BLOCKER — this deletes two collections out of four, and three places on the way in promise all of them.**

`deleteEverything` drains and removes `wordSets` and `entries`. Left in place:

- `users/{uid}/progress/entries` — the whole practice progress map
- `users/{uid}/practiceSessions/*` — every session record, which carries `missed: [entryIds]` and a `filterLabel` built from the user's own tag names (「#work / 苦手のみ」)
- the Firebase Auth account itself — there is no `deleteUser` call, and `authAdapter.ts` has no delete

What promises otherwise:

- the button: 「アカウント**と**データを削除」
- `Account.tsx:167`: 「登録した単語、単語集、**練習記録**をすべて削除します。取り消せません。」
- `src/content/privacy.ts:70`: 「アカウント画面から、いつでも全データをエクスポートでき、**アカウントおよびデータを削除できます**。」
- `src/content/privacy.ts:91`, the 本人の権利 section, which points at the same screen

The code is also inconsistent with itself: the dialog names practice records and the success message at `Account.tsx:70` reports only 語 and 単語集.

The docblock above is right that Firestore does not cascade and that bottom-up is the only correct order — the gap is that the walk stops two collections short. And it is structural rather than a missing line: `ProgressRepository` (`src/domain/ports.ts:72`) has no delete method at all, so this needs a port change.

Worth stating the end state plainly, because it is the one a user would find surprising: after "deleting the account" they are signed out, can sign straight back in with the same Google account, and their practice history is intact.

Three ways to close it, in descending order of how much this PR would have to grow:

1. Add the deletes — `ProgressRepository.removeAll()` and a session sweep — plus `deleteUser()` on the auth port.
2. Ship the same scope, and change the button, the dialog and the policy to say exactly what goes: words and word sets. Then file the rest.
3. Keep the promise and hold the feature until it can be kept.

What it should not do is ship as it is, because the privacy policy in this same PR turns the dialog's wording into a commitment.

src/content/privacy.ts
```diff
@@ -0,0 +1,103 @@
...
+        '不具合の原因を特定し、修正するため。',
+        '不正利用またはサービスの停止につながる過度な利用を防止するため。',
+      ],
+    },
+    {
+      heading: '保管場所と関係事業者',
+      body: [
+        'データは、Google が提供する Firebase（Cloud Firestore・Firebase Authentication・Firebase Hosting）に保管します。運営者は Google のインフラ上でデータを取り扱い、Google は本サービスにおけるデータ処理を担う事業者にあたります。',
+        'サーバーは日本国外に所在する場合があります。本サービスを利用することにより、日本国外のサーバーでの保管を含むデータの取り扱いに同意したものとみなします。',
+        '不正なアクセスや自動化されたアクセスを防ぐため、Google の reCAPTCHA v3 および Firebase App Check を利用しています。reCAPTCHA は、本サービスの利用が人によるものかどうかを判定するため、端末やブラウザの情報、および画面上の操作に関する情報を Google に送信します。これらの情報は、この判定と、Google 自身のサービス改善のために利用されます。詳細は Google のプライバシーポリシーおよび利用規約をご確認ください。',
```
kowinauyeung: **🔴 BLOCKER — this discloses two technologies that do not exist on this branch, and a test now requires the disclosure.**

`tests/unit/content.test.ts` asserts `it.each(['reCAPTCHA', 'App Check'])('discloses %s, which is implemented now')`. Verified at `d759eda`:

```console
$ grep -rn "AppCheck\|app-check\|RECAPTCHA" src/ .env.example
$ echo $?
1
```

Nothing. App Check and the reCAPTCHA v3 provider are on `feat/security-hardening`, which is a separate pull request against the same base.

This is the mirror of the rule the test's own docblock argues for, and it is worth quoting because the reasoning is exactly right: *"A privacy policy naming a processor that processes nothing is not cautious, it is untrue — and it is the kind of untrue that survives, because nobody re-reads a policy looking for something that is not there."*

If this merges first, `deploy-dev.yml` publishes a policy telling readers the service sends their device, browser and on-screen interaction data to Google, when it sends none of it. The commit that added the paragraph is titled "now that it is actually there" — true of the working copy where both branches exist, not of this branch.

Two clean resolutions:

- **Merge #21 first**, and say so on both pull requests. There is now a real ordering constraint between them and nothing enforces it; the two "touch no file in common", so neither will conflict to warn you.
- **Move this paragraph and its test into #21**, which makes each pull request true on its own and removes the ordering constraint entirely.

The second is the one I would take — an ordering constraint that lives only in two descriptions is the kind that gets lost on a Friday.

src/lib/diagnostics.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * A vocabulary id is a Firestore document id and belongs to one person's
+ * notebook; pasting it into a support form leaks which word was being read. The
+ * screen is what a report needs.
+ */
+export function routePattern(pathname: string): string {
+  return (
+    pathname
+      // Long opaque ids — Firestore auto-ids and the migration slugs alike.
+      .replace(/\/[A-Za-z0-9_-]{12,}(?=\/|$)/g, '/:id')
```
kowinauyeung: **🟡 The redaction is by segment length, and the promise it backs is absolute.**

`{12,}` catches every id this app actually mints — I checked all four sources rather than assuming: `entryRepo.ts:98` and `wordSetRepo.ts:74` use `addDoc`, `progressRepo.ts:65` uses `doc(collection)`, and `migration/upload.mjs:183` uses `collection.doc()`. All four are Firestore auto-ids, all 20 characters. **So this is not leaking anything today.**

What makes it worth changing anyway is what it is resting on. The guarantee is stated without qualification in two places — `privacy.ts:78` (「登録した単語の内容は含まれません」) and the line under the copy button in `CopyDiagnostics.tsx` — while the mechanism is a heuristic over the length of a vendor-generated identifier. Firestore auto-id length is not part of this codebase's contract with itself, and a future human-readable slug (`/vocabulary/w-choukou`) would defeat it silently, in the one direction where silence is the whole problem.

The routes are enumerable, which makes the total version no harder than the heuristic:

```ts
.replace(/^\/vocabulary\/[^/]+/, '/vocabulary/:id')
.replace(/^\/wordsets\/[^/]+/, '/wordsets/:id')
```

Then the comment can say it covers the routes rather than the lengths, and it stays true whatever an id looks like later.

*Surfaced by a second reviewer on a different model; the id-shape verification above is mine.*

src/components/CopyDiagnostics.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const text = formatDiagnostics(collectDiagnostics({ projectId, pathname, errorId }));
+
+  return (
+    <details className="rounded-panel border border-line p-3 text-left">
+      <summary className="cursor-pointer text-xs text-muted">診断情報を見る・コピーする</summary>
+      <pre className="mt-2 overflow-x-auto text-[11px] leading-relaxed text-muted">{text}</pre>
+      <button
+        type="button"
+        onClick={() => {
+          void navigator.clipboard.writeText(text).then(() => {
```
kowinauyeung: **🟡 No failure path on the clipboard write, on the screen whose only job is filing a report.**

Two ways this does nothing visible:

- `navigator.clipboard` is `undefined` outside a secure context, so the property access throws synchronously inside the handler. React does not route event-handler errors to an error boundary, so it reaches `window.onerror` and the button appears inert.
- `writeText` rejects on a permission or focus failure. With no `.catch`, that is an unhandled rejection and the label never changes.

Either way the user presses コピー, nothing happens, and there is nothing to tell them to select the text by hand instead. The `<pre>` above means they *can* — that is the saving grace, and it is why this is not a blocker — but they would have to work it out.

```ts
onClick={() => {
  navigator.clipboard
    ?.writeText(text)
    .then(() => { setCopied(true); setTimeout(() => setCopied(false), 1800); })
    .catch(() => setFailed(true));
}}
```

…with the failed state saying 上のテキストを選択してコピーしてください, which is the instruction that actually helps.

src/components/CopyDiagnostics.tsx
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * The preview is not decoration. A support form asking for "diagnostic
+ * information" is asking somebody to paste an opaque blob about themselves, and
+ * the only honest version of that lets them read it first. It also makes the
+ * absence of a uid, an email and a word visible rather than promised.
+ */
+export function CopyDiagnostics({ errorId }: { errorId: string }) {
+  const { pathname } = useLocation();
+  const [copied, setCopied] = useState(false);
+  const text = formatDiagnostics(collectDiagnostics({ projectId, pathname, errorId }));
```
kowinauyeung: **🟡 Recomputed every render, including `at`.**

`collectDiagnostics` defaults `now` to `new Date()`, so the timestamp in the preview changes on each render — including the one caused by `setCopied(true)` immediately after the copy. The text in the box and the text on the clipboard can therefore differ by the moment they were made.

`useMemo` on `[projectId, pathname, errorId]` fixes both this and the drift described on `Account.tsx:160`.

src/routes/Account.tsx
```diff
@@ -1,38 +1,175 @@
...
+          </Link>
+          <Link to="/about" className="hover:underline">
+            語彙庭について
+          </Link>
         </div>
-        <div>
-          <dt className="text-xs text-muted">接続先プロジェクト</dt>
-          <dd className="mt-1">{projectId}</dd>
+        <div className="mt-4">
+          <CopyDiagnostics errorId={newErrorId()} />
```
kowinauyeung: **🟡 `newErrorId()` runs during render, so the id changes on every state change.**

This component sets state for `busy`, `note`, `error` and `confirming` — every one of those re-renders and mints a new id. `ErrorScreen.tsx:22` gets the same call right with `useMemo(() => newErrorId(), [])`.

The id exists so a report and a screenshot can be matched to each other, which only works if it survives the interaction that produced it. Here, pressing エクスポート changes it.

```ts
const errorId = useMemo(() => newErrorId(), []);
```

src/routes/Account.tsx
```diff
@@ -1,38 +1,175 @@
...
+  const runDelete = async () => {
+    setBusy('delete');
+    setError(null);
+    setNote(null);
+    try {
+      const removed = await deleteEverything({
+        entries: entries.repository,
+        wordSets: wordSets.repository,
+      });
+      setNote(`${removed.entries} 語と ${removed.wordSets} 個の単語集を削除しました。`);
```
kowinauyeung: **🟡 The report of what was removed is set and then discarded a line later.**

`setNote(...)` then `await signOutUser()` unmounts the page before anything renders it.

`accountData.ts` argues the design deliberately reports rather than claims success — *"which is why it reports what it removed and asks the user to run it again rather than claiming success"* — and the failure path honours that, because it does not sign out. The success path does not.

Given the blocker above, this may resolve itself: once the dialog says only what is actually deleted, the honest post-delete screen probably belongs after the sign-out rather than before it — a line on the signed-out landing page, or a confirmation step before the session ends.

build-info.ts
```diff
@@ -0,0 +1,33 @@
...
+ * moment it was written, and would differ between a laptop regenerating it and
+ * CI checking it. Which is exactly why `.env.e2e` is committed with a fixed
+ * project id: the login screen prints that too.
+ */
+const E2E_COMMIT = 'e2e0000';
+
+export const buildInfo = (mode: string) => ({
+  __APP_VERSION__: JSON.stringify(pkg.version),
+  __COMMIT_SHA__: JSON.stringify(
+    mode === 'e2e' ? E2E_COMMIT : (process.env.GITHUB_SHA ?? 'dev').slice(0, 7),
```
kowinauyeung: **🟡 `GITHUB_SHA` is the wrong commit on the production deploy path.**

For a `workflow_dispatch` run, `GITHUB_SHA` is the commit the *workflow* ran on — the default branch's head — not an input. `deploy-prod.yml` on `feat/security-hardening` takes a `sha` input, checks it out, and builds that. So the footer, the Account page, the error screen and the copied diagnostics would all name a commit that is not the one deployed.

Harmless while the two are the same, which they usually are. Wrong precisely when they are not: a rollback deploys an older commit and the build line claims the newer one, sending whoever reads a bug report to source that was never running.

The fix belongs in the workflow rather than here — `GITHUB_SHA: ${{ inputs.sha }}` on the build step, or a dedicated variable this file prefers. Raising it here because this is the file that decides what the value means, and because the two pull requests will not conflict to point it out.

src/router.tsx
```diff
@@ -21,4 +31,16 @@ export const router = createBrowserRouter([
     ],
   },
   { path: '/login', lazy: () => import('./routes/Login') },
+  { path: '/about', lazy: () => import('./routes/public/About'), errorElement: <ErrorScreen /> },
```
kowinauyeung, line 36: 🟡 The four documents each get an `errorElement` and `/login` (line 33) does not.

The comment above justifies the boundary with "a stale chunk after a deploy is the common one" — and `/login` is the route where that matters most, since it is the only way back into the app. A visitor who hits it on a stale chunk gets React Router's default error page.

firebase.json
```diff
@@ -11,6 +11,31 @@
...
+          {
+            "key": "Permissions-Policy",
+            "value": "camera=(), microphone=(), geolocation=(), payment=(), usb=(), interest-cohort=()"
+          },
+          {
+            "key": "Cross-Origin-Opener-Policy",
+            "value": "same-origin-allow-popups"
+          },
+          {
+            "key": "Content-Security-Policy-Report-Only",
```
kowinauyeung, line 35: **🟡 Report-Only means `frame-ancestors 'none'` enforces nothing, and it is the reason `X-Frame-Options` was left out.**

`docs/security-headers.md:29` already names this exactly: *"`frame-ancestors 'none'` is the reason there is no `X-Frame-Options`: it is the same protection, and CSP supersedes it — but while the policy is Report-Only it enforces nothing."*

Identifying the gap is not the same as closing it. Until the policy is promoted, the app is framable by anyone, with nothing in the way. `X-Frame-Options: DENY` is one line, takes effect immediately, needs none of the sign-in verification the CSP promotion depends on, and is harmless to leave in place afterwards.

The rest of the Report-Only reasoning I would not touch — a CSP that breaks sign-in after deploy, for users who then cannot get in to report it, is a genuinely nasty failure and Report-Only is the right call. This is just the one protection that does not have to wait for it.

*Surfaced by a second reviewer on a different model; the `security-headers.md` cross-reference is mine.*

src/content/site.ts
```diff
@@ -0,0 +1,46 @@
...
+ * handed to somebody probing the service rather than a courtesy to a user. A
+ * reader who wants the source can find it; a reader who does not want it should
+ * not be given a map of the backend on the error screen.
+ *
+ * If it is ever linked again, link it from a page a signed-in user reaches, not
+ * from the four public documents and not from the crash screen.
+ */
+
+/** `準備中` wherever a contact detail has not been filled in yet. */
+export const orPending = (value: string): string => value || '準備中';
```
kowinauyeung: 🟡 `orPending` is exported and tested, and nothing calls it.

```console
$ grep -rn "orPending" src tests
src/content/site.ts:46
tests/unit/content.test.ts:5,119,120
```

`CLAUDE.md` asks that a new test correspond to a real behaviour, a backlog item or a defect. Either wire it in where a contact detail can be blank — `site.feedbackFormUrl` is the obvious candidate, and the test reads as though that was the intent — or drop the helper and its test together.

src/components/ErrorScreen.tsx
```diff
@@ -0,0 +1,71 @@
...
+   * boundary* — a top-level `path: '*'` never gets a turn. Measured: before this
+   * branch existed, /no-such-page rendered 問題が発生しました with an error id
+   * and a link to the report form, which invites a bug report for a typo.
+   *
+   * Rendered rather than redirected, so the address the user typed stays in the
+   * bar where they can see what went wrong.
+   */
+  if (isRouteErrorResponse(error) && error.status === 404) return <NotFound />;
+
+  console.error(`[${errorId}]`, error);
```
kowinauyeung: 🔵 `console.error` in the render body runs twice per error under `StrictMode`, which `main.tsx:9` enables, and again on any re-render.

`useEffect(() => { console.error(...) }, [errorId, error])` logs once per error instead.

*Surfaced by a second reviewer on a different model; the StrictMode confirmation is mine.*

firebase.json
```diff
@@ -11,6 +11,31 @@
...
+        "headers": [
+          {
+            "key": "Strict-Transport-Security",
+            "value": "max-age=31536000; includeSubDomains"
+          },
+          { "key": "X-Content-Type-Options", "value": "nosniff" },
+          { "key": "Referrer-Policy", "value": "strict-origin-when-cross-origin" },
+          {
+            "key": "Permissions-Policy",
+            "value": "camera=(), microphone=(), geolocation=(), payment=(), usb=(), interest-cohort=()"
```
kowinauyeung: 🔵 `interest-cohort=()` opts out of FLoC, which no longer exists — it was replaced by the Topics API, whose opt-out is `browsing-topics=()`. Harmless as it stands; just not doing anything.

src/components/AppLayout.tsx
```diff
@@ -8,8 +8,10 @@ import { useTheme } from '@/lib/theme';
 import { useClickOutside } from '@/lib/useClickOutside';
 import { VocabDialogProvider } from '@/lib/vocabDialog';
 import { WordSetsProvider } from '@/lib/wordSets';
+import { Component as Home } from '@/routes/Home';
```
kowinauyeung: 🔵 `Home` is imported eagerly while every route below is lazy, so the landing page and `content/about.ts` land in the signed-in bundle for users who will never see them. `PublicFooter` has to be eager, `Home` does not.

kowinauyeung:
## Review summary

**Verdict: `REQUEST CHANGES`** — two blockers. Fourteen inline comments above.

Neither blocker is a bug in the ordinary sense. Both are places where this branch says something is true and it is not yet, and both come from splitting the production-readiness work across two pull requests — which was the right split for reviewing, and has left a seam.

1. **`deleteEverything` deletes words and word sets. The button, the confirmation dialog and the privacy policy all promise more** — practice records, and the account itself. `ProgressRepository` has no delete method, so this is a port change rather than a missing call, and the honest end state today is that a user who deletes their account can sign back in and find their practice history. Detail on `src/lib/accountData.ts:115`.
2. **The privacy policy discloses reCAPTCHA v3 and Firebase App Check, and a new test enforces the disclosure** — neither exists on this branch; both are in #21. This is the rule that test was written for, pointed the other way. It also means #21 now has to merge first, and nothing enforces that: the two branches share no file, so neither will conflict to warn you. Moving the paragraph and its test into #21 removes the constraint instead of documenting it. Detail on `src/content/privacy.ts:54`.

Everything else is a suggestion. The ones I would not skip: the clipboard call has no failure path on the screen whose only purpose is filing a report; `newErrorId()` runs during render on the Account page, so the id changes under the user; and `X-Frame-Options` was consciously left out in favour of a `frame-ancestors` directive that a Report-Only policy does not enforce — `docs/security-headers.md` names that gap without closing it, and one line closes it now rather than at promotion.

### On the screenshot baselines

Both were regenerated and the description does not mention it, so I looked at them rather than taking the diff on trust.

`login-linux.png` (1280×720, unchanged) now carries the footer: the four policy links, and `v0.1.0 · e2e0000 · Local` on the right, with the card centred in the space above rather than in the full viewport. That matches the column restructure in `Login.tsx` exactly, and it confirms the pinned e2e commit renders as intended.

`dashboard-linux.png` grew from 1135 to 1192 pixels. The extra 57 is the same footer, and its links begin at the same left edge as the card grid above — which is the alignment the new assertion in `vocabulary.spec.ts` measures.

Both deltas are fully accounted for by the diff and neither carries anything else. What is missing is one sentence saying they came out of `yarn test:visual:update`; that is the difference between a valid baseline and a macOS-authored one that fails every CI run, and it is the sort of thing worth having in the record.

### Verification

Reviewed at `d759eda` against `d406a07`, in a detached worktree.

`yarn test:unit` passes — 546 tests across 31 files. `tsc -b --noEmit` and `tsc -p tsconfig.test.json --noEmit` both clean; `eslint .` reports 0 errors (17 warnings, all pre-existing in `Practice.tsx`, untouched here). Claims about identifier shapes in the diagnostics comment were checked against all four id sources rather than assumed, and the `flashcards`/`flashcard` distinction in `routePattern` was checked against the route table before being believed — it is correct, the plural is the URL and the singular is the domain value.

Cross-checked with a second reviewer on a different model. It reached the same two blockers independently, and contributed three findings I had missed: the length-based route redaction, the unenforced `frame-ancestors`, and `console.error` in the render body. Each was verified here before being included — the redaction one is downgraded from its original severity as a result, because every id this app mints is a 20-character auto-id and nothing is leaking today.

### One process note

1,687 lines across 36 files. This one would genuinely have been four: public pages, release identity, error handling, account data. They share almost nothing — the release identity reaches the others only through the footer — and the account-data half is the one carrying both blockers and the only part that needed careful reading twice.

kowinauyeung:
### ⚠️ This comment was wrong when it was posted

Superseded. It said there were no review comments; there were fourteen, and a
`REQUEST CHANGES` review, submitted at 10:06 — about forty minutes before this.

The cause is not that the review was hard to find. **I fetched the comments
once, well before the review landed, and then treated that empty result as the
current state when I posted.** The same mistake closed #19 with four comments
unanswered until they were pointed out, and it is written into the handoff as a
lesson. Re-fetching immediately before replying is the fix, and it is now the
rule I hold myself to.

The irony is that this comment exists at all because "I found nothing" and "I
did not look" are indistinguishable from the outside. It turned out to be a
third thing: I looked, at the wrong time, and reported a stale answer with the
confidence of a fresh one.

Working through the review now — the two blockers first.

src/lib/accountData.ts
kowinauyeung: @kowinauyeung
Upheld. Took option 1 — the promise is already made in a privacy policy, and narrowing the wording would have been writing down that the feature does less than the policy says.

Both ports grew a method, since neither existed: `ProgressRepository.removeAll()` and `AuthPort.deleteAccount()`. Sessions delete in batches of 400 because a write batch caps at 500 and a year of daily practice is past that. The account goes last, and that ordering has its own test — it is the one step that cannot be retried, because afterwards there is no session left to delete anything with.

`auth/requires-recent-login` now says what to do instead of reading as a generic failure. And you were right that the success note resolves itself: it is gone, because the page it would render on has just been signed out from.

The sentence I will keep from this is the one about the end state — signed out, sign back in, history intact. That is the description that makes it obviously unshippable, and it is not visible from the diff.

Fixed in bbda126

src/content/privacy.ts
kowinauyeung: @kowinauyeung
Upheld, and reproduced: `grep -rn "AppCheck\|app-check\|RECAPTCHA" src/ .env.example` returns nothing on this branch.

**Your preferred fix is not available, and I checked before choosing.** `src/content/privacy.ts` does not exist on `develop`, so it does not exist on #21 either — there is no file there to move the paragraph into.

So instead of documenting the constraint or reversing it, the test now reads `client.ts` and requires the disclosure **if and only if** the client initialises App Check. The constraint becomes mechanical: this pull request goes red until the two arrive together, which is what you wanted and does not depend on anyone remembering an ordering on a Friday.

Reading a source file from a test is not the usual thing. It earns its place here because the claim genuinely is "the policy matches the implementation", and every cheaper version of that asserts one side and trusts the other — which is how the paragraph got added on the wrong branch in the first place. Verified by putting the paragraph back with App Check still absent: red.

Fixed in bbda126

src/lib/diagnostics.ts
kowinauyeung: @kowinauyeung
Upheld, and thank you for checking all four id sources rather than asserting the risk — that is what makes the change worth doing on a path that leaks nothing today.

The part I had wrong is what the rule was resting on: Firestore auto-id length is not this codebase's contract with itself, and the guarantee it backs is stated without qualification in two places. Now matched by route, so it holds whatever an id looks like later. There is a case for `/vocabulary/w-choukou` and a one-character id, both of which the length rule waved through.

Fixed in bbda126

src/components/CopyDiagnostics.tsx
kowinauyeung: @kowinauyeung
Upheld — and the first of the two failure modes is the one I would not have found, because a synchronous throw in an event handler does not reach an error boundary and so leaves no trace at all.

Both handled, with the failed state carrying the instruction that actually helps. You are right that the `<pre>` is the saving grace; it is only a saving grace if somebody says so.

Fixed in bbda126

src/components/CopyDiagnostics.tsx
kowinauyeung: @kowinauyeung
Upheld. Memoised on `[pathname, errorId]`.

The detail worth recording is the one you spotted: the render that re-read the clock was the one caused by reporting a *successful* copy — so the box and the clipboard could disagree precisely at the moment the user believed they matched.

Fixed in bbda126

src/routes/Account.tsx
kowinauyeung: @kowinauyeung
Upheld, and the comparison is fair: `ErrorScreen` got this right two files away, so it is not that the pattern was unknown here.

An id that survives nothing is worse than no id — a user reads one off the screen, pastes text containing another, and the two cannot be joined. `useMemo(() => newErrorId(), [])`.

Fixed in bbda126

src/routes/Account.tsx
kowinauyeung: @kowinauyeung
Upheld, and it resolved the way you predicted. With the account actually deleted there is nobody left to read a success note, so it is gone rather than moved: it was set and discarded by the sign-out on the next line, which is a message that rendered for one frame at most.

The failure path keeps reporting, because that is the case where somebody is still there and the data may be half gone.

Fixed in bbda126

build-info.ts
kowinauyeung: @kowinauyeung
Upheld, and this is the one that would have been hardest to notice in production: correct on every deploy where the two agree, wrong on exactly the rollback where somebody is reading the footer to work out what is running.

`build-info.ts` prefers `DEPLOY_SHA` now, and the workflow passes `inputs.sha` into it — the second half is on #21, since that is where the workflow lives. You were right that neither pull request would have conflicted to point it out.

Fixed in bbda126 here, and 90f8b02 on #21

src/router.tsx
kowinauyeung: @kowinauyeung
Upheld. `/login` is the route where a stale chunk matters most, since it is the only way back in, and it was the one route without the boundary the comment argues for.

Fixed in bbda126

firebase.json
kowinauyeung: @kowinauyeung
Upheld, and the distinction is the point: naming the gap in `security-headers.md` and closing it are different acts, and I had done the first while writing as though it were the second.

`X-Frame-Options: DENY` added. It takes effect immediately, needs none of the sign-in verification CSP promotion depends on, and is harmless afterwards. The Report-Only reasoning stands untouched.

Fixed in bbda126

src/content/site.ts
kowinauyeung: @kowinauyeung
Upheld — deleted, along with its test.

Wiring it in was the alternative and would have been worse: with the form URL now required and asserted non-empty, a fallback to 準備中 is a branch nothing can reach, and a test covering it would have been the second unreachable thing rather than a fix for the first.

Fixed in bbda126

src/components/ErrorScreen.tsx
kowinauyeung: @kowinauyeung
Upheld. Moved into an effect.

It also had to move above the 404 early return, which is its own small lesson: a hook behind a conditional return is not only against the rule, it would have skipped the log for exactly the errors that reached the 404 branch.

Fixed in bbda126

firebase.json
kowinauyeung: @kowinauyeung
Upheld. `browsing-topics=()`.

An opt-out for a thing that no longer exists is worse than harmless, because it reads as coverage — somebody auditing the header sees a tracking opt-out and stops looking.

Fixed in bbda126

src/components/AppLayout.tsx
kowinauyeung: @kowinauyeung
Upheld. `Home` is lazy now, like every route around it, with `Suspense` for the one place it renders. `PublicFooter` stays eager, as you noted it must.

Fixed in bbda126

build-info.ts
kowinauyeung: Correction to my own reply above: the commit on #21 is `f416070`, not `90f8b02`. I quoted a SHA before pushing rather than after — which is precisely the step this loop has a rule about.

src/lib/accountData.ts
kowinauyeung: @kowinauyeung
Verified at `bbda126`. All four collections and the account, with both ports grown rather than a call bolted on — `ProgressRepository.removeAll()` and `AuthPort.deleteAccount()`, implemented in the Firestore adapter, the auth adapter and the in-memory backend so the end-to-end suite exercises the same shape.

Checked the parts that could have been subtly wrong rather than just the shape:

- `removeAll()` re-queries each pass instead of paging a cursor, which is right — the previous batch removed what it matched, so the first page is always the next unprocessed one. It terminates on `empty` and on a short page, and 400 is under the 500-write batch cap.
- Ordering: data, then the account. The reason in the comment is the real one — after `deleteUser` there is no session left to delete anything with.
- `auth/requires-recent-login` is caught by code and given an instruction rather than the generic message. That is the failure a real user will actually hit.

One residual, not worth reopening for: `users/{uid}` itself is still never deleted, while the docblock two lines above says "the rows first, **the parent last**". It is a no-op today because nothing creates that document — but #21 just changed the rules to permit deleting it, specifically because this was the next caller. Either the `deleteDoc` or a word about why the parent is excluded would close the gap between the comment and the code.

Resolving.

src/content/privacy.ts
kowinauyeung: @kowinauyeung
Verified at `bbda126`, and verified as a mechanism rather than as a diff — this is the fix I would not have thought of, so I tried to break it.

The paragraph is gone and the test now reads `client.ts` and asserts `disclosed === implemented`. Simulating #21 landing, by adding a reference to `initializeAppCheck` to `client.ts` and re-running:

```
FAIL  what the public pages must not disclose > names App Check when, and only when, the client initialises it
AssertionError: expected false to be true
```

So the constraint really is mechanical: whichever branch adds App Check carries the disclosure or CI goes red. That is strictly better than the fix I suggested, which only relocated the problem into two descriptions.

Two limits worth knowing, neither of them objections:

- It matches the literal `initializeAppCheck`. Changing how the client imports or wraps that call makes the guard read `false` and go quiet — the usual cost of grepping source, and acceptable given what it buys.
- It fires on the merge result rather than on either pull request in isolation, unless the second one is brought up to date with `develop` first. Worth knowing which of the two protections you are relying on.

Reading a source file from a test is unusual, and your justification for it is the correct one: the claim under test *is* "the policy matches the implementation", and every cheaper assertion trusts one side. Resolving.

src/lib/diagnostics.ts
kowinauyeung: @kowinauyeung
Verified at `bbda126`, in both directions. Route-based now, with `tests/unit/diagnostics.test.ts` covering `/vocabulary/w-choukou` and `/vocabulary/x` — the cases a length rule let through.

Reverting to the `{12,}` version puts two tests red and nothing else:

```
FAIL  routePattern > names the screen, not the document
FAIL  routePattern > redacts a short or readable id, which a length rule would have let through
```

The `$` anchors inside the practice lookahead are a real improvement over what I suggested — `(?!flashcards$|dictation$)` keeps the two modes without also keeping anything that merely starts with them.

One thing traded in the swap, worth a thought rather than a change: the old rule was total over *paths* and wrong about *ids*; the new one is total over ids and enumerates paths. Today the enumeration is complete — `/vocabulary/:id` and `/wordsets/:id` are the only id-bearing routes — but a route added later carries its id into the diagnostics until somebody remembers this file. Keeping the length rule as a final backstop after the two route rules would cover both, at the cost of one line. Your call; the promise it backs is the unqualified kind.

Resolving.

src/components/CopyDiagnostics.tsx
kowinauyeung: @kowinauyeung
**Half of this is fixed and the comment says both halves are — leaving it open.**

The rejection path works. Verified:

```console
$ node -e "const nav={clipboard:{writeText:()=>Promise.reject(new Error('denied'))}};
  nav.clipboard?.writeText('x').then(()=>console.log('copied')).catch(()=>console.log('failed'))"
failed
```

The undefined-clipboard path does not, and optional chaining is why. `?.` short-circuits **the entire remaining chain**, not just the one access — so when `navigator.clipboard` is undefined, `.then()` and `.catch()` are never evaluated at all:

```console
$ node -e "const navigator={};
  console.log(String(navigator.clipboard?.writeText('x').then(()=>0).catch(()=>0)))"
undefined
```

No throw, but no `setState('failed')` either. The button keeps its original label, the failure message never renders, and the user gets exactly what the comment describes as the original defect: *"the button would simply appear inert."* The change removed the console error and kept the silence.

Splitting the chain handles both:

```ts
onClick={() => {
  const write = navigator.clipboard?.writeText(text);
  if (!write) return setState('failed');
  write
    .then(() => { setState('copied'); setTimeout(() => setState('idle'), 1800); })
    .catch(() => setState('failed'));
}}
```

Still not a blocker — the `<pre>` is selectable and the new message says so, once it renders. But the comment above the button currently asserts a property the code does not have, which is the thing this branch has been careful about everywhere else.

src/components/CopyDiagnostics.tsx
kowinauyeung: @kowinauyeung
Verified. `useMemo` on `[pathname, errorId]`, so `at` is fixed at mount and the preview cannot drift from what was copied. The comment names the specific render that caused it — the one reporting a successful copy — which is the detail that makes it stay fixed. Resolving.

src/routes/Account.tsx
kowinauyeung: @kowinauyeung
Verified. `useMemo(() => newErrorId(), [])`, matching `ErrorScreen`. The comment counts the four setStates that were re-minting it, which is the reason a reader needs. Resolving.

src/routes/Account.tsx
kowinauyeung: @kowinauyeung
Verified — removed rather than relocated, which is the right answer now that the account is deleted too: there is genuinely nobody left to read it. The comment says so where the next person would otherwise re-add it.

The error path still reports honestly, and now distinguishes `requires-recent-login`, so the "some of it may be gone" reasoning survives where it still applies. Resolving.

build-info.ts
kowinauyeung: @kowinauyeung
Verified, both halves, at `bbda126` here and `f416070` on #21:

- `build-info.ts` prefers `DEPLOY_SHA` and falls back to `GITHUB_SHA`.
- `deploy-prod.yml:82` sets `DEPLOY_SHA: ${{ inputs.sha }}` on the Build step.

I re-fetched #21 before checking rather than trusting the SHA in the reply above — which found that the head had moved again since I last looked. Your correction had the right commit.

`||` rather than `??` is right here: an empty-string `DEPLOY_SHA` from an unset input should fall through to `GITHUB_SHA`, and `??` would have taken the empty string. Resolving.

src/router.tsx
kowinauyeung: @kowinauyeung
Verified — `/login` has the boundary, with the reason inline. Resolving.

firebase.json
kowinauyeung: @kowinauyeung
Verified. `X-Frame-Options: DENY` alongside the Report-Only policy, so the interim is covered and promotion makes it redundant rather than necessary. Resolving.

src/content/site.ts
kowinauyeung: @kowinauyeung
Verified — deleted, along with its test, rather than wired up to justify itself. Resolving.

src/components/ErrorScreen.tsx
kowinauyeung: @kowinauyeung
Verified, and placed correctly — above the 404 early return, so the hook is unconditional. The note about a 404 otherwise leaving the previous error's log unrepeated is a consequence I had not worked through; it is right, and it is a better reason than the one I gave. Resolving.

firebase.json
kowinauyeung: @kowinauyeung
Verified — `browsing-topics=()` replaces the FLoC opt-out. Resolving.

src/components/AppLayout.tsx
kowinauyeung: @kowinauyeung
Verified — `lazy` plus a `Suspense` boundary, so the landing page leaves the signed-in bundle.

`fallback={null}` means a signed-out visitor at `/` gets one blank frame while a small same-origin chunk loads. That is the page Google's OAuth review looks at, so it is worth knowing you have chosen it deliberately; the e2e homepage test passes because Playwright waits. Not asking for a change. Resolving.

kowinauyeung:
## Re-review at `bbda126`

**Both blockers closed. Thirteen threads verified and resolved, one left open** — the clipboard fix handles one of the two failure modes its own comment says it handles.

### Suites, at head

| Suite | Result |
| --- | --- |
| `yarn test:unit` | 555 passed, 32 files (was 546 / 31) |
| `yarn test:e2e` | 73 passed, 5 skipped — including both screenshot baselines, unchanged and green |
| `tsc -b` and `tsc -p tsconfig.test.json` | clean |
| `eslint .` | 0 errors |

### Verified by breaking it

Claims about behaviour were checked in both directions rather than read off the diff.

**The App Check guard actually fires.** This is the fix I would not have thought of, so I tried to defeat it: adding a reference to `initializeAppCheck` in `client.ts` — simulating #21 landing — turns the content test red with `expected false to be true`. The ordering constraint between these two pull requests is now mechanical rather than written in two descriptions, which is strictly better than the fix I proposed. Two limits are noted on the thread: it matches the literal `initializeAppCheck`, and it fires on the merge result rather than on either branch alone unless the second is brought up to date with `develop` first.

**The route redaction.** Reverting to the `{12,}` rule puts exactly two tests red, both in the new `diagnostics.test.ts`, including the `/vocabulary/w-choukou` case. The `$` anchors inside the practice lookahead are better than what I suggested.

**Account deletion.** Read the parts that could have been subtly wrong rather than the shape: `removeAll()` re-queries per pass instead of paging a cursor, which is correct because each batch removes what it matched; it terminates on both `empty` and a short page; 400 is under the 500-write cap; the account goes last for the reason stated, since after `deleteUser` there is no session left to delete with. Both ports grew a method and all three backends implement it, so the end-to-end suite exercises the same shape the browser will.

`build-info.ts` was checked against **both** halves — `DEPLOY_SHA` here, and `deploy-prod.yml:82` on #21 at `f416070`. I re-fetched #21 before checking rather than trusting the SHA quoted in the thread, which is how I found the head had moved again; the correction posted there had the right commit.

### Still open — one thread

`CopyDiagnostics.tsx` — the rejection path is handled, the undefined-`navigator.clipboard` path is not, because `?.` short-circuits the *whole* chain: when `clipboard` is undefined, neither `.then()` nor `.catch()` is evaluated, so `setState('failed')` never runs and the button keeps its original label. Verified both ways in node; detail and a two-line fix on the thread.

Not a blocker — the `<pre>` is selectable and the new message says so, once something renders it. Worth closing because the comment above the button asserts that both modes are handled, and this branch has been unusually careful about comments that outrun the code.

### Two residuals, mentioned on their threads and not reopened

- `users/{uid}` itself is still never deleted, while the docblock says "the rows first, **the parent last**". A no-op today, since nothing creates that document — but #21 changed the rules to permit deleting it *because* this was the next caller. Either the call or a word about why the parent is excluded.
- `routePattern` swapped a rule that was total over paths and wrong about ids for one that is total over ids and enumerates paths. The enumeration is complete today. Keeping the length rule as a backstop after the two route rules would cover both, for one line.

### On the process

Worth recording, since the correction above it is in the record too: fourteen comments, two of them blockers, every one upheld with the reasoning checked rather than accepted — and three of the fixes are better than what I asked for. The App Check guard replaced a documented constraint with an enforced one; the `$` anchors closed a case I had not considered; and the `console.error` placement came with a consequence I had missed, that a 404 would otherwise leave the previous error's log unrepeated.

The remaining open thread is the one place where a comment claims more than the code does, which is the same shape as both blockers. That pattern is worth watching for on its own.

src/components/CopyDiagnostics.tsx
kowinauyeung: @kowinauyeung
Upheld, and you were right to leave it open — the change removed the throw and kept the silence, which is the defect the comment above it describes.

Reproduced independently before touching anything:

```console
$ node -e "const navigator = {};
  console.log(String(navigator.clipboard?.writeText('x').then(()=>0).catch(()=>0)))"
undefined
```

Neither handler is evaluated. The button keeps its label, the message never renders, and the `<pre>` is the only thing left — which the paragraph beside it calls a saving grace only if somebody says so, and nothing said so.

**The comment was also wrong about the mechanism**, which is worth recording separately because it is what made the fix look complete. It said "the property access throws synchronously". Reading an undefined property does not throw; calling it does. So the sentence describing the original defect misdescribed it, and the `?.` that replaced it looked like it addressed what the sentence said.

Taken as you wrote it, with one change: `written === undefined` rather than `!written`. `no-misused-promises` rejects the truthiness test, and it is right to — a promise is always truthy, so reading one as a boolean is a question nobody means to ask. Noted inline so the next person does not shorten it back.

`tests/component/CopyDiagnostics.test.tsx` covers all three: no clipboard, a rejecting `writeText`, and success.

Red before green, and specifically red on the right half: restoring the chained form fails **exactly** the no-clipboard case and leaves the other two green. A test that also went red on the rejection path would have been evidence about the wrong thing, since that path already worked.

560 unit and component, 73 emulator, 73 end-to-end.

Fixed in `e756d20`

kowinauyeung:
## `e756d20` — the clipboard blocker, and `develop` merged in

**The blocker was real and my previous fix addressed the wrong half.** Optional chaining short-circuits the entire remaining chain, so with `navigator.clipboard` undefined neither `.then` nor `.catch` was ever evaluated: the throw was removed and the silence kept. The comment beside it was wrong about the mechanism too — reading an undefined property does not throw, calling it does — which is a large part of why the fix looked complete. Both corrected, with three component cases and a red-first check that fails exactly the no-clipboard case.

**`develop` is merged in, cleanly, and it made one thing due.**

`tests/unit/content.test.ts` asserts that the privacy policy names App Check **if and only if** `client.ts` initialises it. That shape was chosen in the last round because the reviewer's preferred fix — moving the paragraph to #21 — was checked and turned out to be impossible: `src/content/` is created by this pull request, so there was nowhere on #21 to move it to.

After the merge, `client.ts` really does initialise App Check, and the test went red on its own. Nobody had to remember. The ordering constraint between two pull requests stopped being a sentence in a description and became a failing check at exactly the moment it started applying, which is what it was written for. The two sentences about reCAPTCHA v3 and App Check are restored.

**Gate:** 560 unit and component (up from 555), 73 emulator, 73 end-to-end, lint clean, three typecheck passes.

src/components/CopyDiagnostics.tsx
kowinauyeung: @kowinauyeung
Verified at `e756d20`, in both directions, and the fix is right.

Split call, and `written === undefined` rather than the truthiness test I wrote — you were right to change it, and I checked rather than taking it:

```console
$ eslint src/components/CopyDiagnostics.tsx      # with `if (!written)`
  55:16  error  Expected non-Promise value in a boolean conditional  @typescript-eslint/no-misused-promises
```

My suggestion would not have passed lint. Noting the reason inline is the right call, since `!written` is what anybody would reach for when shortening it.

**Red on the right half, confirmed independently.** Restoring the chained form and running the new component test:

```
FAIL  CopyDiagnostics — the failure paths > reports failure when there is no clipboard at all, rather than going quiet
Tests  1 failed | 2 passed (3)
```

One case, and it is the one that was broken. A test that had also gone red on the rejection path would have been evidence about the half that already worked — worth stating in the commit message, as you did.

**The correction about the mechanism is the part I would keep.** You are right that reading an undefined property does not throw and calling it does, and right that this is why the `?.` looked sufficient: the sentence describing the defect described the wrong operation, so a fix aimed at that sentence addressed something that was never happening. The comment now names the short-circuit explicitly, which is the thing the next person will otherwise re-introduce — `?.` really is the obvious-looking way to write this.

The secure-context note added to the commit message is worth having too: a plain-HTTP origin is exactly the state somebody is in while filing a report about the site not working.

Resolving — and that closes every thread on this pull request.

kowinauyeung:
## Re-review at `e756d20`

**All fourteen threads resolved. No blockers, nothing open.**

Three commits since the last pass: the clipboard fix, the merge of `develop` now that #21 has landed, and the reCAPTCHA disclosure coming back.

### Suites, at head

| Suite | Result |
| --- | --- |
| `yarn test:unit` | 560 passed, 34 files |
| `yarn test:emulator` | 73 passed — against the rules #21 merged, which this branch now runs under |
| `tsc -b` | clean |
| `eslint .` | 0 errors |

### The guard fired for real

The headline of this round is not a fix, it is that a mechanism written last round did its job without anybody watching.

`tests/unit/content.test.ts` was made a biconditional — the policy must name App Check if and only if `client.ts` initialises it — precisely because the two halves lived on branches that share no file and would never conflict. Merging `develop` brought App Check in while the disclosure was still removed. Checked out the merge commit `1fcaedb` on its own and ran it:

```
initializeAppCheck in client.ts: 2
reCAPTCHA in privacy.ts:         0

FAIL  what the public pages must not disclose > names App Check when, and only when, the client initialises it
AssertionError: expected false to be true
```

So the ordering constraint between two pull requests stopped being a line in two descriptions and became a red check at the exact moment it started applying. That is the outcome the design was aiming at, and it is now demonstrated rather than argued — which is a stronger position than either pull request was in a few hours ago.

`e756d20` restores the two sentences and the suite is green again. The second sentence — that the reCAPTCHA score is not joined to vocabulary or practice records — answers the question a reader of a study app actually has, and is the kind of thing a policy usually omits.

### The clipboard fix

Verified in both directions.

`written === undefined` rather than the truthiness test I suggested, and the correction is right — my version would not have passed lint:

```console
$ eslint src/components/CopyDiagnostics.tsx      # with `if (!written)`
  55:16  error  Expected non-Promise value in a boolean conditional  @typescript-eslint/no-misused-promises
```

Restoring the chained form fails exactly one of the three new component cases, the no-clipboard one — the half that was broken, not the half that already worked.

The part worth keeping is the correction to the comment: reading an undefined property does not throw, calling it does. The original sentence described the wrong operation, which is *why* `?.` looked like it addressed the defect. The comment now names the short-circuit outright, because `?.` is the obvious-looking way to write this and is the bug.

### Where this leaves it

Two blockers found on the first pass, both closed. Fourteen threads, all resolved, all verified against the code rather than the commit messages. Three of the fixes came back better than what was asked for — the App Check biconditional, the `$` anchors in the route pattern, and the `no-misused-promises` correction above.

`APPROVE` from my side, with the standing caveats that neither legal document has been through a lawyer and that this is still one pull request doing four jobs.

PR #21, opened 2026-08-13
feat(security): bound writes, gate on a claim, and add App Check

kowinauyeung:
### ⚠️ This changes the deploy procedure. Read the order before deploying rules.

Setting the claim **before** deploying the rules is a seamless change. Deploying
the rules first locks everybody out, including the operator, for up to an hour.

```
1. yarn allow you@example.com        # old rules ignore the claim entirely
2. sign out, sign in                 # the new token carries it
3. yarn rules:dev / rules:prod       # only now
```

`yarn allow` needs the account to have signed in at least once — the uid has to
be the one Google issued.

### Summary

The security half of the external production review. The product half is
`feat/production-readiness`; the two touch no file in common.

Branch protection on `main` is already applied and is not in this diff:
`verify`, `emulator` and `e2e` are now all required, with `strict`, linear
history and conversation resolution. Approvals stay at 0 deliberately — GitHub
does not let an author approve their own pull request, so requiring one would
lock a solo maintainer out of their own repository.

### What changed

**The gate is a custom claim, not a document read.** `exists(/allowedUsers/$(uid))`
is billed as a read on every request to every path — on a public service that
doubles the read cost of the application to answer a question whose answer
changes twice in an account's lifetime.

**Writes are bounded.** Sizes and list lengths on entries, word sets and
sessions; `ownerUid` must match the path; `createdAt` cannot be rewritten.

**App Check**, off until a site key is configured.

**The recursive wildcard had to go, and that is the whole change.** Rules are a
union: `match /{document=**}` granting write is an unconditional allow, so every
validation added beneath it would have been evaluated *beside* it rather than
*after*. Hardening while leaving it in place would have looked like hardening
and enforced nothing. Verified by putting it back — seven of the new cases go
green that should be red. The cost is that an unlisted collection is now denied
rather than allowed, which is the right direction and is itself a test.

**Revocation became two operations.** A claim lives in the ID token, so clearing
it changes nothing until that token expires. `admin/allow-user.ts` also revokes
the refresh tokens, and doing only the first step leaves up to an hour of access
after a ban. The test says so where somebody will read it.

**Bounds are sizes, not a schema.** Requiring every field to exist turns each
schema change into a rules deploy, and a rules deploy lagging a client release
denies writes from a version already shipped. **What rules cannot check is
recorded in the file**: there is no way to iterate a list, so the length of
`senses[3].description` is invisible there.

**App Check is off unless a key is set.** A build enforcing it with no key
attaches no token and every request is refused — a missing environment variable
becoming an outage with nothing pointing at the cause. Enforcement stays off in
the console until its metrics show real traffic attesting.

**reCAPTCHA v3, not Enterprise.** Enterprise was taken from the review without
checking what it needs: it is a Cloud product whose API cannot be enabled
without a billing account, so on Spark the key could never have been created.

### A gap this found in the existing suite

`tests/integration` connects with `mockUserToken: 'owner'` — the emulator's
owner credential, which **bypasses rules entirely**. A green adapter suite has
therefore never said anything about whether the documents it produces are
writable in production. That did not matter while the only rule was ownership;
it matters the moment a well-formed write can be rejected. Three cases now build
the real payloads from the production factories and put them through the rules.

### Impact, measured

- **Existing data**: nothing is near a cap. Across the 67 migrated entries the
  worst case is 25 keys against a limit of 40; `definition` peaks at 218
  characters against 20,000. Rules gate writes and not reads, so an
  over-sized document would still be readable but no longer editable — there
  are none.
- **The adapters**: `entryRepo.create`, `wordSetRepo.create` and both documents
  `recordSession` writes in one batch are accepted.
- **`allowedUsers`**: nothing reads it any more. It is dead data, not deleted.

### Testing

516 unit, 64 rules and adapter, 62 end-to-end. Two breaks verified: restoring
the recursive wildcard, and dropping the claim check.

### Not in this PR

- **Opening signups.** The decision is taken — production allows any verified
  account, development keeps an allowlist — and the mechanism is two rules files
  with a test asserting they differ only in `isAllowed()`. It is sequenced after
  App Check has a key and its metrics look right, because opening registration
  before then is the window with neither a gate nor attestation.
- **Production deploy lockdown** and **`EntriesProvider` loading the whole
  notebook**, which is the largest remaining quota risk and is not mentioned in
  the review at all.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,9 +12,15 @@ VITE_FIREBASE_STORAGE_BUCKET=
 VITE_FIREBASE_MESSAGING_SENDER_ID=
 VITE_FIREBASE_APP_ID=
 
-# reCAPTCHA Enterprise site key for App Check. Optional: with it unset the app
-# runs without App Check rather than failing attestation on every request, which
-# is what a build enforcing App Check with no key would do. Create the key in
-# the Firebase console under App Check, and register a debug token there for
-# each machine running `yarn dev`.
+# reCAPTCHA v3 site key for App Check — v3, not Enterprise. The two are created
+# in different consoles and Enterprise needs a billing account, which these
+# projects do not have; the client constructs a ReCaptchaV3Provider.
+#
+# Optional: with it unset the app runs without App Check rather than failing
+# attestation on every request, which is what a build enforcing App Check with
+# no key would do. Register the key under App Check in the Firebase console,
+# and register a debug token there for each machine running `yarn dev`.
+#
+# One key covers both projects: a v3 "site" is a list of domains and belongs to
+# no Firebase project, so this value is deliberately not split dev/prod.
 VITE_RECAPTCHA_SITE_KEY=
--- a/.env.example
+++ b/.env.example
@@ -11,3 +11,10 @@ VITE_FIREBASE_PROJECT_ID=
 VITE_FIREBASE_STORAGE_BUCKET=
 VITE_FIREBASE_MESSAGING_SENDER_ID=
 VITE_FIREBASE_APP_ID=
+
+# reCAPTCHA Enterprise site key for App Check. Optional: with it unset the app
+# runs without App Check rather than failing attestation on every request, which
+# is what a build enforcing App Check with no key would do. Create the key in
+# the Firebase console under App Check, and register a debug token there for
+# each machine running `yarn dev`.
+VITE_RECAPTCHA_SITE_KEY=
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -69,6 +69,11 @@ jobs:
             exit 1
           fi
 
+      # VITE_RECAPTCHA_SITE_KEY is deliberately not PROD_-prefixed: one
+      # reCAPTCHA v3 site key covers both projects, because a v3 "site" is a
+      # list of domains and is not owned by a Firebase project. The App Check
+      # *registration* is per project and there are two of them, each holding
+      # the same secret. Enterprise keys would need splitting; v3 does not.
       - name: Build
         env:
           VITE_FIREBASE_API_KEY: ${{ secrets.PROD_VITE_FIREBASE_API_KEY }}
@@ -80,9 +85,41 @@ jobs:
           VITE_RECAPTCHA_SITE_KEY: ${{ secrets.VITE_RECAPTCHA_SITE_KEY }}
         run: yarn build
 
+      # Every PROD_ secret must exist before anything is checked against it.
+      #
+      # This job has no `environment:`, so it reads repository-level secrets
+      # only — putting the production values in the `production` environment,
+      # which is the natural reading of "production secrets", hands this job
+      # empty strings instead. Nothing downstream would catch that: `client.ts`
+      # validates the config at runtime, and `vite build` never evaluates the
+      # module, so the build stays green and the bundle throws on first paint.
+      - name: Require every production value
+        env:
+          API_KEY: ${{ secrets.PROD_VITE_FIREBASE_API_KEY }}
+          AUTH_DOMAIN: ${{ secrets.PROD_VITE_FIREBASE_AUTH_DOMAIN }}
+          PROJECT_ID: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
+          STORAGE_BUCKET: ${{ secrets.PROD_VITE_FIREBASE_STORAGE_BUCKET }}
+          SENDER_ID: ${{ secrets.PROD_VITE_FIREBASE_MESSAGING_SENDER_ID }}
+          APP_ID: ${{ secrets.PROD_VITE_FIREBASE_APP_ID }}
+        run: |
+          missing=''
+          for name in API_KEY AUTH_DOMAIN PROJECT_ID STORAGE_BUCKET SENDER_ID APP_ID; do
+            eval "value=\$$name"
+            if [ -z "$value" ]; then missing="$missing PROD_VITE_FIREBASE_$name"; fi
+          done
+          if [ -n "$missing" ]; then
+            echo "::error title=Missing secrets::Not set at repository level:$missing. Environment-scoped secrets are not visible to this job."
+            exit 1
+          fi
+
       # Proves the bundle is the one this run built, and that it points where
       # it is about to be deployed. A dist/ built against the dev project would
       # otherwise deploy happily and talk to the wrong database.
+      #
+      # `grep -q ""` matches every line and exits 0, so this check was a no-op
+      # whenever EXPECTED was empty — it failed open in exactly the case it
+      # existed for. The step above is what stops that, and this one no longer
+      # has to trust its input.
       - name: Check the bundle points at production
         env:
           EXPECTED: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -1,0 +1,156 @@
+name: Deploy (production)
+
+# Manual only. Production has no branch that deploys itself: a push to `main`
+# is a merge, and a merge is not a decision to release. The decision is this
+# button, and it names the commit it is releasing.
+on:
+  workflow_dispatch:
+    inputs:
+      sha:
+        description: 'Full commit SHA on main to deploy (40 characters)'
+        required: true
+
+permissions:
+  contents: read
+
+# Never two at once, and never cancelled halfway: a cancelled deploy between
+# the rules step and the hosting step leaves the two out of step with each
+# other, which is the one state this workflow orders itself to avoid.
+concurrency:
+  group: deploy-prod
+  cancel-in-progress: false
+
+jobs:
+  # Same split as the dev workflow and for the same reason: this job runs
+  # install scripts, so it holds no credential. Unlike dev it also has to
+  # establish that the commit is one that may be released at all.
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          # The whole history, because the ancestry check below needs main.
+          fetch-depth: 0
+          ref: ${{ inputs.sha }}
+
+      # The check that makes the input safe. Without it this workflow deploys
+      # any commit in the repository — including one that only ever existed on
+      # a branch, or in a pull request that was closed unmerged.
+      - name: Refuse a commit that is not on main
+        env:
+          SHA: ${{ inputs.sha }}
+        run: |
+          git fetch --no-tags origin main
+          if ! git merge-base --is-ancestor "$SHA" origin/main; then
+            echo "::error title=Not on main::$SHA is not an ancestor of origin/main. Production deploys only from main."
+            exit 1
+          fi
+          echo "::notice::Deploying $SHA ($(git log -1 --format='%s' "$SHA"))"
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: .nvmrc
+          cache: yarn
+
+      - run: yarn install --frozen-lockfile
+
+      # Dev warns and carries on when this is unset. Production does not: a
+      # build with no App Check ships a public Firebase config that nothing
+      # attests, and the console message saying so is seen by nobody.
+      - name: Require App Check
+        env:
+          KEY: ${{ secrets.VITE_RECAPTCHA_SITE_KEY }}
+        run: |
+          if [ -z "$KEY" ]; then
+            echo "::error title=App Check::VITE_RECAPTCHA_SITE_KEY is not set. Refusing to build production without App Check."
+            exit 1
+          fi
+
+      - name: Build
+        env:
+          VITE_FIREBASE_API_KEY: ${{ secrets.PROD_VITE_FIREBASE_API_KEY }}
+          VITE_FIREBASE_AUTH_DOMAIN: ${{ secrets.PROD_VITE_FIREBASE_AUTH_DOMAIN }}
+          VITE_FIREBASE_PROJECT_ID: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
+          VITE_FIREBASE_STORAGE_BUCKET: ${{ secrets.PROD_VITE_FIREBASE_STORAGE_BUCKET }}
+          VITE_FIREBASE_MESSAGING_SENDER_ID: ${{ secrets.PROD_VITE_FIREBASE_MESSAGING_SENDER_ID }}
+          VITE_FIREBASE_APP_ID: ${{ secrets.PROD_VITE_FIREBASE_APP_ID }}
+          VITE_RECAPTCHA_SITE_KEY: ${{ secrets.VITE_RECAPTCHA_SITE_KEY }}
+        run: yarn build
+
+      # Proves the bundle is the one this run built, and that it points where
+      # it is about to be deployed. A dist/ built against the dev project would
+      # otherwise deploy happily and talk to the wrong database.
+      - name: Check the bundle points at production
+        env:
+          EXPECTED: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
+        run: |
+          if ! grep -rq "$EXPECTED" dist/assets/*.js; then
+            echo "::error title=Wrong project::The built bundle does not contain $EXPECTED."
+            exit 1
+          fi
+          if grep -rq 'goitei-dev' dist/assets/*.js; then
+            echo "::error title=Wrong project::The built bundle mentions goitei-dev."
+            exit 1
+          fi
+
+      - uses: actions/upload-artifact@v7
+        with:
+          name: dist-prod
+          path: dist
+          retention-days: 1
+
+  # Holds the credentials, runs no install script from the build, and is gated
+  # on the `production` environment — which is where a required reviewer or a
+  # wait timer is configured. The gate is in GitHub rather than in this file so
+  # that changing it does not need a commit.
+  deploy:
+    needs: build
+    runs-on: ubuntu-latest
+    environment: production
+    permissions:
+      contents: read
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.sha }}
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: .nvmrc
+          cache: yarn
+
+      # Before authenticating, so no postinstall script sees a credential.
+      - run: yarn install --frozen-lockfile
+
+      - uses: actions/download-artifact@v8
+        with:
+          name: dist-prod
+          path: dist
+
+      - uses: google-github-actions/auth@v3
+        with:
+          workload_identity_provider: ${{ secrets.GCP_PROD_WORKLOAD_IDENTITY_PROVIDER }}
+          service_account: ${{ secrets.GCP_PROD_DEPLOY_SERVICE_ACCOUNT }}
+
+      # Rules first, hosting second, and never in one command.
+      #
+      # The window between them is the whole reason for the order. Rules ahead
+      # of the client means the client is briefly older than the rules it runs
+      # under, which fails closed — a request the new rules refuse. Hosting
+      # first means a client is briefly newer than its rules, which fails open:
+      # it writes a shape the old rules still permit.
+      - name: Deploy Firestore rules
+        run: yarn firebase deploy --only firestore:rules --project prod --non-interactive
+
+      - name: Deploy Hosting
+        run: yarn firebase deploy --only hosting --project prod --non-interactive
+
+      - name: Record what was released
+        env:
+          SHA: ${{ inputs.sha }}
+        run: echo "::notice title=Released::$SHA to goitei"
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -97,19 +97,32 @@ jobs:
       # empty strings instead. Nothing downstream would catch that: `client.ts`
       # validates the config at runtime, and `vite build` never evaluates the
       # module, so the build stays green and the bundle throws on first paint.
+      #
+      # The names are listed in full rather than rebuilt from a short loop
+      # variable. Five of the six matched; `SENDER_ID` printed
+      # `PROD_VITE_FIREBASE_SENDER_ID` while the secret read is
+      # `PROD_VITE_FIREBASE_MESSAGING_SENDER_ID` — so the one run this step
+      # exists for sent the reader off to create a secret nothing reads.
       - name: Require every production value
         env:
-          API_KEY: ${{ secrets.PROD_VITE_FIREBASE_API_KEY }}
-          AUTH_DOMAIN: ${{ secrets.PROD_VITE_FIREBASE_AUTH_DOMAIN }}
-          PROJECT_ID: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
-          STORAGE_BUCKET: ${{ secrets.PROD_VITE_FIREBASE_STORAGE_BUCKET }}
-          SENDER_ID: ${{ secrets.PROD_VITE_FIREBASE_MESSAGING_SENDER_ID }}
-          APP_ID: ${{ secrets.PROD_VITE_FIREBASE_APP_ID }}
+          PROD_VITE_FIREBASE_API_KEY: ${{ secrets.PROD_VITE_FIREBASE_API_KEY }}
+          PROD_VITE_FIREBASE_AUTH_DOMAIN: ${{ secrets.PROD_VITE_FIREBASE_AUTH_DOMAIN }}
+          PROD_VITE_FIREBASE_PROJECT_ID: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
+          PROD_VITE_FIREBASE_STORAGE_BUCKET: ${{ secrets.PROD_VITE_FIREBASE_STORAGE_BUCKET }}
+          PROD_VITE_FIREBASE_MESSAGING_SENDER_ID: ${{ secrets.PROD_VITE_FIREBASE_MESSAGING_SENDER_ID }}
+          PROD_VITE_FIREBASE_APP_ID: ${{ secrets.PROD_VITE_FIREBASE_APP_ID }}
         run: |
           missing=''
-          for name in API_KEY AUTH_DOMAIN PROJECT_ID STORAGE_BUCKET SENDER_ID APP_ID; do
+          for name in \
+            PROD_VITE_FIREBASE_API_KEY \
+            PROD_VITE_FIREBASE_AUTH_DOMAIN \
+            PROD_VITE_FIREBASE_PROJECT_ID \
+            PROD_VITE_FIREBASE_STORAGE_BUCKET \
+            PROD_VITE_FIREBASE_MESSAGING_SENDER_ID \
+            PROD_VITE_FIREBASE_APP_ID
+          do
             eval "value=\$$name"
-            if [ -z "$value" ]; then missing="$missing PROD_VITE_FIREBASE_$name"; fi
+            if [ -z "$value" ]; then missing="$missing $name"; fi
           done
           if [ -n "$missing" ]; then
             echo "::error title=Missing secrets::Not set at repository level:$missing. Environment-scoped secrets are not visible to this job."
@@ -185,6 +198,13 @@ jobs:
       # under, which fails closed — a request the new rules refuse. Hosting
       # first means a client is briefly newer than its rules, which fails open:
       # it writes a shape the old rules still permit.
+      #
+      # **Two steps come before this workflow runs, and it cannot check them.**
+      # These rules gate on a custom claim, which reaches a client only in a
+      # freshly minted ID token — so every account must be granted with
+      # `yarn allow` and must sign in again first, or it is locked out for up to
+      # an hour. See "Operator runbook" in README.md, which is in the working
+      # tree after merge in a way a pull request description is not.
       - name: Deploy Firestore rules
         run: yarn firebase deploy --only firestore:rules --project prod --non-interactive
 
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -76,6 +76,10 @@ jobs:
       # the same secret. Enterprise keys would need splitting; v3 does not.
       - name: Build
         env:
+          # The released commit, not the workflow's. On a workflow_dispatch run
+          # GITHUB_SHA is the default branch's head, so a rollback would ship a
+          # footer naming source that was never running — see build-info.ts.
+          DEPLOY_SHA: ${{ inputs.sha }}
           VITE_FIREBASE_API_KEY: ${{ secrets.PROD_VITE_FIREBASE_API_KEY }}
           VITE_FIREBASE_AUTH_DOMAIN: ${{ secrets.PROD_VITE_FIREBASE_AUTH_DOMAIN }}
           VITE_FIREBASE_PROJECT_ID: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
--- a/admin/allow-user.ts
+++ b/admin/allow-user.ts
@@ -1,0 +1,65 @@
+import { cert, getApps, initializeApp } from 'firebase-admin/app';
+import { getAuth } from 'firebase-admin/auth';
+import { applicationDefault } from 'firebase-admin/app';
+
+/**
+ * Grant or revoke access, by custom claim.
+ *
+ *   yarn allow you@example.com          # grant on goitei-dev
+ *   yarn allow you@example.com --revoke
+ *   yarn allow you@example.com prod
+ *
+ * The security rules read `request.auth.token.allowed`, so this is the only
+ * thing that opens the door. It replaced an `allowedUsers` document that rules
+ * checked with `exists()` on every request — one billed read per request, to
+ * answer a question whose answer changes twice in an account's lifetime.
+ *
+ * **Revoking is two operations, and doing only the first is the mistake.** A
+ * claim lives inside the ID token the client already holds, so clearing it
+ * changes nothing until that token expires — up to an hour of continued access
+ * after a ban. `revokeRefreshTokens` forces the client to re-authenticate, and
+ * the new token is the one without the claim. Both run here, in that order.
+ */
+
+const [email, ...rest] = process.argv.slice(2);
+const revoke = rest.includes('--revoke');
+const env = rest.includes('prod') ? 'prod' : 'dev';
+
+if (!email) {
+  console.error('usage: allow-user.ts <email> [prod] [--revoke]');
+  process.exit(1);
+}
+
+const projectId = env === 'prod' ? 'goitei' : 'goitei-dev';
+
+if (getApps().length === 0) {
+  const key = process.env.GOOGLE_APPLICATION_CREDENTIALS;
+  initializeApp({
+    credential: key ? cert(key) : applicationDefault(),
+    projectId,
+  });
+}
+
+const auth = getAuth();
+
+const user = await auth.getUserByEmail(email).catch(() => null);
+if (!user) {
+  // Deliberately not created here: the uid has to be the one Google issued, so
+  // the account must have signed in once before it can be allowed.
+  console.error(`${email} has never signed in to ${projectId}; ask them to try once first.`);
+  process.exit(1);
+}
+
+const claims = { ...user.customClaims, allowed: revoke ? undefined : true };
+await auth.setCustomUserClaims(user.uid, claims);
+
+if (revoke) {
+  // Without this the client keeps its existing token, claim and all, until it
+  // expires. This is what makes a ban take effect now rather than within an hour.
+  await auth.revokeRefreshTokens(user.uid);
+}
+
+console.log(
+  `${revoke ? 'revoked' : 'allowed'}: ${email} (${user.uid}) on ${projectId}` +
+    (revoke ? ' — refresh tokens revoked, the client must sign in again' : ''),
+);
--- a/admin/allow-user.ts
+++ b/admin/allow-user.ts
@@ -14,11 +14,16 @@ import { applicationDefault } from 'firebase-admin/app';
  * checked with `exists()` on every request — one billed read per request, to
  * answer a question whose answer changes twice in an account's lifetime.
  *
- * **Revoking is two operations, and doing only the first is the mistake.** A
- * claim lives inside the ID token the client already holds, so clearing it
- * changes nothing until that token expires — up to an hour of continued access
- * after a ban. `revokeRefreshTokens` forces the client to re-authenticate, and
- * the new token is the one without the claim. Both run here, in that order.
+ * **A ban lands within the hour, not on the keystroke.** A claim lives inside
+ * the ID token the client already holds, and Firestore rules have no revocation
+ * check, so clearing the claim changes nothing until that token expires — up to
+ * an hour. Revoking the refresh tokens does not shorten that: it stops the next
+ * refresh from succeeding, which ends the session, but the token already in
+ * hand keeps its claims until it runs out.
+ *
+ * It is done anyway, because ending the session is worth doing on its own. What
+ * it is not is immediate, and an operator responding to abuse has to know that
+ * — deleting the offending data is the part that takes effect now.
  */
 
 const [email, ...rest] = process.argv.slice(2);
@@ -54,12 +59,14 @@ const claims = { ...user.customClaims, allowed: revoke ? undefined : true };
 await auth.setCustomUserClaims(user.uid, claims);
 
 if (revoke) {
-  // Without this the client keeps its existing token, claim and all, until it
-  // expires. This is what makes a ban take effect now rather than within an hour.
+  // Ends the session at the next refresh. It does not invalidate the token the
+  // client is holding right now — see the note above.
   await auth.revokeRefreshTokens(user.uid);
 }
 
 console.log(
   `${revoke ? 'revoked' : 'allowed'}: ${email} (${user.uid}) on ${projectId}` +
-    (revoke ? ' — refresh tokens revoked, the client must sign in again' : ''),
+    (revoke
+      ? ' — refresh tokens revoked. The ID token already issued stays valid until it expires, so access ends within the hour rather than immediately.'
+      : ''),
 );
--- a/admin/allow-user.ts
+++ b/admin/allow-user.ts
@@ -1,6 +1,5 @@
-import { cert, getApps, initializeApp } from 'firebase-admin/app';
+import { applicationDefault, cert, getApps, initializeApp } from 'firebase-admin/app';
 import { getAuth } from 'firebase-admin/auth';
-import { applicationDefault } from 'firebase-admin/app';
 
 /**
  * Grant or revoke access, by custom claim.
@@ -27,13 +26,21 @@ import { applicationDefault } from 'firebase-admin/app';
  */
 
 const [email, ...rest] = process.argv.slice(2);
-const revoke = rest.includes('--revoke');
-const env = rest.includes('prod') ? 'prod' : 'dev';
 
-if (!email) {
+// Rejected rather than ignored, because the mistake this catches is silent and
+// natural: `--revoke` is spelled with dashes, so `--prod` is how one reaches
+// for the other — and `rest.includes('prod')` is false for it, so the run would
+// grant access on **dev** and say so only in its last line.
+const unknown = rest.filter((arg) => arg !== 'prod' && arg !== '--revoke');
+
+if (!email || unknown.length > 0) {
+  if (unknown.length > 0) console.error(`unknown argument: ${unknown.join(' ')}`);
   console.error('usage: allow-user.ts <email> [prod] [--revoke]');
   process.exit(1);
 }
+
+const revoke = rest.includes('--revoke');
+const env = rest.includes('prod') ? 'prod' : 'dev';
 
 const projectId = env === 'prod' ? 'goitei' : 'goitei-dev';
 
--- a/firestore.rules
+++ b/firestore.rules
@@ -68,17 +68,144 @@ service cloud.firestore {
     //
     // To open signups: delete this function and its call sites. Do that
     // together with App Check and a budget-alert kill switch, not before.
+    //
+    // **A custom claim, not a document read.** exists() here is billed as a
+    // document read on *every request to every path*, which on a public service
+    // doubles the read cost of the whole application to answer a question whose
+    // answer changes perhaps twice in an account's lifetime.
+    //
+    // What that trades away is immediacy, and it is worth stating: a claim
+    // lives in the ID token, so removing it takes effect when the token next
+    // refreshes — up to an hour. Revocation is therefore two steps, not one:
+    // clear the claim *and* revoke the refresh tokens, which forces the client
+    // to re-authenticate immediately. `admin/allow-user.ts` does both.
     function isAllowed() {
-      return signedIn()
-        && exists(/databases/$(database)/documents/allowedUsers/$(request.auth.uid));
+      return signedIn() && request.auth.token.allowed == true;
     }
 
-    // Everything private, in one rule. No field-level conditions, no visibility
-    // flags, nothing to get subtly wrong.
+    // ---- shape and size ----
+    //
+    // These exist for one threat and it is not shape: an account that is
+    // legitimately signed in can write **as many documents as it likes, as
+    // large as it likes**, and on a metered project that is somebody else's
+    // bill. Firestore caps a document at 1 MiB, which is 1 MiB per document and
+    // no limit at all on how many.
+    //
+    // **What rules cannot check is worth knowing before trusting this.** There
+    // is no way to iterate a list, so the length of `senses[3].description` is
+    // invisible here — only the number of senses and the size of the top-level
+    // strings can be bounded. Anything relying on per-element limits has to
+    // enforce them somewhere that can see them.
+    //
+    // Deliberately not a full schema. Requiring every field to exist and match
+    // a type turns every schema change into a rules deploy, and a rules deploy
+    // that lags a client release denies writes from a version already shipped.
+    function str(value, max) {
+      return value is string && value.size() <= max;
+    }
+
+    function list(value, max) {
+      return value is list && value.size() <= max;
+    }
+
+    function ownedBy(uid) {
+      return request.resource.data.ownerUid == uid;
+    }
+
+    // Bookkeeping a client may never rewrite. On an update `request.resource`
+    // is the merged result, so an untouched field compares equal to itself.
+    //
+    // `resource == null` first, and not for tidiness: on a create there is no
+    // existing document, and reaching into `resource.data` there is a null
+    // dereference — which rules report as an evaluation error and treat as a
+    // denial, so every create was refused with a message about null rather than
+    // about ownership.
+    function keepsCreatedAt() {
+      return resource == null
+        || !('createdAt' in resource.data)
+        || request.resource.data.createdAt == resource.data.createdAt;
+    }
+
+    function validEntry(uid) {
+      let d = request.resource.data;
+      return ownedBy(uid)
+        && d.keys().size() <= 40
+        && str(d.headword, 200) && d.headword.size() > 0
+        && str(d.reading, 200)
+        && str(d.definition, 20000)
+        && str(d.definitionSub, 20000)
+        && str(d.source, 500)
+        && str(d.citationForm, 200)
+        && list(d.pos, 20)
+        && list(d.tags, 50)
+        && list(d.senses, 100)
+        && list(d.examples, 100)
+        && list(d.related, 100)
+        && (d.pitchAccent == null || d.pitchAccent is int);
+    }
+
+    function validWordSet(uid) {
+      let d = request.resource.data;
+      return ownedBy(uid)
+        && d.keys().size() <= 20
+        && str(d.name, 200) && d.name.size() > 0
+        && str(d.description, 5000)
+        // The one unbounded list in the schema, and the one a set is made of.
+        && list(d.entryIds, 2000)
+        && list(d.topics, 50);
+    }
+
+    function validSession(uid) {
+      let d = request.resource.data;
+      return ownedBy(uid)
+        && d.keys().size() <= 20
+        && str(d.filterLabel, 500)
+        && d.total is int && d.total >= 0 && d.total <= 2000
+        && d.correct is int && d.correct >= 0 && d.correct <= d.total
+        && list(d.missed, 2000);
+    }
+
+    // Private data, one collection at a time.
+    //
+    // **There is deliberately no `match /{document=**}` here any more.** Rules
+    // are a union: a recursive wildcard granting write is an unconditional
+    // allow that every validation below would be evaluated *beside* rather than
+    // *after*, so leaving it in place while adding shape checks would have
+    // looked like hardening and enforced nothing.
+    //
+    // The cost of enumerating is that a collection nobody listed is denied
+    // rather than allowed. That is the right way round, and it is why adding a
+    // path to the app now means adding it here too.
     match /users/{uid} {
-      allow read, write: if isOwner(uid) && isAllowed();
-      match /{document=**} {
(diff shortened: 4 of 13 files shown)


firestore.rules
```diff
@@ -68,17 +68,144 @@ service cloud.firestore {
...
+    // **A custom claim, not a document read.** exists() here is billed as a
+    // document read on *every request to every path*, which on a public service
+    // doubles the read cost of the whole application to answer a question whose
+    // answer changes perhaps twice in an account's lifetime.
+    //
+    // What that trades away is immediacy, and it is worth stating: a claim
+    // lives in the ID token, so removing it takes effect when the token next
+    // refreshes — up to an hour. Revocation is therefore two steps, not one:
+    // clear the claim *and* revoke the refresh tokens, which forces the client
+    // to re-authenticate immediately. `admin/allow-user.ts` does both.
```
kowinauyeung: **🔴 BLOCKER — the two-step revocation does not buy back immediacy, and three places here say it does.**

This comment says revoking the refresh tokens "forces the client to re-authenticate immediately". `admin/allow-user.ts:20` says the same, and its line 57 goes further: "This is what makes a ban take effect now rather than within an hour."

`revokeRefreshTokens()` invalidates the *refresh* token. An ID token already issued stays valid until it expires, and Firestore rules do not consult revocation state — there is no equivalent of `verifyIdToken(..., checkRevoked: true)` in the rules evaluator. The JS SDK holds its current ID token and refreshes it only shortly before expiry, so a banned account keeps `allowed: true` in the token it is presenting for up to about an hour. That is exactly the window the comment claims to have closed.

Worth stating plainly because of what it does to the trade-off written above it: the `exists()` gate this replaces *was* immediate — it was re-evaluated per request. The claim gate is not, and revoking refresh tokens does not change that. The paragraph reaches the right conclusion about cost and the wrong one about immediacy.

Two ways out, and the choice is a real one:

- **Accept the window.** Correct the wording here, in `admin/allow-user.ts`, and in the operator runbook, so that whoever is responding to an abuse incident knows a ban lands within the hour rather than on the keystroke.
- **Actually close it**, with Firebase's documented pattern: store a revocation timestamp and compare `request.auth.token.auth_time` against it in `isAllowed()`. That reintroduces the per-request document read this PR removed, which may well not be worth it for a two-person allowlist — but it should be a decision, not an assumption.

The test at line 184 below is honest about this ("this test is the state *after* that has happened"), so the gap is already known at the test layer; it is the prose that overstates.

*Reasoned from documented Firebase behaviour rather than measured: `@firebase/rules-unit-testing` mints tokens directly, so the harness structurally cannot exercise revocation.*

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,156 @@
...
+
+      - name: Build
+        env:
+          VITE_FIREBASE_API_KEY: ${{ secrets.PROD_VITE_FIREBASE_API_KEY }}
+          VITE_FIREBASE_AUTH_DOMAIN: ${{ secrets.PROD_VITE_FIREBASE_AUTH_DOMAIN }}
+          VITE_FIREBASE_PROJECT_ID: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
+          VITE_FIREBASE_STORAGE_BUCKET: ${{ secrets.PROD_VITE_FIREBASE_STORAGE_BUCKET }}
+          VITE_FIREBASE_MESSAGING_SENDER_ID: ${{ secrets.PROD_VITE_FIREBASE_MESSAGING_SENDER_ID }}
+          VITE_FIREBASE_APP_ID: ${{ secrets.PROD_VITE_FIREBASE_APP_ID }}
+          VITE_RECAPTCHA_SITE_KEY: ${{ secrets.VITE_RECAPTCHA_SITE_KEY }}
```
kowinauyeung, line 89: **🔴 BLOCKER — production builds with the unprefixed reCAPTCHA secret while every other production value is `PROD_`-prefixed.**

Lines 74-79 read `PROD_VITE_FIREBASE_*`; this line reads `secrets.VITE_RECAPTCHA_SITE_KEY`, the same secret `deploy-dev.yml:72` uses.

An App Check registration is per Firebase project, and a reCAPTCHA v3 key is bound to a domain list. If dev and prod have separate registrations — which is the default and matches how every other secret here is split — the production bundle ships a key that cannot produce a valid App Check token for `goitei`.

The consequence is worse than "no protection", because of how this PR sequences enforcement: the plan is to ship the client, watch the App Check metrics, and turn enforcement on once real traffic is attesting. Those metrics would never go green, and the reason would not be visible from them.

If one key really is registered in both projects deliberately, that is fine — but then it needs a comment, because the asymmetry with the six lines above it reads as an oversight. Otherwise: `PROD_VITE_RECAPTCHA_SITE_KEY`.

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,156 @@
...
+        run: yarn build
+
+      # Proves the bundle is the one this run built, and that it points where
+      # it is about to be deployed. A dist/ built against the dev project would
+      # otherwise deploy happily and talk to the wrong database.
+      - name: Check the bundle points at production
+        env:
+          EXPECTED: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
+        run: |
+          if ! grep -rq "$EXPECTED" dist/assets/*.js; then
```
kowinauyeung, line 144: **🔴 BLOCKER — the guard that stops a wrong-project deploy fails open.**

`grep -q ""` matches every file and exits 0, so an empty `EXPECTED` turns this check into a no-op. Verified locally:

```console
$ printf 'hello\n' > f; grep -q "" f; echo $?
0
```

`EXPECTED` can be empty in a way that is easy to reach: the `PROD_*` secrets are read in the `build` job, which has no `environment:` — only `deploy` does, at line 112. So they have to be repository-level secrets. Putting them in the `production` environment instead, which is the natural reading of "production secrets", hands this job empty strings.

Nothing upstream catches that either. `client.ts:26` does check for missing config, but it throws at *runtime*; `vite build` never evaluates the module, so the build stays green. The chain is: empty secret → green build → green guard → a bundle deployed to production that throws on first paint.

The other half of the check (line 94, `grep -rq 'goitei-dev'`) is a literal and so is unaffected.

```yaml
- name: Check the bundle points at production
  env:
    EXPECTED: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
  run: |
    if [ -z "$EXPECTED" ]; then
      echo "::error title=Wrong project::PROD_VITE_FIREBASE_PROJECT_ID is not set."
      exit 1
    fi
    if ! grep -rq "$EXPECTED" dist/assets/*.js; then
```

Worth extending to the other five `PROD_VITE_FIREBASE_*` values while you are there — the same empty-secret path ships a bundle missing any of them.

firestore.rules
```diff
@@ -68,17 +68,144 @@ service cloud.firestore {
...
+    }
+
+    function validEntry(uid) {
+      let d = request.resource.data;
+      return ownedBy(uid)
+        && d.keys().size() <= 40
+        && str(d.headword, 200) && d.headword.size() > 0
+        && str(d.reading, 200)
+        && str(d.definition, 20000)
+        && str(d.definitionSub, 20000)
```
kowinauyeung: **🟡 These rules are a full required-field schema, which is the opposite of what the comment above them says.**

The comment at line 122 reads: "Deliberately not a full schema. Requiring every field to exist and match a type turns every schema change into a rules deploy, and a rules deploy that lags a client release denies writes from a version already shipped."

`str(d.definitionSub, 20000)` reaches into a key that may not be there, and in rules that is an evaluation error, which is a denial. So every field named in `validEntry` *is* required. Verified against the emulator at `855e9bf`:

```
create without definitionSub → PERMISSION_DENIED: Property definitionSub is undefined on object. @ L189
create without citationForm  → PERMISSION_DENIED: Property citationForm is undefined on object. @ L189
create without pitchAccent   → PERMISSION_DENIED: Property pitchAccent is undefined on object. @ L189
create without tags          → PERMISSION_DENIED: Property tags is undefined on object. @ L189
```

Nothing breaks today: every write path goes through `emptyDraft()` or `sanitizeEntry` + `toDraft`, which always emit the full set. The problem is the next person. The comment tells them an optional field costs no rules deploy, and it costs exactly one — and `deploy-prod.yml` deploys rules *before* hosting, so the rules that reject the new field are live before the client that fills it.

Either make the genuinely-optional ones `d.get('definitionSub', '')`, or drop the paragraph. Right now the file documents a property it does not have.

firestore.rules
```diff
@@ -68,17 +68,144 @@ service cloud.firestore {
...
+        && str(d.source, 500)
+        && str(d.citationForm, 200)
+        && list(d.pos, 20)
+        && list(d.tags, 50)
+        && list(d.senses, 100)
```
kowinauyeung: **🟡 Worth a sentence: the per-document ceiling is still roughly 1 MiB.**

The note at line 116 about lists being un-iterable is correct and is the right thing to have written down. The consequence is worth making explicit next to these bounds, because the PR title says "bound writes": `senses` is capped at 100 *elements* and each element is unbounded, so a document of 100 × 3,000 characters is accepted. Verified against the emulator.

Not a defect — it is the honest limit of what rules can see. But a reader who has just read "bounded" will assume a size bound, and the bound is on cardinality.

firestore.rules
```diff
@@ -68,17 +68,144 @@ service cloud.firestore {
...
     match /users/{uid} {
-      allow read, write: if isOwner(uid) && isAllowed();
-      match /{document=**} {
-        allow read, write: if isOwner(uid) && isAllowed();
+      function mine() {
+        return isOwner(uid) && isAllowed();
+      }
+
+      allow read: if mine();
+      allow write: if mine() && request.resource.data.keys().size() <= 20;
```
kowinauyeung: **🟡 This denies deleting `users/{uid}` itself, via a null dereference.**

`write` covers delete, and on a delete `request.resource` is null, so `request.resource.data.keys()` errors out. Verified:

```
delete users/{uid} → PERMISSION_DENIED: Null value error. for 'delete' @ L185
```

This is the same trap `keepsCreatedAt` guards against at line 108, and the three subcollections below all split `read, delete` from `create, update` correctly — the parent document is the one place that did not get the split.

Nothing deletes it today. But account deletion is the obvious next caller, and it would fail with a message about null rather than about permission — which is precisely the failure mode the `keepsCreatedAt` comment describes as worth avoiding.

```
allow read: if mine();
allow delete: if mine();
allow create, update: if mine() && request.resource.data.keys().size() <= 20;
```

firestore.rules
```diff
@@ -68,17 +68,144 @@ service cloud.firestore {
...
+        allow create, update: if mine() && validSession(uid) && keepsCreatedAt();
+      }
+
+      // One document holding every entry's progress — see ProgressRepository
+      // for why it is a map rather than a collection. Its size is bounded by
+      // the 1 MiB document cap and by the notebook it mirrors, so the only
+      // thing worth asserting is that it belongs to the owner.
+      match /progress/{documentId} {
+        allow read, delete: if mine();
+        allow create, update: if mine();
```
kowinauyeung: **🟡 `progress` is the one path with no bounds at all, and it is a wildcard.**

The section header at line 112 states the threat: an account that is legitimately signed in "can write as many documents as it likes, as large as it likes, and on a metered project that is somebody else's bill". This path is that threat, unmitigated. Verified against the emulator — an allowed account can create `progress/junk-1`, `progress/junk-2`, … each holding a 500 KB string, all accepted.

The comment above says "the only thing worth asserting is that it belongs to the owner", but `mine()` asserts the owner of the *path*, not the `ownerUid` *field* — and `progressRepo.recordSession` does write `ownerUid: uid`. So this is also the only write path in the file where that field goes unchecked while every sibling checks it.

`progressRepo.ts:39` only ever addresses `progress/entries`, one fixed id, so narrowing costs nothing:

```
match /progress/entries {
  allow read, delete: if mine();
  allow create, update: if mine() && ownedBy(uid)
    && request.resource.data.keys().size() <= 5;
}
```

The 1 MiB cap the comment relies on is a per-document cap, and the wildcard is what makes the document count unbounded.

admin/allow-user.ts
```diff
@@ -0,0 +1,65 @@
+import { cert, getApps, initializeApp } from 'firebase-admin/app';
```
kowinauyeung: **🟡 `admin/` is outside the ESLint scope, so CI never sees this file.**

`eslint.config.js` has no `files` block matching `admin/**` — the widest is `migration/**/*.mjs` at line 208. Running it at this commit:

```console
$ eslint admin
ESLint: You are linting "admin", but all of the files matching the glob pattern "admin" are ignored.
```

`yarn lint` runs `eslint .`, which does not error on a directory it matches nothing in, so the Lint step in `ci.yml:38` stays green while skipping the script that grants production access. Adding `admin/**/*.ts` to the config would also have caught the duplicate import three lines down.

admin/allow-user.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getAuth } from 'firebase-admin/auth';
+import { applicationDefault } from 'firebase-admin/app';
```
kowinauyeung: 🔵 Second import from `firebase-admin/app`; `applicationDefault` can join line 1.

admin/allow-user.ts
```diff
@@ -0,0 +1,65 @@
...
+ * answer a question whose answer changes twice in an account's lifetime.
+ *
+ * **Revoking is two operations, and doing only the first is the mistake.** A
+ * claim lives inside the ID token the client already holds, so clearing it
+ * changes nothing until that token expires — up to an hour of continued access
+ * after a ban. `revokeRefreshTokens` forces the client to re-authenticate, and
+ * the new token is the one without the claim. Both run here, in that order.
+ */
+
+const [email, ...rest] = process.argv.slice(2);
```
kowinauyeung, line 28: 🔵 Unknown arguments are accepted silently, so `yarn allow you@example.com --prod` targets **dev** — `rest.includes('prod')` is false for `--prod`, and `--revoke` is the only flag spelled with dashes, which makes the mistake a natural one.

The success line at 62 does name the project, so it is recoverable rather than dangerous. Rejecting anything that is not `prod` or `--revoke` is cheaper than noticing.

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,156 @@
...
+
+      # Rules first, hosting second, and never in one command.
+      #
+      # The window between them is the whole reason for the order. Rules ahead
+      # of the client means the client is briefly older than the rules it runs
+      # under, which fails closed — a request the new rules refuse. Hosting
+      # first means a client is briefly newer than its rules, which fails open:
+      # it writes a shape the old rules still permit.
+      - name: Deploy Firestore rules
+        run: yarn firebase deploy --only firestore:rules --project prod --non-interactive
```
kowinauyeung, line 209: **🟡 The deploy order that makes this safe lives only in the pull request description.**

The three steps in the description — `yarn allow` first, sign out and back in, *then* deploy the rules — are what stop the claim-gated rules from locking everybody out for up to an hour. This step automates the third one and nothing carries the first two.

That matters more than usual here because the first production run of this workflow *is* the migration: it puts `isAllowed()` on a claim that no live token has yet. A description is not in the working tree after merge.

Worth a runbook next to this file, or a short block in the README that the workflow's own comment points at.

src/infra/firebase/client.ts
```diff
@@ -31,6 +32,65 @@ if (missing.length) {
...
+ * The debug token below is for `yarn dev` only. It is compiled out of a
+ * production build by the mode check, so it cannot ship: a debug token is a
+ * bypass, and a bypass in a released bundle is the absence of App Check with
+ * extra steps. Verified against a real build rather than assumed — the string
+ * does appear in the bundle, but every occurrence is the SDK reading it, and
+ * the assignment is not there.
+ *
+ * **This needs a Content-Security-Policy change before the policy is enforced.**
+ * reCAPTCHA Enterprise loads from `https://www.google.com/recaptcha/` and frames
+ * a challenge from the same origin, and the policy in `firebase.json` allows
```
kowinauyeung: **🟡 This describes a Content-Security-Policy that does not exist on this branch.**

`firebase.json` on `develop`, and at this commit, has no `headers` block at all — the Report-Only policy is on the other branch. So the claim "the policy in `firebase.json` allows neither today. It is Report-Only" cannot be checked by a reviewer reading this branch, and whichever of the two merges second silently decides whether the hosts are ever added.

The technical point is right and worth keeping. Two refinements:

- "reCAPTCHA Enterprise" is stale after 07d88d3 — v3 is what is imported two lines from here.
- v3 pulls `https://www.gstatic.com/recaptcha/…` as well as `https://www.google.com/recaptcha/`. gstatic is already in `script-src`; naming both makes the eventual edit copy-pasteable rather than a re-derivation.

Since promoting the policy and breaking App Check would be silent in both directions, this is a good candidate for a check that fails: a test, or a grep in CI, asserting `www.google.com` is in `script-src` and `frame-src` before `Content-Security-Policy-Report-Only` becomes `Content-Security-Policy`.

tests/rules/firestore-rules.test.ts
```diff
@@ -257,3 +312,159 @@ describe('collections that no longer exist', () => {
...
+    await assertFails(db.doc(`users/${ALICE}/somethingNew/x`).set({ ownerUid: ALICE }));
+  });
+});
+
+/**
+ * The shapes the app actually writes, against the rules that will judge them.
+ *
+ * This exists because nothing else checks it. `tests/integration` connects with
+ * `mockUserToken: 'owner'`, which is the emulator's owner credential and
+ * **bypasses rules entirely** — so a green adapter suite says nothing about
```
kowinauyeung, line 404: **🟡 This block only exercises creates, and the interesting path is update.**

The gap it was written to close is real and well described. But `entryRepo.update` (`entryRepo.ts:115`) is the call that meets `keepsCreatedAt` against an actual stored document — a create takes the `resource == null` branch and never touches the comparison. `wordSetRepo.update` re-sends the whole draft over an existing set, same story.

One update case each would cover what this block exists for. `serverTimestamp()` in the merged result versus a stored `Timestamp` is exactly the kind of thing that passes by hand and fails for real.

Also unpinned in either direction: the required-field behaviour in `validEntry` (see the comment on `firestore.rules:136`). Whichever way that decision goes, a test naming it would stop the comment and the rules drifting apart again.

.env.example
```diff
@@ -11,3 +11,10 @@ VITE_FIREBASE_PROJECT_ID=
 VITE_FIREBASE_STORAGE_BUCKET=
 VITE_FIREBASE_MESSAGING_SENDER_ID=
 VITE_FIREBASE_APP_ID=
+
+# reCAPTCHA Enterprise site key for App Check. Optional: with it unset the app
```
kowinauyeung: 🔵 "reCAPTCHA **Enterprise** site key" is stale after 07d88d3 — the client constructs a `ReCaptchaV3Provider`. Worth fixing here specifically: this file is the instruction someone follows when provisioning, and the two key types are created in different consoles, with Enterprise needing the billing account this PR moved away from.

tsconfig.scripts.json
```diff
@@ -10,6 +11,8 @@
   // folder writes is what goes to production.
   "compilerOptions": {
     "target": "ES2022",
+    // Top-level await, which the admin scripts are written with.
+    "module": "ESNext",
```
kowinauyeung: 🔵 `"module": "ESNext"` is now declared twice — once here and once at line 17. `tsc -p tsconfig.scripts.json --noEmit` accepts it (verified, exit 0), so this is cosmetic; the new comment above reads as if it is introducing the setting.

kowinauyeung:
## Review summary

**Verdict: `REQUEST CHANGES`** — three blockers, all small. Fifteen inline comments above.

The shape of this change is right and the reasoning is written down where it will be read. Removing `match /{document=**}` is the part everything else depends on, and proving it by putting the wildcard back and naming the seven cases that went green is the correct way to establish it. The `mockUserToken: 'owner'` gap in `tests/integration` is a suite that has been claiming coverage it never had; finding that while writing something else, and then closing it, is the good outcome.

What is blocking is narrower than the change itself.

### Blockers

1. **Revocation is not immediate, and `firestore.rules:80`, `admin/allow-user.ts:20` and `admin/allow-user.ts:57` all say it is.** `revokeRefreshTokens()` invalidates the refresh token; the ID token already in the client's hands keeps its `allowed: true` until it expires, and rules have no revocation check. The `exists()` gate being replaced *was* immediate. Either correct the wording and the operator runbook, or adopt the `auth_time` comparison and pay the read back. Full detail on `firestore.rules:81`.
2. **`deploy-prod.yml:80` builds production with the dev reCAPTCHA secret** — the only value on that step without the `PROD_` prefix. The failure is quiet in the way that matters: the App Check metrics this PR plans to gate enforcement on would never go green, and would not say why.
3. **`deploy-prod.yml:90` fails open on an empty secret.** `grep -q ""` matches everything, `vite build` never runs the config check in `client.ts`, and the `PROD_*` secrets are read in a job with no `environment:`. Empty secret → green build → green guard → a white screen in production.

### Also worth resolving before merge

- **`firestore.rules` is a full required-field schema and the comment above it says it deliberately is not.** Confirmed on the emulator: omitting `definitionSub`, `citationForm`, `pitchAccent` or `tags` is denied. Harmless today because every writer goes through `emptyDraft()`; misleading the moment someone adds an optional field, and rules deploy ahead of hosting.
- **`users/{uid}` cannot be deleted** — `request.resource` is null on a delete, so the `keys()` bound is a null dereference. The subcollections got the `read, delete` / `create, update` split; the parent did not.
- **`progress/{documentId}` has no bounds and no `ownerUid` check**, which is the section's own stated threat. The adapter only ever writes one fixed document id.
- **Public write paths were not part of the bounding.** `publicEntries`, `publicSets` and `publicSets/{id}/entries` (lines 222-253) still accept any size and any extra fields from an allowed owner. If "writes are bounded" is the claim, those are the other half of it — and they are the half every allowed user can read. Either bound them or say in the file why they are excluded.
- **`admin/` is not linted.** No `files` block in `eslint.config.js` matches it, so `yarn lint` skips the script that grants production access.
- **The deploy order lives only in this description.** The first production run of `deploy-prod.yml` is the migration to claim-gated rules; the two manual steps that keep it seamless need to survive the merge.

### Verification

Reviewed at `855e9bf` against `d406a07`, in a detached worktree.

Rules findings were run against the Firestore emulator rather than reasoned about — the missing-field denials, the `users/{uid}` delete denial, the unbounded `progress` writes and the 100 × 3,000-character `senses` document are all quoted from actual `PERMISSION_DENIED` output or actual accepted writes. `tsc -p tsconfig.scripts.json --noEmit` passes. The `grep -q ""` behaviour was checked directly. The revocation finding is the one exception and is marked as such inline: it is reasoned from documented Firebase behaviour, because `@firebase/rules-unit-testing` mints tokens directly and cannot exercise revocation.

Cross-checked with a second reviewer on a different model. It independently raised the revocation window and the unbounded public write paths, both of which I had missed; every finding it contributed was verified here before being included.

One process note: 677 lines is past the point where a reviewer can hold the whole thing at once. This one hangs together better than most changes that size — the rules, the script that feeds them and the workflow that deploys them are genuinely one unit — so this is an observation rather than a request.

firestore.rules
Upheld, and this is the one that mattered most: the claim was wrong, not imprecise, and it was wrong in three places that each reinforced the other two.

Reasoned through rather than measured, for the reason you gave — the harness mints tokens directly, so the window is structurally unreachable from it. Firestore rules expose `request.auth` from the verified ID token and nothing else; there is no revocation state to consult, and no rules equivalent of `checkRevoked`. So an already-issued token keeps its claims for its full life regardless of what the Admin SDK was told.

What I had backwards is worth naming: `revokeRefreshTokens` ends the *session* at the next refresh, which is a real thing and worth doing — it is just not the thing I claimed. Corrected in the rules, in `admin/allow-user.ts`, and in the line it prints after a revoke, which now says access ends within the hour.

Took your first option. The second is recorded in the rules as available if abuse becomes a live problem, because it reintroduces the per-request read this whole change removes — a trade worth making deliberately and not by default.

Fixed in 8748d4e

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Partially: the value is right, the silence about it was not.

One reCAPTCHA v3 site key covers both projects, and that is a property of v3 rather than a shortcut — a v3 "site" is a list of domains and belongs to no Firebase project. The App Check *registration* is per project and there are two, each holding the same secret. Enterprise keys are project-scoped and would need splitting, which is exactly the asymmetry you would expect to see and did not.

You are right that it reads as an oversight next to six `PROD_`-prefixed lines. It now carries a comment saying which of the two it is.

Fixed in 8748d4e

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Upheld and reproduced:

```console
$ printf 'hello\n' > f; grep -q "" f; echo $?
0
```

The path you traced is the part I would not have found: the `build` job has no `environment:`, so putting the production values where "production secrets" belongs hands it empty strings — and nothing downstream notices, because `client.ts` validates at runtime and `vite build` never evaluates the module.

Taken further than the one value, as you suggested: all six are required before anything is compared against them, with an error naming which are missing and why environment-scoped secrets are invisible here. The comparison step no longer has to trust its input.

Fixed in 8748d4e

firestore.rules
Upheld, and reproduced against the emulator before changing anything — a create without `definitionSub` is denied with `Property definitionSub is undefined on object`.

This is the worst kind of comment to have written: it does not merely fail to describe the code, it tells the next person a cost they will not pay. And the direction you point out makes it concrete — rules deploy before hosting, so the denial would be live before the client that fills the field.

`get` with a default throughout now. `headword` stays a direct read, since an entry without one is not an entry and the form refuses it too; there is a case asserting both.

Fixed in 8748d4e

firestore.rules
Agreed, and the phrasing was doing real work in the wrong direction — a reader arriving at "bound writes" will read a size bound.

Added next to the bounds: read it as cardinality. 100 senses of 3,000 characters each is accepted, and the only size limit on a document here is Firestore's own 1 MiB — a ceiling rather than a budget, since it says nothing about how many documents there are.

Fixed in 8748d4e

firestore.rules
Upheld and reproduced. Split as you wrote it.

The detail worth keeping is that this is the third time in this branch the same null has bitten: `keepsCreatedAt` was written specifically to guard it, the comment explaining why is directly above, and the parent document was still left with the combined form. Knowing about a trap is not the same as having checked every place it applies.

There is a case asserting the owner can delete their own user document now, so the next caller — account deletion — finds it covered.

Fixed in 8748d4e

firestore.rules
Upheld, and the framing is the part I will carry: this path was the threat the section header names, sitting underneath the header that names it.

Narrowed to `progress/entries`, which is the only id `progressRepo` addresses, with `ownedBy` alongside `mine()` — you are right that it was the one write path where the field went unchecked while every sibling checked it. Two cases: the wildcard id is refused, and so is a document claiming another owner.

Fixed in 8748d4e

firestore.rules
Verified at `8748d4e`. The correction landed in all three places, and the new text is accurate: "Revoking the refresh tokens does not shorten it. It stops the *next* refresh from succeeding, which ends the session — but the ID token already in hand keeps its claims and keeps being accepted until it runs out."

`admin/allow-user.ts` matches, including the console line — which is the one an operator actually reads at the moment they care, so it was the important one of the three.

Recording the `auth_time` alternative as an available decision rather than doing it is the right call at this size, and it now says what it would cost. Resolving.

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Accepted. The v3 reasoning is correct — a v3 site key is registered against a domain list in the reCAPTCHA console and is not owned by any Firebase project, so one key pair registered in both projects' App Check settings is a legitimate configuration rather than a shortcut. The comment at line 72 says so where the next reader will hit it, which is what was missing.

One operational precondition it does not state, and it is the thing that would break this quietly: the key's domain list has to cover both hosts. If production later moves to a custom domain and the list is not updated, attestation fails on production only, and the App Check metrics are exactly where that would show up. Worth a line in the runbook rather than another round here.

Resolving.

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Verified at `8748d4e`. The new "Require every production value" step checks all six and names the environment-scoping trap in its own comment, so the failure is self-explaining when it fires.

Two notes, neither reopening this:

- The error message builds its names as `PROD_VITE_FIREBASE_$name`, which is right for five of the six and wrong for `SENDER_ID` — the real secret is `PROD_VITE_FIREBASE_MESSAGING_SENDER_ID`. An operator following the message would create a secret that nothing reads. Left as a separate comment on line 108.
- The step sits after `Build` rather than before it. The deploy is blocked either way, so the blocker is closed; it just spends a build first.

Resolving.

firestore.rules
Verified at `8748d4e`, and verified in both directions rather than on the diff.

Green: the suite passes at 70 cases, up from 64. Red: reverting `str(d.get('definitionSub', ''), 20000)` to `str(d.definitionSub, 20000)` and re-running produces exactly one failure —

```
FAIL  what the comments in the rules claim > accepts an entry missing every optional field
  PERMISSION_DENIED: Property definitionSub is undefined on object. for 'create' @ L223
```

— one test, the right one, nothing else disturbed.

`d.get('pitchAccent', null) == null || d.get('pitchAccent', 0) is int` is correct through all four states: absent, null, int, and a wrong type, which still denies. Keeping `headword` and `name` read directly is the right exception, and "an entry without one is not an entry" is the reason a reader needs. Resolving.

firestore.rules
Verified. "Read 'bounded' as cardinality, not size" is the sentence that was missing, and pairing it with the 1 MiB ceiling — "a ceiling rather than a budget: it stops one document being enormous and says nothing about how many documents there are" — makes the remaining exposure legible instead of implied. Resolving.

firestore.rules
Verified at `8748d4e`, in both directions. Split as `allow read, delete` / `allow create, update`, matching the subcollections, and `lets the owner delete their own user document` covers it.

Reverting the split puts that test red and nothing else. The observation in your reply is the part worth keeping in the file: `keepsCreatedAt` existed specifically to guard this null, the explanation sat directly above it, and the parent document still had the combined form. Resolving.

firestore.rules
Verified at `8748d4e`, in both directions. `match /progress/entries` with `ownedBy(uid)` and a key bound, and two cases covering it — the wildcard path and the wrong-owner field.

Reverting to `match /progress/{documentId}` with the unbounded rule turns both red and nothing else:

```
FAIL  refuses a progress document other than the one the adapter writes
FAIL  refuses a progress document claiming another owner
```

Together with the `users/{uid}` revert that is three failures across the two changes, each landing on its own case. The `ownedBy` addition was not in the original comment and is the better half of the fix — it was the only write path in the file not checking the field. Resolving.

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,193 @@
...
+          AUTH_DOMAIN: ${{ secrets.PROD_VITE_FIREBASE_AUTH_DOMAIN }}
+          PROJECT_ID: ${{ secrets.PROD_VITE_FIREBASE_PROJECT_ID }}
+          STORAGE_BUCKET: ${{ secrets.PROD_VITE_FIREBASE_STORAGE_BUCKET }}
+          SENDER_ID: ${{ secrets.PROD_VITE_FIREBASE_MESSAGING_SENDER_ID }}
+          APP_ID: ${{ secrets.PROD_VITE_FIREBASE_APP_ID }}
+        run: |
+          missing=''
+          for name in API_KEY AUTH_DOMAIN PROJECT_ID STORAGE_BUCKET SENDER_ID APP_ID; do
+            eval "value=\$$name"
+            if [ -z "$value" ]; then missing="$missing PROD_VITE_FIREBASE_$name"; fi
```
kowinauyeung: **🔵 Five of these six names are right and one is not.**

`missing="$missing PROD_VITE_FIREBASE_$name"` reconstructs the secret name from the loop variable, which works for `API_KEY`, `AUTH_DOMAIN`, `PROJECT_ID`, `STORAGE_BUCKET` and `APP_ID`. For `SENDER_ID` it prints `PROD_VITE_FIREBASE_SENDER_ID`, while line 102 reads `PROD_VITE_FIREBASE_MESSAGING_SENDER_ID`.

So the one case this step exists for — somebody has put the values in the wrong place and is reading the error to find out which — sends them to create a secret that nothing reads, and the build fails again with the same message.

Listing the full names rather than deriving them removes the class of mistake:

```yaml
run: |
  missing=''
  for name in \
    PROD_VITE_FIREBASE_API_KEY \
    PROD_VITE_FIREBASE_AUTH_DOMAIN \
    PROD_VITE_FIREBASE_PROJECT_ID \
    PROD_VITE_FIREBASE_STORAGE_BUCKET \
    PROD_VITE_FIREBASE_MESSAGING_SENDER_ID \
    PROD_VITE_FIREBASE_APP_ID
  do
    eval "value=\$$name"
    [ -z "$value" ] && missing="$missing $name"
  done
```

…with the `env:` keys renamed to match, which also drops the mapping a reader currently has to hold in their head between lines 97-102 and line 106.

kowinauyeung:
## Re-review at `8748d4e`

**Both blockers are closed.** Seven threads verified and resolved, eight still open, one new minor finding on the fix itself.

### Verified

Every rules claim was checked against the emulator rather than read off the diff, and the red side was reproduced independently rather than taken from the commit message.

Green: the emulator suite passes at **70 cases, up from 64**.

Red, reverting each fix in turn:

| Reverted | Failures | Expected |
| --- | --- | --- |
| `str(d.get('definitionSub', ''), …)` → `str(d.definitionSub, …)` | `accepts an entry missing every optional field` | 1, matched |
| `users/{uid}` split → combined `allow write` | `lets the owner delete their own user document` | 1, matched |
| `progress/entries` → `progress/{documentId}` unbounded | `refuses a progress document other than the one the adapter writes`, `refuses a progress document claiming another owner` | 2, matched |

Three reverts, four failures, each landing on its own case and nothing else disturbed. The "verified red" claim holds.

The revocation correction is accurate in all three places, including the console line — which is the one an operator reads at the moment it matters, and so the one that was actually doing the damage. Keeping the `auth_time` alternative as a recorded decision rather than implementing it is right at this size, and it now states what it would cost.

The `get`-with-default rewrite is correct through every state I could think to check, including `pitchAccent` when absent, null, an int, and a wrong type. Keeping `headword` and `name` read directly is the right exception.

Two things in the fixes were better than what was asked for: `ownedBy(uid)` on `progress` was not in the original comment and closes the one write path in the file that was not checking the owner field, and the new `what the comments in the rules claim` block names the *pattern* — comments describing a property the file does not have — rather than the four cases, which is the thing worth guarding.

### New, on the fix

- `deploy-prod.yml:108` — the error message derives secret names from the loop variable, which is right for five of six and prints `PROD_VITE_FIREBASE_SENDER_ID` for the sixth, where the real name is `PROD_VITE_FIREBASE_MESSAGING_SENDER_ID`. Sends an operator to create a secret nothing reads, in the one situation the step exists for. Comment inline.
- Not worth a comment, but noted: the new step runs after `Build` rather than before it. The deploy is blocked either way; it just spends a build first.

### Still open — eight threads

Unaddressed by this commit and left unresolved. Nothing here is a blocker.

| Thread | Why it is still open |
| --- | --- |
| `admin/allow-user.ts:1` — `admin/` outside the ESLint scope | `eslint.config.js` untouched; CI still skips the file |
| `admin/allow-user.ts:3` — duplicate `firebase-admin/app` import | still two imports |
| `admin/allow-user.ts:29` — unknown args accepted silently | `rest.includes('prod')` unchanged |
| `deploy-prod.yml:185` — deploy order lives only in the description | no runbook or README change |
| `client.ts:70` — CSP comment describes a policy not on this branch | `client.ts` untouched |
| `firestore-rules.test.ts:404` — adapter block only covers creates | no update cases added |
| `.env.example:15` — "reCAPTCHA Enterprise" stale after 07d88d3 | untouched |
| `tsconfig.scripts.json:15` — `"module"` declared twice | untouched |

Two of those are worth a second look before merge rather than after, for reasons outside their own severity:

- **`client.ts:70`** is the same class of problem as #20's privacy policy naming reCAPTCHA before it exists — a statement about a sibling branch that cannot be checked from this one. The two pull requests now have a real ordering relationship in both directions and nothing enforces it.
- **`admin/` not being linted** is why the duplicate import two lines below it survived a green CI run. Fixing the scope is what stops the next one.

### Verification

`8748d4e` against `855e9bf` and `d406a07`, in a detached worktree. Emulator suite run at head and at three separate reverts. Every claimed fix read in the diff before being accepted; no thread resolved on the commit message alone.

admin/allow-user.ts
kowinauyeung: @kowinauyeung
Upheld, and wider than `admin/`. Reproduced first:

```console
$ npx eslint admin
ESLint: You are linting "admin", but all of the files matching the glob pattern "admin" are ignored.
$ echo $?
2
```

The part that makes it invisible is the one you named: `eslint .` does not error on a directory it matches nothing in, so the Lint step stayed green while skipping the file.

`migration/accents.ts` has the same hole — the config had `migration/**/*.mjs` only, and that folder was plain ESM until `accents.ts` arrived. So the new block covers `admin/**/*.ts` and `migration/**/*.ts`, which is exactly what `tsconfig.scripts.json` includes; scoping them together is what stops the lint pass and the typecheck pass disagreeing about what counts as a script.

One thing that did not work: `projectService: true`, which is what every other typed block here uses. It resolves the nearest `tsconfig.json`, and this repository's root one is a solution file with `"files": []`, so every script comes back as not found by the service. Typed through `project: "./tsconfig.scripts.json"` instead, with a comment saying why so the next person does not "fix" it back.

Red before green, on your exact example: restoring the duplicate import puts `eslint .` — the command CI runs, not `eslint admin` — at one error, `perfectionist/sort-imports`. Nothing else moved.

Fixed in `d32be26`

admin/allow-user.ts
kowinauyeung: @kowinauyeung
Fixed — `applicationDefault` joins line 1.

Worth recording that this is the comment above it in practice: the rule that would have caught it was already configured and simply never ran on this file.

Fixed in `d32be26`

admin/allow-user.ts
kowinauyeung: @kowinauyeung
Agreed, and the reasoning about *why* it is a natural mistake is the part that decided the fix rather than the severity: `--revoke` is the only flag spelled with dashes, so `--prod` is what a hand reaches for after typing it once.

Unknown arguments are now refused, and the unknown one is named:

```
unknown argument: --prod
usage: allow-user.ts <email> [prod] [--revoke]
```

Taking the cheaper option you offered rather than the safer-sounding one: rejecting is a filter and a length check, while noticing means the success line at the end has to be read after the grant has already happened.

Fixed in `d32be26`

.env.example
kowinauyeung: @kowinauyeung
Fixed, and expanded for the reason you gave — this file is the instruction someone follows while provisioning, so it is the one place the distinction has to survive.

It now says v3 *not* Enterprise, that the two are created in different consoles, and that Enterprise needs the billing account these projects do not have. Added one thing you did not ask for and I think belongs beside it: that one key covers both projects, because a v3 site is a list of domains and belongs to no Firebase project. That is the fact behind the unprefixed secret you queried on `deploy-prod.yml:80`, and `.env.example` is where someone will next wonder why it is not split.

Fixed in `d32be26`

tsconfig.scripts.json
kowinauyeung: @kowinauyeung
Fixed. The duplicate is gone and the comment now sits on the line that survives, which is the half that mattered — as written it introduced a setting that was already three lines above it.

Confirmed your reading that it was cosmetic: `tsc -p tsconfig.scripts.json --noEmit` was green before and after.

Fixed in `d32be26`

src/infra/firebase/client.ts
kowinauyeung: @kowinauyeung
Upheld on all three, and the framing changed what I did about it.

**"reCAPTCHA Enterprise" is stale** — corrected to v3, and both hosts are named now: the script comes from `https://www.gstatic.com/recaptcha/` *and* `https://www.google.com/recaptcha/`, and the frame from `https://www.google.com`. As you said, gstatic is already allowed; naming both makes the eventual edit a copy-paste.

**The claim about `firebase.json` is gone rather than corrected.** Confirmed there is no `headers` block on this branch at all. Anything asserted here would be true or false depending on which pull request merged first, so the comment now says only what is true of the client, and points at a check instead.

**The check that fails.** `tests/unit/csp.test.ts`, written as the conditional rather than as a fixed expectation: *if* an enforcing `Content-Security-Policy` is present, `script-src` and `frame-src` must allow `https://www.google.com`. The key match is exact, so `Content-Security-Policy-Report-Only` is the state it deliberately permits.

Verified in three states, since a guard that can only be checked in one of them is the thing you are warning about:

- no headers, which is this branch — green
- the #20 policy as Report-Only — green
- the same policy with `-Report-Only` dropped — **red**, `expected 'script-src 'self' https://apis.goog…' to contain 'https://www.google.com'`

A second case pins the premise, because the first passes vacuously with no policy and would go on passing if App Check were removed and the requirement quietly became wrong: it asserts `client.ts` still constructs a `ReCaptchaV3Provider`.

This does put the guard on #21 while the policy lives on #20 — the mirror image of the ordering problem on the other pull request. That is deliberate: it is green in either merge order and fires at promotion whichever way round they land.

Fixed in `d32be26`

tests/rules/firestore-rules.test.ts
kowinauyeung: @kowinauyeung
Upheld. The `resource == null` point is exactly right and is the reason a create suite cannot be evidence about `keepsCreatedAt`: the comparison is never reached.

Three cases, all against the emulator:

- `accepts the whole-draft update entryRepo.update sends over a stored entry` — `updateDoc` with the entire draft and a fresh `updatedAt`, without `createdAt` or `ownerUid`, so both arrive in `request.resource.data` only through the merge.
- the same for `wordSetRepo.update`.
- `refuses an update that re-sends a stored Timestamp createdAt as a string` — your `serverTimestamp()` versus stored `Timestamp` point, turned into the case that documents *why* the adapter omits the field. Round-tripping it through the domain type gives back an ISO string, which is not the same value, so an edit that changes nothing is refused.

`createdAt` is stored with a `Date` rather than an ISO string in these, for the reason you gave — the comparison is against whatever type is really in the document, and a string-to-string test would be the by-hand version that passes and proves nothing.

Red in two directions, and the second is the one that answers "is this just the create test again":

- `keepsCreatedAt` neutered to `return true` → exactly 2 red, both about `createdAt`, 71 green.
- `str(d.get('updatedAt', ''), 40)` added to `validEntry` and `validWordSet` → exactly the 2 new whole-draft updates red, **every create green**. That is what shows the update path presents a shape a create never does: a merged `Timestamp` where the create sends a string.

On the second half — the required-field behaviour is already pinned, in `8748d4e`, in the `what the comments in the rules claim` block: `accepts an entry missing every optional field`, `still refuses one with no headword at all`, and the word-set and session equivalents. That was the fix for your `firestore.rules:136` comment, and you are right that it needed a test naming it either way; it has one.

Suite is 73, up from 70.

Fixed in `d32be26`

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Upheld, and looking for somewhere to put the runbook is what turned this into the largest change in the round.

**README.md now has an "Operator runbook" section** covering granting, revoking and the three-step release, and the rules step in this workflow points at it — with the reason stated where the order is, not just the order: these rules require a claim no token issued before step 1 carries, so reversing steps 1 and 3 locks out everyone already signed in, for up to an hour, **including whoever is running the deploy**. Your observation that the first production run of this workflow *is* the migration is in there too, because it is the one run where every account is in that state at once.

The thing I did not expect, found while looking for the right place: the README was already describing the previous release model, and this pull request had not touched it.

- "**Production is not automated.** Nothing in `.github/workflows/` can reach `goitei`" — this pull request adds the workflow that does.
- "Access is gated on `allowedUsers/{uid}` existing" — replaced by the claim, in this same branch.

So on merge it would have told the next person to deploy production by hand and to grant access by editing a collection nothing reads any more. Both corrected. Not raised by you and squarely inside what the comment was about: the operational knowledge for this change was living somewhere that does not survive the merge, and the place it should have been living was actively wrong.

Fixed in `d32be26`

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Upheld. Reproduced the mismatch before changing it, with the fifth value emptied:

```console
$ sh t.sh
missing: PROD_VITE_FIREBASE_SENDER_ID
```

against `PROD_VITE_FIREBASE_MESSAGING_SENDER_ID` on line 102 — so the message named a secret nothing reads, in the one run the step exists for.

Taken as written: full names in the loop, `env:` keys renamed to match, and `missing="$missing $name"` rather than reconstructing a prefix. Same script after the change:

```console
$ sh t.sh
missing: PROD_VITE_FIREBASE_MESSAGING_SENDER_ID
```

The second half of your point is the better half and is why the shape changed rather than just the one string: there is no longer a mapping between the `env:` block and the loop for a reader to hold, so the class of mistake is gone rather than this instance of it.

Fixed in `d32be26`

kowinauyeung:
## Review round 2 — `d32be26`

Nine comments answered, all upheld, none disputed.

**First, the reason this is a second round.** Only one of the nine was new. The other eight were submitted in the same review as the seven answered in `8748d4e`, at the same second, every one of them anchored to a line in the diff. I replied to seven and reported the round complete. Nothing about those eight was hard to find; I stopped reading half way down a list.

That is the third time on these two pull requests that I have treated one look at the comments as the whole of them, so the fix is not "be more careful": before claiming a round is done, the count of threads I have replied to has to match the count of threads that exist.

| Comment | Change |
| --- | --- |
| `admin/allow-user.ts:1` | ESLint now covers `admin/**/*.ts` and `migration/**/*.ts` |
| `admin/allow-user.ts:3` | Duplicate `firebase-admin/app` import merged |
| `admin/allow-user.ts:24` | Unknown arguments refused, so `--prod` no longer grants on dev |
| `deploy-prod.yml:108` | Missing-secret error names the secret that is actually read |
| `deploy-prod.yml:148` | Operator runbook in README.md; the workflow points at it |
| `client.ts:70` | v3 not Enterprise; the `firebase.json` claim replaced by a test |
| `firestore-rules.test.ts:400` | Three update cases, including the Timestamp round-trip |
| `.env.example:15` | v3, and why the key is deliberately not split dev/prod |
| `tsconfig.scripts.json:15` | Duplicate `"module"` removed |

**Two things worth pulling out of the threads.**

`migration/accents.ts` had the same lint hole as `admin/`, which the comment did not mention. The config had `migration/**/*.mjs` and that folder was plain ESM until `accents.ts` arrived. Both folders are now linted, scoped to exactly what `tsconfig.scripts.json` includes.

The README was already describing the release model this branch replaces — "production is not automated", and access gated on an `allowedUsers` document. Neither has been true since `8352939`, in this same branch. Merged as it stood, it would have told the next person to deploy production by hand and to grant access by editing a collection nothing reads. Found while looking for somewhere to put the runbook.

**Verified red before green**, per the repository's rule:

- `keepsCreatedAt` → `return true`: exactly 2 red, both `createdAt` cases.
- `str(d.get('updatedAt', ''), 40)` added to `validEntry`/`validWordSet`: exactly the 2 new whole-draft updates red, every create green — which is what shows the update path presents a merged `Timestamp` that no create produces.
- Duplicate import restored: `eslint .` (the command CI runs) at one error.
- `firebase.json` policy promoted from Report-Only: the new unit test red; back to Report-Only, green.

518 unit, 73 emulator (up from 70), lint clean, three typecheck passes.

admin/allow-user.ts
kowinauyeung: @kowinauyeung
Verified at `d32be26`. `eslint admin` now lints the file rather than reporting it ignored, and `eslint .` stays at 0 errors. Scoping the block to the same two folders `tsconfig.scripts.json` includes is the right join — the lint pass and the typecheck pass can no longer disagree about what a script is — and the note about `project` over `projectService` is correct for a root solution file.

Checked that it catches the thing it was added for. Restoring the original duplicate import turns `eslint admin` red:

```
admin/allow-user.ts
  3:1  error  Expected "firebase-admin/app" to come before "firebase-admin/auth"  perfectionist/sort-imports
✖ 1 problem (1 error, 0 warnings)
```

**One thing worth knowing, because it is narrower than it looks:** what caught it is import *ordering*, not duplication. A duplicate that happens to be in sorted order goes through:

```ts
import { cert, getApps, initializeApp } from 'firebase-admin/app';
import { applicationDefault } from 'firebase-admin/app';   // same module, sorted
import { getAuth } from 'firebase-admin/auth';
```

```console
$ eslint admin; echo $?
0
```

So the original defect is caught by accident of where it sat. `no-duplicate-imports` in the new block would cover the class rather than the instance — one line, and this thread's actual ask (the scope) is fixed either way.

Resolving.

admin/allow-user.ts
kowinauyeung: @kowinauyeung
Verified — one import, and it is now inside the lint scope rather than only inside a review comment. Resolving.

admin/allow-user.ts
kowinauyeung: @kowinauyeung
Verified. `unknown` is computed before anything acts on `rest`, and the rejection names the offending argument as well as the usage line.

Checked the cases: `--prod` is rejected; a bare `yarn allow` still prints usage; `--revoke` and `prod` in either order still work. Passing the flag first (`yarn allow --revoke you@example.com`) reports `unknown argument: you@example.com`, which is a slightly confusing message for an unambiguous mistake — but it refuses, which is the whole point of the change.

The comment explains *why* the mistake is natural rather than just guarding it, which is the part that will stop somebody "simplifying" this later. Resolving.

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Verified. The README now carries an Operator runbook with the three steps in order, and this step points at it from the workflow.

The two things that make it worth having rather than decorative: the ordering paragraph explains *why* it is not a preference — the rules being deployed require a claim no pre-existing token carries — and it names the case that actually bites, that the first production run of this workflow **is** the migration onto the claim. The revocation entry carries the corrected within-the-hour wording too, so the runbook and the code now say the same thing.

Also right to have removed the by-hand production commands from the README in the same edit. Leaving them next to "production is a button" is how a runbook drifts back into having two procedures. Resolving.

src/infra/firebase/client.ts
kowinauyeung: @kowinauyeung
Verified at `d32be26`, and this is the better answer to what I raised — I asked for the claim to be made checkable; you made it enforced, and dropped the cross-branch assertion instead of trying to keep it true.

`tests/unit/csp.test.ts` reads `firebase.json` and requires the reCAPTCHA host in `script-src` and `frame-src` **only if** an exactly-named `Content-Security-Policy` is present. Ran it through all three states rather than trusting the description:

| `firebase.json` state | Result |
| --- | --- |
| no `headers` block (this branch today) | passes |
| `Content-Security-Policy-Report-Only` only | passes |
| promoted to `Content-Security-Policy`, no `www.google.com` | **fails** — `expected 'script-src \'self\' https://apis.goog…' to contain 'https://www.google.com'` |
| promoted, hosts added to both directives | passes |

The exact-match filter is load-bearing and the comment says so, since `-Report-Only` starts with the same string. The second case — asserting `ReCaptchaV3Provider` still exists — is what stops the first one passing vacuously for the wrong reason, and that is the failure mode a conditional guard usually has.

One thing to confirm at promotion rather than now: the guard covers `script-src` and `frame-src`, not `connect-src`. v3 does its exchange from inside the frame, so I would not expect the parent policy to matter — but "I would not expect" is the wrong basis for it, and the documented procedure already says to promote with the console open against a real sign-in. That is where it would show.

Resolving.

tests/rules/firestore-rules.test.ts
kowinauyeung: @kowinauyeung
Verified at `d32be26` — the emulator suite is at 73 cases, up from 70, and both update paths are covered against a stored document.

The third case is the one I did not think to ask for and is the most valuable of the three: an update that re-sends `createdAt` as an ISO string is refused, because the stored value is a `Timestamp` and a client round-tripping it through the domain type sends back something that is not equal. That documents *why* `entryRepo.update` omits the field, which was previously only a comment in the adapter — and it is exactly the "edit that looks like it changes nothing" a user would report as a mystery.

Using `new Date(...)` rather than an ISO string in the fixtures is right for the same reason: the comparison is against whatever type is really in the document. Resolving.

.env.example
kowinauyeung: @kowinauyeung
Verified — v3 rather than Enterprise, with the reason (different console, Enterprise needs billing) and the note that one key deliberately covers both projects. That last line is the one that stops somebody "fixing" the unprefixed secret in `deploy-prod.yml`, so it is in the right file. Resolving.

tsconfig.scripts.json
kowinauyeung: @kowinauyeung
Verified — one declaration, with the comment now sitting on the line it describes. `tsc -p tsconfig.scripts.json --noEmit` still clean. Resolving.

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Verified at `d32be26`. Full names in the `env:` block and in the loop, so nothing is reconstructed, and the mapping a reader had to hold between the two lists is gone as well.

`eval "value=\$$name"` still does the indirection, which is fine here — the names are literals in the same file rather than anything from input. Resolving.

kowinauyeung:
## Re-review at `d32be26`

**All threads on this pull request are now resolved.** Sixteen in total: two blockers, then thirteen suggestions and nitpicks, then the one I opened on the first round of fixes.

### Suites, at head

| Suite | Result |
| --- | --- |
| `yarn test:unit` | 518 passed, 29 files |
| `yarn test:emulator` | 73 passed (was 70) |
| `tsc -b`, `tsc -p tsconfig.scripts.json` | clean |
| `eslint .` | 0 errors — and now actually reaching `admin/` |

### Verified by breaking it

**The CSP guard behaves in all four states**, which is the whole reason it is worth having. I edited `firebase.json` to simulate each rather than reading the description:

| State | Result |
| --- | --- |
| no `headers` block (this branch today) | passes |
| `Content-Security-Policy-Report-Only` only | passes |
| promoted, without `www.google.com` | **fails**, naming the directive |
| promoted, hosts in `script-src` and `frame-src` | passes |

This is a better answer than the one I asked for. I wanted the cross-branch claim made checkable; you deleted the claim and made the requirement enforced. The second assertion — that `ReCaptchaV3Provider` still exists — is what stops the conditional passing vacuously for the wrong reason, which is the usual failure of a guard shaped like this.

**The lint scope now reaches the file it was added for.** Restoring the original duplicate import turns `eslint admin` red. One caveat, recorded on the thread and narrower than it looks: what catches it is `perfectionist/sort-imports`, so a duplicate import that happens to be *in sorted order* still passes at 0 errors. The scope — this thread's actual subject — is fixed; covering the class rather than the instance would take `no-duplicate-imports` in the new block.

**The three update-path rules cases** are the ones the create-only block could not reach. The third is the one I did not think to ask for: an update re-sending `createdAt` as an ISO string is refused, because the stored value is a `Timestamp`. That turns a comment in the adapter into a fact the suite holds, and it is exactly the "edit that appears to change nothing" a user would report as a mystery.

### One thing to confirm later, not now

`csp.test.ts` guards `script-src` and `frame-src`, not `connect-src`. reCAPTCHA v3 does its exchange from inside its own frame, so I would not expect the parent policy to matter — but "would not expect" is the wrong basis, and the promotion procedure already says to do it against a real sign-in with the console open. That is where it would surface.

### Where this leaves the two pull requests

Both blockers on each are closed and every thread here is resolved. #20 has one open thread — the clipboard fix handles one of the two failure modes its comment claims — and is otherwise done.

The ordering relationship between the two is now held by three mechanical guards rather than by anybody's memory, which is the part of this round I would keep: #20's content test goes red if App Check lands without the disclosure, this branch's `csp.test.ts` goes red if the policy is promoted without the reCAPTCHA hosts, and `DEPLOY_SHA` is set on one branch and consumed on the other. None of those depends on reading a pull request description.

Issue #14, opened 2026-08-09
parse.mjs has no regression test, and cannot have one


kowinauyeung:
Split out of #2, where the last item under "Add beyond that" read:

> **Migration golden file** — `migration/output.json` is committed and is exactly
> that snapshot; if `parse.mjs` is ever touched, diff against it.

That test cannot be written. `parse.mjs` reads `bak/**/*.md`, and those notes are
not in this repository — they were removed once the import was verified, and the
corrections made during migration were never committed. There is no input to run
the parser against, so there is nothing to diff its output with.

This is not a gap to close later. It is a permanent property of the repository,
and this issue exists so that the next person to open `parse.mjs` finds it
already answered rather than spending an afternoon on it.

## What that means in practice

`migration/output.json` is not a *snapshot of* `parse.mjs`'s behaviour. It is the
**artefact of record** — the only complete description of what was loaded into
Firestore, and the only way to replay production from the repository alone.
`migration/README.md` already says as much.

So:

- **`parse.mjs` and `normalize.mjs` are provenance, not live code.** They
  document how each field was derived. Changing them produces nothing, because
  there is no input to feed them.
- **`output.json` is never regenerated.** If it has to change, it is edited or
  rewritten deliberately, and the edit is the thing under review.
- **Nothing runs at deploy time.** `upload.mjs` is the only script still meant to
  be executed, and it reads `output.json` directly.

## What is tested instead

`tests/unit/migrationOutput.test.ts` (added in #13) checks the artefact rather
than the parser that produced it:

- 67 entries, unique ids, both required fields present on every one
- every `jlpt` / `origin` / `style` / `politeness` / `pos` inside the current
  enums, and every `freq` on the 1–5 scale
- every `learnedOn` a real calendar day — the `2026-02-31` class of defect,
  checked against the data rather than a literal
- `wordSets` empty everywhere, which is what makes `upload.mjs` dropping the
  field provably lossless
- and the end-to-end one: mapping each entry exactly as `upload.mjs` does and
  reading it back through `sanitizeEntry` loses no content

That is a pre-flight check on an upload that has not run yet, which the golden
diff would not have been.

## Done when

Nothing to implement. Close this once `migration/README.md` states plainly that
`parse.mjs` is unrunnable and `output.json` is authoritative rather than
derived — the README currently notes `bak/` is absent in a table row, which is
easy to read past.

kowinauyeung:
[...]

Issue #2, opened 2026-08-06
Add vitest suite: port ad-hoc verification into durable coverage


kowinauyeung:
Phase 1 shipped with no committed test suite. This was a deliberate scheduling
decision — one coherent plan once the practice features exist, rather than
assertions written against a shape that was still moving. Two Phase 1 renames
(tag matching AND → OR, and `definitionJa` → `definition`) would each have
invalidated tests written before them.

That tradeoff has now cost something concrete, which is why this issue exists
rather than another deferral. Ad-hoc checks written during a review round were
run from throwaway esbuild bundles, so they were never re-run afterwards — and a
second round found an invalid-calendar-date bug (`2026-02-31` was accepted,
because `Date` rolls it forward to `2026-03-03` instead of returning `NaN`) that
durable coverage would have caught.

## Scope

Add **vitest**. It reuses the Vite config, so the `@/*` alias and TypeScript
resolve with no extra setup.

### Port first — hostile input (58 checks already written)

These exist today only as throwaway scripts. They cover two bugs that reached
review, both reachable by hand-editing a URL or pasting JSON.

| Module | Coverage |
|---|---|
| `lib/filters.ts` | `freq` of `-1` / `Infinity` / `1e9` / `abc` / `2.7`; allowlists for `sort`, `pos`, `origin`, `style`, `jlpt`; URL round-trip |
| `lib/sanitize.ts` | nulls inside `senses` / `examples`; invalid enums; out-of-range `freq`; non-object `context` / `usage`; a simulated legacy document |
| `lib/sanitize.ts` (dates) | month-rollover dates (`2026-02-31`, `2025-02-29`), out-of-range values, shape failures, leap days |

### Port next — behaviour verified ad hoc during the build (~53 checks)

| Area | Coverage |
|---|---|
| Filters & search | substring match across all fields, each filter dimension, AND across dimensions, OR within one, sort orders, no mutation |
| Furigana | segmentation for all 40 entries with readings, okurigana anchoring, fallback when alignment fails, lossless reconstruction |
| JSON import | minimal / partial / full objects, defaults, code fences, malformed input, tag parsing, user-typed context winning over the assistant's, prompt generation |

### Add beyond that

- **Firestore rules** via `@firebase/rules-unit-testing` — verify a wrong account
  is denied on every collection. **The only security-relevant test here; do this
  one first if the list gets cut.**
- **`Ruby` component** — a layout class on `<ruby>` drops it out of its ruby
[...]

PR #22, opened 2026-08-13
fix(errors): report a denied read as denied, not as a failed load

kowinauyeung:
Every screen reported "単語を読み込めませんでした" and its three siblings while the
actual failure was `permission-denied`. The two are indistinguishable to a reader,
and they need opposite responses: one clears itself, the other never does until the
reader signs out and back in.

### What happened

`deploy-dev.yml` deployed the rules from #21 to `goitei-dev`. Those rules gate every
read under `users/{uid}` on the `allowed` custom claim, and no account carried it —
the `allowedUsers` document that used to grant access is read by nothing now. So
entries, word sets, practice sessions and progress were all denied at once, and all
four providers caught the same rejection and reported it as a load failure.

That is the lockout the Operator runbook in `README.md` predicts for a production
release. Dev deploys itself on merge, so dev reached it first — and the app said
nothing that would point anyone at it.

### The change

`src/lib/loadError.ts` maps `permission-denied` and `unauthenticated` to a message
that names the situation and the fix. Both codes share one message because both are
facts about the account rather than about the data, and both are cleared by the same
action — a claim reaches a client only in a freshly minted ID token, so an account
granted access mid-session stays denied until it signs in again.

The four subject-specific fallbacks are untouched. A dropped connection really is
about the thing being read, and telling that reader to sign in again sends them
nowhere.

Duck-typed on `code` rather than `instanceof FirebaseError`: `src/lib` sits above the
infrastructure seam, and an SDK import here would put a Firebase type into every
caller for what is one string comparison.

### This does not restore access

It makes the failure legible. Granting access is still `yarn allow <email>` followed
by signing out and back in.

### Tests

**Layer: `tests/unit`.** The classifier is a function of its inputs, which is the
cheapest layer that can see the defect. It takes no clock, no DOM and no emulator.

Red before green, and red in the right place. Reverting `loadErrorMessage` to the
shape it replaced — returning the fallback and ignoring the cause — fails **exactly
two** cases:

```
× tells a denied account what to do instead of blaming the connection
× treats an expired token the same way, since re-signing in is the same fix
```

The other six stay green, which is what they are there for: they assert behaviour the
old code already had, so a change that broke them would be this fix trading one
indiscriminate message for another.

568 unit and component. Lint clean, three typecheck passes. Emulator and end-to-end
not run locally — this touches no rule and no adapter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,16 +79,25 @@ Pointing this at a fresh Firebase project takes a few more steps, in order:
 4. `yarn firebase login`, then deploy the rules with `yarn rules:dev` (once the
    alias points at your project) or
    `yarn firebase deploy --only firestore:rules --project <your-project-id>`.
-5. Sign in once, then add yourself to the allowlist — the rules deny everything
-   until `allowedUsers/{your-uid}` exists. `yarn migrate:upload` writes it for
-   you, or create the document by hand in the console.
+5. Sign in once, then grant yourself access — the rules deny everything until
+   your account carries the `allowed` custom claim. `yarn allow <your-email>`
+   sets it, and the account must have signed in before that, because the claim
+   goes on the uid Google issued and there is nothing to set it on until then.
+6. **Sign out and sign back in.** A claim reaches the client only in a freshly
+   minted ID token, so the session you granted access to is still denied — for
+   up to an hour, until its token expires on its own.
 
 `firebase-tools` is a local devDependency, so the CLI is reached through
 `yarn firebase …` unless you have installed it globally.
 
 Step 3 only changes local configuration; nothing is enforced until step 4
-deploys the rules successfully. After step 5 every read and write is scoped to
+deploys the rules successfully. After step 6 every read and write is scoped to
 your account.
+
+Step 6 is a step rather than a note because leaving it out looks exactly like a
+broken build: every screen reports that it could not load, and running step 5
+again changes nothing. `yarn allow` prints the account it granted, so the last
+line of step 5 is not evidence that step 6 has happened.
 
 `.env.development` and `.env.production` are gitignored. Their values ship inside
 the JS bundle and are not secrets — Firestore rules are what enforce access — but
--- a/README.md
+++ b/README.md
@@ -83,6 +83,11 @@ Pointing this at a fresh Firebase project takes a few more steps, in order:
    your account carries the `allowed` custom claim. `yarn allow <your-email>`
    sets it, and the account must have signed in before that, because the claim
    goes on the uid Google issued and there is nothing to set it on until then.
+   **`yarn allow` is the one step here that cannot be pointed at your project.**
+   `admin/allow-user.ts` names `goitei-dev` and `goitei` directly, never reads
+   `.firebaserc`, and rejects any argument other than `prod` and `--revoke` — so
+   on a fresh project it will tell you the account has never signed in. Set the
+   claim through the Admin SDK yourself; that script is the worked example.
 6. **Sign out and sign back in.** A claim reaches the client only in a freshly
    minted ID token, so the session you granted access to is still denied — for
    up to an hour, until its token expires on its own.
--- a/src/lib/loadError.ts
+++ b/src/lib/loadError.ts
@@ -23,14 +23,41 @@
 
 /**
  * `permission-denied` is the rules refusing the request; `unauthenticated` is
- * the token being absent or expired. Both mean the same thing to a reader —
- * this account cannot see this right now — and both are fixed by the same
- * action, so they share a message.
+ * the token being absent or expired. They share a message because a reader
+ * cannot tell them apart and neither can this module.
  */
 const DENIED_CODES = new Set(['permission-denied', 'unauthenticated']);
 
+/**
+ * Two sentences, and the second one is why there is a third.
+ *
+ * `permission-denied` is not one situation. It is at least three, and only one
+ * of them is cleared by signing in again:
+ *
+ *  1. **A token minted before the grant.** Re-authenticating fixes it, which is
+ *     what the first instruction is for.
+ *  2. **No claim on the account at all** — nobody has run `yarn allow`, or rules
+ *     requiring the claim were deployed to a project where nobody carries it.
+ *     Signing out and back in mints another token with no claim, so the reader
+ *     follows the instruction, lands on the same screen, and is out of moves.
+ *  3. **App Check rejecting the request.** `src/infra/firebase/client.ts`
+ *     initialises it whenever a site key is present, and an unregistered domain
+ *     or a dead reCAPTCHA key surfaces here as `permission-denied` like the
+ *     other two. Re-authenticating does nothing at all.
+ *
+ * Three messages would be worse, not better: the reader has no way to tell which
+ * one they are in, and the note below on subject-specific wording applies just
+ * as much to splitting one cause four ways. What the message must not do is
+ * dead-end — so it names サポート, which is where 2 and 3 are actually resolved.
+ *
+ * That exit has to be reachable from where this renders, and it is: `/support`
+ * is a top-level route outside the auth gate, and `Account.tsx` links it from
+ * inside the app, which a signed-in-but-denied account can still open — the gate
+ * that failed is Firestore's, not Firebase Auth's.
+ */
 export const ACCESS_DENIED_MESSAGE =
-  'アクセスが許可されていません。一度サインアウトして、サインインし直してください。';
+  'アクセスが許可されていません。一度サインアウトして、サインインし直してください。' +
+  '解決しない場合は、サポートページからお問い合わせください。';
 
 export function isAccessDenied(cause: unknown): boolean {
   if (typeof cause !== 'object' || cause === null) return false;
--- a/src/lib/loadError.ts
+++ b/src/lib/loadError.ts
@@ -1,0 +1,49 @@
+/**
+ * What to show when a read fails.
+ *
+ * Every provider used to catch everything and report the same sentence — "単語を
+ * 読み込めませんでした" and its three siblings — so a signed-in account with no
+ * access rendered exactly as a dropped connection. That is the one failure the
+ * generic wording is worst for: it is not transient, retrying never clears it,
+ * and the thing that does clear it (sign out, sign back in) is not something a
+ * reader would guess from "could not load".
+ *
+ * It is also the failure this application produces most easily. The gate is a
+ * custom claim, which reaches a client only in a freshly minted ID token, so an
+ * account granted access while signed in stays denied for up to an hour and
+ * every screen says the words could not be loaded. Deploying rules that require
+ * the claim to a project where nobody has it yet does the same to everyone at
+ * once.
+ *
+ * Deliberately duck-typed rather than an `instanceof FirebaseError`: `src/lib`
+ * is above the infrastructure seam and importing the SDK here would put a
+ * Firebase type in every caller's signature, for a check that is one string
+ * comparison. The codes are Firestore's own and are part of its API.
+ */
+
+/**
+ * `permission-denied` is the rules refusing the request; `unauthenticated` is
+ * the token being absent or expired. Both mean the same thing to a reader —
+ * this account cannot see this right now — and both are fixed by the same
+ * action, so they share a message.
+ */
+const DENIED_CODES = new Set(['permission-denied', 'unauthenticated']);
+
+export const ACCESS_DENIED_MESSAGE =
+  'アクセスが許可されていません。一度サインアウトして、サインインし直してください。';
+
+export function isAccessDenied(cause: unknown): boolean {
+  if (typeof cause !== 'object' || cause === null) return false;
+  const code: unknown = (cause as { code?: unknown }).code;
+  return typeof code === 'string' && DENIED_CODES.has(code);
+}
+
+/**
+ * `fallback` stays subject-specific — 単語, 単語集, 練習の記録, 練習履歴 — because
+ * a transient failure really is about the thing being read. Denial is not: it is
+ * a fact about the account, identical on every screen, and saying it four
+ * different ways would suggest four different problems.
+ */
+export function loadErrorMessage(cause: unknown, fallback: string): string {
+  return isAccessDenied(cause) ? ACCESS_DENIED_MESSAGE : fallback;
+}
--- a/src/lib/entries.tsx
+++ b/src/lib/entries.tsx
@@ -11,6 +11,7 @@ import type { ReactNode } from 'react';
 import type { Entry } from '@/domain/entry';
 import type { EntryRepository } from '@/domain/ports';
 import { entryRepositoryFor } from '@/lib/backend';
+import { loadErrorMessage } from '@/lib/loadError';
 
 interface EntriesValue {
   entries: Entry[];
@@ -103,7 +104,7 @@ export function EntriesProvider({ uid, children }: { uid: string; children: Reac
       if (walk.current === mine) setEntries(all);
     } catch (cause) {
       console.error(cause);
-      if (walk.current === mine) setError('単語を読み込めませんでした。');
+      if (walk.current === mine) setError(loadErrorMessage(cause, '単語を読み込めませんでした。'));
     } finally {
       if (walk.current === mine) setLoading(false);
     }
--- a/src/lib/progress.tsx
+++ b/src/lib/progress.tsx
@@ -3,6 +3,7 @@ import type { ReactNode } from 'react';
 import type { ProgressRepository } from '@/domain/ports';
 import type { EntryProgress, PracticeSessionDraft } from '@/domain/practice';
 import { progressRepositoryFor } from '@/lib/backend';
+import { loadErrorMessage } from '@/lib/loadError';
 import { weakIdsOf } from '@/lib/practice';
 
 interface ProgressValue {
@@ -57,7 +58,7 @@ export function ProgressProvider({ uid, children }: { uid: string; children: Rea
       })
       .catch((cause: unknown) => {
         console.error(cause);
-        if (!cancelled) setError('練習の記録を読み込めませんでした。');
+        if (!cancelled) setError(loadErrorMessage(cause, '練習の記録を読み込めませんでした。'));
       })
       .finally(() => {
         if (!cancelled) setLoading(false);
--- a/src/lib/wordSets.tsx
+++ b/src/lib/wordSets.tsx
@@ -11,6 +11,7 @@ import type { ReactNode } from 'react';
 import type { WordSetRepository } from '@/domain/ports';
 import type { WordSet } from '@/domain/wordSet';
 import { wordSetRepositoryFor } from '@/lib/backend';
+import { loadErrorMessage } from '@/lib/loadError';
 
 interface WordSetsValue {
   sets: WordSet[];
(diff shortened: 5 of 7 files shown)


src/lib/loadError.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * `permission-denied` is the rules refusing the request; `unauthenticated` is
+ * the token being absent or expired. Both mean the same thing to a reader —
+ * this account cannot see this right now — and both are fixed by the same
+ * action, so they share a message.
+ */
+const DENIED_CODES = new Set(['permission-denied', 'unauthenticated']);
+
+export const ACCESS_DENIED_MESSAGE =
+  'アクセスが許可されていません。一度サインアウトして、サインインし直してください。';
```
kowinauyeung: 🟡 **SUGGESTION** — the message prescribes a fix that does not apply to this PR's own headline scenario.

> アクセスが許可されていません。一度サインアウトして、サインインし直してください。

The description says dev broke because **no account carried the claim at all**. In that state, signing out and back in never clears it — the reader follows the instruction, lands on the same screen, and has no next move. The advice is only correct for the narrower case the comment above describes: a token minted before a grant that already exists.

There is a second cause with the same shape. App Check is initialised in `src/infra/firebase/client.ts`, and a rejected App Check token also surfaces as `permission-denied`. An unregistered domain or an expired reCAPTCHA site key would print "sign out and sign back in" too, and that loops just as tightly.

So `permission-denied` is not one situation — it is at least three, and only one of them is fixed by re-authenticating. That does not argue for three messages (the reader cannot tell them apart, and the module comment is right that four subject-specific denials would suggest four problems). It argues for one message that does not dead-end when the prescribed action fails. Something closing on the effect of 解除されない場合は管理者にお問い合わせください gives the other two causes an exit while leaving the stale-token case exactly as actionable as it is now.

This is the one item worth settling before merge, because it is a single string and the rest of the change is already correct.

kowinauyeung:
## Review — reporting a denied read as denied

**Verdict: APPROVE.** No blockers. The change is correct, it is tested at the right layer, and I confirmed the mechanism it depends on rather than taking the description's word for it.

### What I verified myself

| Check | Result |
| --- | --- |
| `yarn test:unit` | 568 passed across 35 files — matches the claim |
| `yarn typecheck` | all three passes clean |
| Red before green | reverted `loadErrorMessage` to `return fallback`, re-ran: **exactly two** failures, the two named in the description, 566 green. Restored, tree clean |
| **Does `code` survive to the catch block?** | Yes. Nothing in `src/infra/firebase/` wraps a thrown error — the only `catch` there is around the `Timestamp` constructor in `cursor.ts` and is unrelated. The `FirestoreError` reaches the provider intact |

That last row is the one that decides whether this PR does anything at all. An adapter that re-threw a plain `Error` would have left `isAccessDenied` permanently false and every screen saying exactly what it says today. It does not, so the fix works end to end.

The red-before-green run is worth calling out specifically. Reverting only the `return` in `loadErrorMessage` leaves `isAccessDenied` intact, so the two tests that went red are the two that assert the new behaviour, and the six asserting the behaviour the old code already had stayed green. That is the distinction `CLAUDE.md` asks for — reproducing the defect rather than something next to it — and it is visible in which tests failed.

### Suggestions (none blocking)

**1. `src/routes/Practice.tsx:224` discards the new message.** `progressError` is collapsed to a boolean before it reaches the setup panel:

```ts
weakState={progressError ? 'error' : progressLoading ? 'loading' : 'ready'}
```

`PracticeSetup.tsx:237` then renders a hard-coded `'記録を読み込めませんでした'`. The message `progress.tsx` now computes cannot get through that seam. The practical window is narrow — in a real lockout `entries` fails at the same moment and `Practice.tsx:212` returns the access-denied message first — but this PR leaves four of five error surfaces able to distinguish denial and one structurally unable to.

**2. Writes are untouched, and they fail the same way.** The lockout denies writes too: `EntryFormModal.tsx:92` reports `'保存できませんでした。'` with the same ambiguity this PR just removed from reads. Separately, `Account.tsx:60` does `setError(cause instanceof Error ? cause.message : …)`, which puts the SDK's English `"Missing or insufficient permissions."` straight into a Japanese UI. Deferring both is reasonable scoping; they are worth a follow-up rather than silence.

**3. `README.md:82-84` still points at the mechanism that was replaced.** Setup step 5 says the rules deny everything until `allowedUsers/{your-uid}` exists and suggests `yarn migrate:upload` writes it. But `firestore.rules:95` gates on `request.auth.token.allowed`, and the comment at `firestore.rules:61` already notes that `allowedUsers` has no match. A new operator following the README walks straight into the lockout this PR exists to make legible, and the remedy the README hands them does nothing. Cheapest possible fix, and this is the natural PR for it.

One inline comment on `src/lib/loadError.ts` covers the remaining point — the message prescribes an action that dead-ends for two of the three causes of `permission-denied`. It is a single string, so it is the only item I would settle before merge.

### What is done well

- **The layer choice is right and stated.** A classifier that is a function of its inputs, tested where that is cheapest to see — no clock, no DOM, no emulator.
- **The duck-typing rationale holds.** `src/lib` sits above the infrastructure seam; importing `FirebaseError` for one string comparison would put a Firebase type into every caller's signature.
- **`it.each([null, undefined, 'permission-denied', 42, …])` is well chosen.** The bare string case is the sharp one: this function is called from inside a `catch`, where reading a property off a non-object would replace a load failure with a blank screen. The test name says exactly that.
- **The subject-specific fallbacks are preserved deliberately**, with the reasoning written down. A transient failure really is about the thing being read, and this change does not trade one indiscriminate message for another.

No acceptance criteria were supplied, so there is no coverage table. Emulator and end-to-end suites were not run — consistent with the description, since this touches no rule and no adapter.

src/lib/loadError.ts
kowinauyeung: @kowinauyeung
Upheld. The message contradicted this pull request's own description: it says dev broke because no account carried the claim at all, and in that state signing out and back in mints another token with no claim. The one scenario the change was written for is the one its advice cannot fix.

All three claims checked before agreeing:

- **No claim anywhere.** The headline scenario, as above.
- **App Check.** `src/infra/firebase/client.ts:91` initialises it whenever a site key is present, and one is present in both `.env.development` and `deploy-prod.yml`. A rejected token surfaces as `permission-denied` alongside the other two.
- **Only the stale-token case re-authenticates away.** Which is the case the module comment describes, and the only one the sentence was written against.

Agreed too that this is not an argument for three messages, for the reason you give — the reader cannot tell which one they are in.

**Taken with one change: サポートページ rather than 管理者.** There is deliberately no administrator address in this project. `src/content/site.ts` sets out why: an address on three public pages is a permanent, unrevocable disclosure, so every contact goes through the form. "Ask an administrator" without saying where is the same dead end one step later.

The exit is reachable from where this renders, which I checked rather than assumed:

- `/support` is a top-level route (`src/router.tsx:44`), outside the auth gate.
- `src/routes/Account.tsx:163` links it from inside the app, and a signed-in-but-denied account can open that page — the gate that failed is Firestore's, not Firebase Auth's.

> アクセスが許可されていません。一度サインアウトして、サインインし直してください。解決しない場合は、サポートページからお問い合わせください。

The module comment now enumerates the three causes and records why there is one message rather than three.

**One test added, and the reason it is a separate one is worth stating.** The two existing cases assert `toBe(ACCESS_DENIED_MESSAGE)` — they compare against the constant, so no wording change can ever fail them. The regression you found was invisible to the suite as written.

Red before green, and red in one place only: restoring the two-sentence message fails exactly

```
× loadErrorMessage > offers a way out for the denials that signing in again cannot clear
```

and leaves the other eight green.

569 unit and component. Lint clean, three typecheck passes.

Fixed in 785d594

kowinauyeung: Verified in `785d594`, and I checked the reachability claim rather than accepting it — it is the part the new sentence stakes everything on.

- **`/support` is outside the gate.** `src/router.tsx:44` puts it as a sibling of the `/` `AppLayout` route, alongside `/login` and the three documents, not among its `children`. A signed-in-but-Firestore-denied account is still authenticated, so `AppLayout` never blocks it either way.
- **The link exists from inside the app.** `src/routes/Account.tsx:163`. It is also in the shared footer — `src/components/PublicLayout.tsx:11` — which `AppLayout` renders, so on a desktop the exit is on the same screen as the message.
- **App Check is genuinely live, so cause 3 is real.** A site key is present in `.env.development`, and `deploy-prod.yml:68` fails the production build outright when `VITE_RECAPTCHA_SITE_KEY` is unset. Both environments attest, so a rejected token really can surface here.

Red before green reproduced independently: restoring the two-sentence message fails exactly one test — `offers a way out for the denials that signing in again cannot clear` — with 568 green. 569 passing on the branch, three typecheck passes clean. Working tree restored after.

On the new test being separate: agreed, and it is not the tautology it resembles. The two existing cases assert `toBe(ACCESS_DENIED_MESSAGE)`, so they compare the constant against itself and no wording change can ever fail them. The gap was real and the new case is the only thing standing in it.

One thing worth recording, because it means the promise cannot rot silently: `tests/e2e/auth.spec.ts:62` already asserts `/support` serves a visitor with no account. If the route were removed, that goes red — so the message cannot end up pointing at a page that no longer exists without something failing first.

A note rather than a request: `AppLayout.tsx:132` wraps the footer in `hidden nav:block`, so on a phone the footer link is not on screen and the path is the avatar menu → `/account` → the link at `Account.tsx:163`. `AvatarMenu` renders unconditionally, so it is reachable, just one hop longer than on a desktop. Not worth changing for this.

Resolving.

kowinauyeung:
@kowinauyeung
Suggestion 3 taken, in `7063454`. Suggestions 1 and 2 deferred, with reasons.

**Setup step 5 was telling a new operator to do the thing that does not work.** It said the rules deny everything until `allowedUsers/{your-uid}` exists and that `yarn migrate:upload` writes it. `firestore.rules:95` gates on `request.auth.token.allowed`, and `firestore.rules:61` already records that `allowedUsers` has no match block — so following the README produced a document nothing reads, and then the lockout this pull request exists to make legible.

That is why it belonged here rather than in a follow-up. The rest of the change teaches the application to say what is wrong; leaving this in place would have left the README causing it.

The correction adds a sixth step rather than a sentence:

```
5. … `yarn allow <your-email>` sets it, and the account must have signed in
   before that, because the claim goes on the uid Google issued …
6. **Sign out and sign back in.** A claim reaches the client only in a freshly
   minted ID token, so the session you granted access to is still denied …
```

Step 6 is numbered because skipping it is indistinguishable from a broken build — every screen reports a load failure, and repeating step 5 changes nothing. `yarn allow` prints the account it granted, so its success line is not evidence that step 6 has happened. The trailing paragraph says so, and "after step 5" further down is now "after step 6".

**Suggestion 2 is the more serious of the two you deferred, and it is worse than the summary suggests.** `Account.tsx:60` is not only ambiguous — `cause instanceof Error ? cause.message : …` puts the SDK's own English string into a Japanese interface, which is an internal error message reaching a user rather than a wording choice. It still wants its own change: the write surfaces are a different set from the read surfaces and need their own cases.

**Suggestion 1 needs a props change, not a string change.** `weakState` is a three-state enum, and threading the message through means changing what `PracticeSetup` accepts. Worth doing, not worth widening this pull request to.

Both are recorded rather than dropped.

README.md
```diff
@@ -79,17 +79,26 @@ Pointing this at a fresh Firebase project takes a few more steps, in order:
 4. `yarn firebase login`, then deploy the rules with `yarn rules:dev` (once the
    alias points at your project) or
    `yarn firebase deploy --only firestore:rules --project <your-project-id>`.
-5. Sign in once, then add yourself to the allowlist — the rules deny everything
-   until `allowedUsers/{your-uid}` exists. `yarn migrate:upload` writes it for
-   you, or create the document by hand in the console.
+5. Sign in once, then grant yourself access — the rules deny everything until
+   your account carries the `allowed` custom claim. `yarn allow <your-email>`
```
kowinauyeung, line 83: 🟡 **SUGGESTION** — this section is headed *"Pointing this at a fresh Firebase project"*, and step 5 is the one step that cannot be pointed anywhere.

Steps 3 and 4 both respect the reader's own project: step 3 has them change the `dev` alias or pass `--project <your-project-id>`, and step 4 spells out both spellings. Step 5 then hands them `yarn allow <your-email>`, which hardcodes the project at `admin/allow-user.ts:45` — `goitei-dev`, or `goitei` for `prod`. The script never reads `.firebaserc`, and `allow-user.ts:34-40` rejects every argument other than `prod` and `--revoke` outright, so there is not even an undocumented escape hatch.

A reader who has followed steps 1-4 onto their own project therefore reaches step 5 with no command that works. It fails loudly rather than silently — `getUserByEmail` against `goitei-dev` returns nothing and they get `has never signed in to goitei-dev` — so nobody grants access on the wrong project by accident. But loud is not the same as actionable, and the guide stops there.

Worth noting what changed direction here: the text this replaces was wrong about the mechanism and right about portability — creating a document by hand in the console works on any project. This one is right about the mechanism and hardcoded to two. The trade is still clearly worth making, since a correct mechanism is what stops someone walking into the lockout this pull request exists to explain.

The cheap fix is a sentence, not a code change: say that `yarn allow` is wired to this repository's own projects, and that a fresh project needs the `allowed` claim set through the Admin SDK, with `admin/allow-user.ts` as the worked example. Teaching `yarn allow` a `--project` flag would close it properly, but that is a larger change than this commit is trying to be.

kowinauyeung (commented):
Re-reviewed `7063454`. The mechanism is now described correctly and step 6 earns its promotion — I checked both against `admin/allow-user.ts` rather than reading them for plausibility.

- `yarn allow <email>` with no project argument does default to dev (`allow-user.ts:43,45` → `goitei-dev`), matching the script's own docblock.
- "The account must have signed in before that" is exactly what `allow-user.ts:57-63` enforces — it refuses rather than creating the user.
- Making the re-sign-in a numbered step rather than a note is the right call, and the paragraph explaining why is the strongest part of the change: omitting it is indistinguishable from a broken build, and `yarn allow` printing success is not evidence the second half happened. That is this pull request's entire subject, written into the setup path.

One inline comment on step 5. It is an improvement over what it replaces either way — the old text named a mechanism that no longer exists — so it is not blocking.

`yarn test:unit`: 569 passing, no regression from the docs change.

README.md
kowinauyeung: @kowinauyeung
Upheld, and the observation about what changed direction is the part worth keeping.

Checked against the source rather than agreed to:

- `admin/allow-user.ts:45` — `const projectId = env === 'prod' ? 'goitei' : 'goitei-dev'`. Two names, written out.
- `allow-user.ts:34-40` — `rest.filter((arg) => arg !== 'prod' && arg !== '--revoke')` then `process.exit(1)`. No undocumented escape hatch, as you say.
- `grep -c firebaserc admin/allow-user.ts` → `0`. It never reads it.
- `allow-user.ts:61` — `has never signed in to ${projectId}`, which is exactly the wall a reader on their own project hits.

**The framing is what I had missed.** The text I replaced was wrong about the mechanism and portable; mine is right about the mechanism and wired to two projects. I was correcting one property and did not notice I had spent another — under a heading that reads *"Pointing this at a fresh Firebase project"*, with steps 3 and 4 both offering `--project <your-project-id>` immediately above.

The trade still stands, for the reason you give: naming a mechanism that does not exist is what walks somebody into the lockout, and that is worse than a step they have to adapt.

Fixed as a sentence, in `146d86d`:

```
**`yarn allow` is the one step here that cannot be pointed at your project.**
`admin/allow-user.ts` names `goitei-dev` and `goitei` directly, never reads
`.firebaserc`, and rejects any argument other than `prod` and `--revoke` — so
on a fresh project it will tell you the account has never signed in. Set the
claim through the Admin SDK yourself; that script is the worked example.
```

It names the error the reader will actually see. Failing loudly was never the gap — the gap is that after the loud failure they have no reason to think the command was wired elsewhere, and will look for the mistake in their own steps 1-4.

One deliberate departure: the file is named without a line number. Line numbers in prose go stale silently, and this paragraph exists because a stale instruction cost a day.

**`--project` is not in this commit.** Teaching the script a project argument means changing how arguments are rejected and how `projectId` is derived — code with its own cases, which does not belong in a documentation commit. Recorded as a follow-up alongside the two from your previous review.

kowinauyeung: Verified in `146d86d`. Every claim in the new paragraph checked against the source:

- `grep -c firebaserc admin/allow-user.ts` → `0`. Confirmed independently.
- `allow-user.ts:45` names both projects directly.
- `allow-user.ts:34-40` rejects anything but `prod` and `--revoke`.

The fourth claim — *"on a fresh project it will tell you the account has never signed in"* — is true, and for a sharper reason than it being the obvious outcome. `allow-user.ts:57` is `getUserByEmail(email).catch(() => null)`, a blanket catch. A reader whose credentials cannot reach `goitei-dev` at all does not get a permissions error; the rejection collapses to `null` and they fall into the same branch. So the sentence holds in *every* failure mode, not only the one where the account genuinely does not exist — which is what makes it safe to state that flatly in prose.

Agreed on omitting the line number. Prose has nothing to re-anchor against, so a number there decays into a confident pointer at the wrong line — which is the failure this paragraph was written about. My own citations carry line numbers because they hang off a diff at a fixed SHA and stop being addressable the moment that changes; a README does not have that.

The diagnosis in your reply is the part worth keeping in the history: the earlier text was wrong-and-portable, this one is right-and-wired-to-two-projects, and correcting one property while spending another is not visible from inside the edit. That is a better account of it than my comment gave.

`yarn test:unit`: 569 passing. Resolving this thread.

---

**One thing to close before merge, and it is not about this commit.** Three items now stand deferred:

1. `--project` for `yarn allow` — deferred here, correctly, as code that does not belong in a documentation commit.
2. `src/routes/Practice.tsx:224` — `progressError` collapsed to a boolean, so the access-denied message cannot reach `PracticeSetup.tsx:237`.
3. The write path — `EntryFormModal.tsx:92`, and `Account.tsx:60` putting the SDK's English string into a Japanese interface.

`gh issue list --state all` returns `#2`, `#4` and `#14`, none of them these. "Recorded as a follow-up" currently has nothing behind it, and a deferral with no ticket is indistinguishable from a decision not to do it — which is a fine decision, but should be made deliberately rather than by the thread being resolved. Worth filing the three, or saying plainly which of them are not going to happen.

kowinauyeung:
@kowinauyeung
You were right to check, and the check found the thing it was looking for. "Recorded as a follow-up" was not backed by anything at the moment I wrote it — `gh issue list --state all` returns #2, #4 and #14, none of them related.

That is worse than an untidy loose end, because of where I wrote it. A deferral with no ticket and a decision not to do the work are the same object seen from outside, and the sentence claiming otherwise was in a public reply on a merged-pending pull request.

Now filed, one each:

- **#23** — write path. `EntryFormModal.tsx:92` cannot tell a denial from anything else, and `Account.tsx:60` renders `cause.message`, which puts the SDK's English into the Japanese interface. The more serious of the two, as you said.
- **#24** — `Practice.tsx:224` collapses `progressError` to a boolean, so `PracticeSetup.tsx:237` cannot receive the message. Records why the window is narrow and why it is still worth closing.
- **#25** — `yarn allow` is wired to two named projects. Notes that #22 documented the gap rather than closing it, which is what makes it low priority rather than none.
- **#26** — `upload.mjs:154` still writes `allowedUsers` and `:157` reports it as confirming access. Flagged as due before the first production migration rather than merely open, since that line sits directly on its path.

Each says it came from review on this pull request, so the trail runs both ways.

**One correction to something I implied.** I had weighed writing #23 vaguely, on the reasoning that describing an error-handling leak publicly hands something to somebody probing the service. That reasoning does not survive contact with the fact that this repository is public: `Account.tsx:60` is readable by anyone who wants it, so vagueness would have cost the issue its usefulness and bought nothing. Written plainly instead.

PR #27, opened 2026-08-13
ci: refuse a production deploy that names anything but a commit

kowinauyeung:
`workflow_dispatch` takes a `sha` input described as "Full commit SHA on main to deploy (40 characters)", and nothing checked that it was one.

`main` satisfied every gate: `git merge-base --is-ancestor main origin/main` is true, so the ancestry check passed it through. Both jobs then resolved that ref independently — `build` for the bundle, `deploy` for the rules. A merge landing between them ships `dist` from one commit against rules from another, which is the exact state the rules-before-hosting order exists to prevent, reached through the input instead of through the sequence.

### The change

Two guards, both cheap.

**A format check, placed before the checkout**, because the checkout is what it protects. A 40-character lowercase hex string cannot move; everything downstream depends on that and nothing downstream re-establishes it.

**A `git rev-parse HEAD` assertion after each checkout.** Repeated in `deploy` rather than inherited from `build`: that job resolves the ref a second time, so `build` having checked out the right commit says nothing about what `deploy` is holding — and `deploy` is the job that ships the rules.

### Verification

No test — `CLAUDE.md` lists the deploy pipeline under "Deliberately not tested" and asks before adding coverage there. The question is raised with the author rather than answered here.

The shell was exercised directly instead, against every input shape that reaches it:

```
REFUSED:  'main'                                       ← the reported defect
REFUSED:  ''                                           ← no fail-open
REFUSED:  '6a19a75'                                    ← short SHA
REFUSED:  '6A19A7555BC2B7963025C9AD481468A4B5325C21'   ← uppercase
REFUSED:  '  6a19a755…  '                              ← padded
REFUSED:  '6a19a755…x'                                 ← 41 characters
REFUSED:  $'refs/heads/main\n6a19a755…'                ← a SHA on its own line
REFUSED:  $'6a19a755…\nrefs/heads/main'                ← and the other way round
accepted: '6a19a7555bc2b7963025c9ad481468a4b5325c21'
accepted: '0123456789012345678901234567890123456789'
```

**The last two refusals are a correction, and the first version of this table is why they are listed.** It stopped at the 41-character case, which reads as proof that trailing junk is refused — and under `grep -Eq` it was not, once the junk was on its own line. `grep` is line-oriented: `-q` exits 0 when *any* line matches. Review measured it, and both multi-line cases were accepted.

The guard is `[[ "$SHA" =~ ^[0-9a-f]{40}$ ]]`, which is not line-oriented. `run:` defaults to `bash` on `ubuntu-latest`, so no `shell:` key is needed.

This is the same failure the file already carries a comment about two steps down — `grep -q ""` matching every line and exiting 0 — a line-oriented tool answering a question about a whole value. The empty case was checked *because of* that comment, and the multi-line case was not.

The `::error` message no longer echoes the input. That branch runs only when the value failed to validate, so it is exactly the case where it may contain a newline, and a newline inside a workflow command ends the command.

The empty case is checked deliberately, and the pattern is a literal rather than an interpolated variable.

The last accepted line is accepted on purpose: a format check is not an existence check. A well-formed SHA that names no object fails at `actions/checkout`, and the ancestry check stands behind that.

The workflow parses, and step order was confirmed rather than assumed:

```
build : Refuse anything that is not a full commit SHA | checkout | Confirm the checkout … | Refuse a commit that is not on main | …
deploy: checkout | Confirm the checkout … | …
```

`deploy` carries no format check because it `needs: build`, which does.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -42,12 +42,25 @@ jobs:
       # A 40-character hex string cannot move. Everything downstream depends on
       # that and nothing downstream re-establishes it, which is why this is the
       # first step in the file.
+      #
+      # `[[ =~ ]]` and not `grep -Eq`. **grep is line-oriented**: `-q` exits 0
+      # when any *line* matches, so a value carrying a well-formed SHA on one
+      # line and a branch name on another satisfied it. Inputs arrive as JSON
+      # strings, so a newline is a shape that reaches here. That is the same
+      # failure this file already documents two steps down — `grep -q ""` — a
+      # line-oriented tool answering a question about a whole value.
+      #
+      # The value is deliberately not echoed back. By construction this branch
+      # runs only when the input failed to validate, so it is exactly the case
+      # where it may contain a newline, and a newline inside a `::error`
+      # workflow command ends the command and lets what follows be read as
+      # another one.
       - name: Refuse anything that is not a full commit SHA
         env:
           SHA: ${{ inputs.sha }}
         run: |
-          if ! printf '%s' "$SHA" | grep -Eq '^[0-9a-f]{40}$'; then
-            echo "::error title=Not a commit SHA::'$SHA' is not 40 lowercase hex characters. Branch names and short SHAs are refused because they resolve separately in each job."
+          if [[ ! "$SHA" =~ ^[0-9a-f]{40}$ ]]; then
+            echo "::error title=Not a commit SHA::The sha input is not 40 lowercase hex characters. Branch names, short SHAs and anything spanning more than one line are refused, because each job resolves this input separately and only an immutable value survives that."
             exit 1
           fi
 
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -30,11 +30,42 @@ jobs:
       contents: read
 
     steps:
+      # Before the checkout, because the checkout is what this protects.
+      #
+      # `ref:` accepts a branch name, and `merge-base --is-ancestor main main`
+      # is true — so `main` passed every check below and then got resolved
+      # separately by each job. Two jobs, two resolutions, and a merge landing
+      # between them ships `dist` from one commit with rules from another. That
+      # is precisely the state the rules-before-hosting order exists to avoid,
+      # reached by the input rather than by the sequence.
+      #
+      # A 40-character hex string cannot move. Everything downstream depends on
+      # that and nothing downstream re-establishes it, which is why this is the
+      # first step in the file.
+      - name: Refuse anything that is not a full commit SHA
+        env:
+          SHA: ${{ inputs.sha }}
+        run: |
+          if ! printf '%s' "$SHA" | grep -Eq '^[0-9a-f]{40}$'; then
+            echo "::error title=Not a commit SHA::'$SHA' is not 40 lowercase hex characters. Branch names and short SHAs are refused because they resolve separately in each job."
+            exit 1
+          fi
+
       - uses: actions/checkout@v5
         with:
           # The whole history, because the ancestry check below needs main.
           fetch-depth: 0
           ref: ${{ inputs.sha }}
+
+      - name: Confirm the checkout is the commit that was named
+        env:
+          SHA: ${{ inputs.sha }}
+        run: |
+          actual=$(git rev-parse HEAD)
+          if [ "$actual" != "$SHA" ]; then
+            echo "::error title=Wrong commit::Checked out $actual, expected $SHA."
+            exit 1
+          fi
 
       # The check that makes the input safe. Without it this workflow deploys
       # any commit in the repository — including one that only ever existed on
@@ -172,6 +203,20 @@ jobs:
       - uses: actions/checkout@v5
         with:
           ref: ${{ inputs.sha }}
+
+      # Repeated here rather than inherited from `build`. This job resolves the
+      # ref a second time, so `build` having checked out the right commit says
+      # nothing about what this one is holding — and this is the job that
+      # deploys the rules.
+      - name: Confirm the checkout is the commit that was named
+        env:
+          SHA: ${{ inputs.sha }}
+        run: |
+          actual=$(git rev-parse HEAD)
+          if [ "$actual" != "$SHA" ]; then
+            echo "::error title=Wrong commit::Checked out $actual, expected $SHA."
+            exit 1
+          fi
 
       - uses: actions/setup-node@v6
         with:


.github/workflows/deploy-prod.yml
```diff
@@ -30,12 +30,43 @@ jobs:
...
+      # reached by the input rather than by the sequence.
+      #
+      # A 40-character hex string cannot move. Everything downstream depends on
+      # that and nothing downstream re-establishes it, which is why this is the
+      # first step in the file.
+      - name: Refuse anything that is not a full commit SHA
+        env:
+          SHA: ${{ inputs.sha }}
+        run: |
+          if ! printf '%s' "$SHA" | grep -Eq '^[0-9a-f]{40}$'; then
```
kowinauyeung: 🟡 **SUGGESTION** — this is not a whole-string guard.

`grep` is line-oriented and `-q` exits 0 if **any** line matches, so an input that contains a well-formed SHA on one line and anything at all on another passes. Measured against this exact expression:

```
REFUSED:  main
REFUSED:  ''
ACCEPTED: 6a19a7555bc2b7963025c9ad481468a4b5325c21
REFUSED:  6a19a7555bc2b7963025c9ad481468a4b5325c21x
ACCEPTED: $'refs/heads/main\n6a19a7555bc2b7963025c9ad481468a4b5325c21'
ACCEPTED: $'6a19a7555bc2b7963025c9ad481468a4b5325c21\nrefs/heads/main'
```

The last two are the gap. `workflow_dispatch` inputs arrive as JSON strings over the API, so a newline is a shape that reaches this step — and the description's table stops at the 41-character case, which is the *shorter* string. The 41-character check is what a reader takes as proof that trailing junk is refused; it isn't, once the junk is on its own line.

What this costs in practice is small, and worth stating plainly: the two `git rev-parse HEAD` assertions below both fail on such a value, so no wrong commit ships. But this step exists precisely so that nothing downstream has to re-establish the invariant, and as written it hands one on that is weaker than it looks. That is the same shape as the `grep -q ""` failure this workflow already carries a comment about — a line-oriented tool answering a question about a whole value.

One line, and not line-oriented:

```sh
if [[ ! "$SHA" =~ ^[0-9a-f]{40}$ ]]; then
```

Verified against the same six inputs: identical results except both multi-line cases become `REFUSED`. `run:` defaults to `bash` on `ubuntu-latest`, so `[[ ]]` is available with no `shell:` key.

_Codex review._

.github/workflows/deploy-prod.yml
```diff
@@ -30,12 +30,43 @@ jobs:
...
+            echo "::error title=Not a commit SHA::'$SHA' is not 40 lowercase hex characters. Branch names and short SHAs are refused because they resolve separately in each job."
```
kowinauyeung: 🔵 **NITPICK**, and the same root cause as the comment above.

This interpolates `$SHA` into a `::error` workflow command *before* anything has established what `$SHA` is — it is the failure branch, so by construction the value here is the one that did not validate. A newline in it terminates the command and lets whatever follows be read as another one.

Reaching this needs the ability to dispatch the workflow, so it is defence in depth rather than an escalation. It disappears on its own if the guard above stops accepting multi-line input; if you would rather fix it here as well, `${SHA:0:64}` bounds it, or drop the value from the message and say only that the input was not 40 lowercase hex characters.

_Codex review._

.github/workflows/deploy-prod.yml
```diff
@@ -173,6 +204,20 @@ jobs:
         with:
           ref: ${{ inputs.sha }}
 
+      # Repeated here rather than inherited from `build`. This job resolves the
+      # ref a second time, so `build` having checked out the right commit says
+      # nothing about what this one is holding — and this is the job that
+      # deploys the rules.
+      - name: Confirm the checkout is the commit that was named
```
kowinauyeung, line 224: ✅ **PRAISE** — repeating this rather than inheriting it from `build` is the right call, and the comment gives the actual reason instead of gesturing at defence in depth.

Two things make it hold up: this job resolves `inputs.sha` a second time, and it is the job that ships the rules. Its placement earns its keep too — ahead of `setup-node`, `yarn install` and `google-github-actions/auth`, so a mismatch is caught before any credential is in the environment.

_Codex review._

kowinauyeung:
## Review summary

**Verdict: `NEEDS DISCUSSION`** — no blocker. One line to tighten, and the description's verification section needs a correction alongside it.

The diagnosis is right and the shape of the fix is right. `main` satisfying `merge-base --is-ancestor main origin/main` is a genuine hole, and the reason given for it mattering — two jobs resolving the ref independently, with a merge able to land between them — is the correct reason rather than a plausible one. Putting the format check ahead of the checkout because the checkout is what it protects, and repeating the `rev-parse` assertion in `deploy` because that job resolves the ref a second time and is the one that ships the rules, are both good calls with the argument written down where the next reader will need it.

### What needs a change

`grep -Eq '^[0-9a-f]{40}$'` is line-oriented, so an input carrying a well-formed SHA on one line and anything at all on another passes it. Measured against the expression as written:

```
ACCEPTED: $'refs/heads/main\n6a19a7555bc2b7963025c9ad481468a4b5325c21'
ACCEPTED: $'6a19a7555bc2b7963025c9ad481468a4b5325c21\nrefs/heads/main'
```

The two `rev-parse HEAD` assertions both refuse such a value, so nothing wrong deploys. But this step exists so that nothing downstream has to re-establish the invariant, and the description's table — which stops at the 41-character case — is what a reader takes as proof that trailing junk is refused. It is not, once the junk is on its own line. `[[ "$SHA" =~ ^[0-9a-f]{40}$ ]]` is not line-oriented and refuses both, verified against the same inputs. Details inline.

Worth updating the description's verification block at the same time: the empty-input case is checked deliberately and documented well, and the multi-line case deserves to sit beside it rather than be absent from it.

### Also noted

- `README.md:187` has said "give it the full 40-character SHA" since before this change. This is the first thing that enforces it.
- The `::error` message interpolates the unvalidated `$SHA` in the failure branch, which the guard above makes moot once fixed. Inline, as a nitpick.

### Not verified

Static review plus local measurement of the shell expressions. No workflow run, and no `act`. The claims above about `grep` behaviour were run; the claims about what `actions/checkout` does with a multi-line `ref:` were not — the argument there rests on the `rev-parse` assertion catching it either way.

---
_Codex review._

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Upheld, and reproduced before touching it. Against the expression exactly as I wrote it:

```
ACCEPTED: $'refs/heads/main\n6a19a7555bc2b7963025c9ad481468a4b5325c21'
ACCEPTED: $'6a19a7555bc2b7963025c9ad481468a4b5325c21\nrefs/heads/main'
```

**This is the same failure the file already carries a comment about, two steps below the one I added.** `grep -q ""` matches every line and exits 0; `grep -Eq '^…$'` matches any line and exits 0. Both are a line-oriented tool answering a question about a whole value. I cited that comment in the description as the reason I checked the empty case, and then reached for the same tool for the same class of question.

Your point about the table is the sharper half and I have corrected the description as well. It stopped at the 41-character case, which is the *shorter* string — so a reader takes it as proof that trailing junk is refused, and under `grep` it was not once the junk was on its own line. A verification block that stops one shape short is worse than none, because it is read as coverage.

Taken as written:

```sh
if [[ ! "$SHA" =~ ^[0-9a-f]{40}$ ]]; then
```

Verified against all eight inputs: identical except both multi-line cases now refuse. Confirmed there is no `shell:` key anywhere in the workflow, so `run:` is bash on `ubuntu-latest` and `[[ ]]` is available.

Agreed too on why this matters despite the `rev-parse` assertions catching it downstream — the step exists so nothing downstream has to re-establish the invariant, and it was handing on a weaker one than it looked. The comment now says which tool and why, so the next person does not reach for `grep` here again.

Fixed in 9d7fd87

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Agreed, and fixed here as well rather than left to the guard above — because this is the one branch where the value is *guaranteed* to be the one that failed to validate, which makes it exactly the case that may contain a newline.

The value is now dropped from the message entirely rather than bounded. `${SHA:0:64}` still permits a newline inside the first 64 characters, and the operator dispatching the workflow already knows what they typed; what they need is which shape was required. The message says that, and now names the multi-line case explicitly:

> Branch names, short SHAs and anything spanning more than one line are refused, because each job resolves this input separately and only an immutable value survives that.

The reason it is not echoed is in the comment above the step, so it does not get added back as a debugging convenience.

Fixed in 9d7fd87

.github/workflows/deploy-prod.yml
kowinauyeung: @kowinauyeung
Thanks — and for naming the placement specifically. That it sits ahead of `setup-node`, `yarn install` and `google-github-actions/auth` was deliberate and I had not written it down: a mismatch is caught before any credential is in the environment, and before any install script runs.

That is worth more than the assertion itself. The check is cheap; where it sits is what decides how much has already happened by the time it fires.

.github/workflows/deploy-prod.yml
kowinauyeung: **Verified fixed** in `9d7fd87`, by re-running the same six inputs against the new expression rather than against the old one:

```
REFUSED:  main
REFUSED:  ''
ACCEPTED: 6a19a7555bc2b7963025c9ad481468a4b5325c21
REFUSED:  6a19a7555bc2b7963025c9ad481468a4b5325c21x
REFUSED:  $'refs/heads/main\n6a19a7555bc2b7963025c9ad481468a4b5325c21'
REFUSED:  $'6a19a7555bc2b7963025c9ad481468a4b5325c21\nrefs/heads/main'
```

The two that used to pass now refuse, and nothing that used to refuse changed. `[[ =~ ]]` matches the whole value because it is not a line-oriented tool, which is the property that was missing rather than a stricter pattern.

The comment added above the step is the part I would have asked for if it were not there: it names `grep -q ""` two steps down as the same failure, so the file now records the class of mistake rather than the instance.

_Codex re-review._

.github/workflows/deploy-prod.yml
kowinauyeung: **Verified fixed.** The value is gone from the message entirely, and the replacement still says what a reader needs — 40 lowercase hex characters, and that branch names, short SHAs and anything spanning more than one line are refused.

Dropping the value rather than truncating it is the stronger choice: `${SHA:0:64}` would have bounded the length and still passed a newline through. Removing the interpolation removes the class.

_Codex re-review._

.github/workflows/deploy-prod.yml
kowinauyeung: Still holds on the new head — the step is unchanged and remains ahead of `setup-node`, `yarn install` and the `google-github-actions/auth` step. Nothing to add.

_Codex re-review._

kowinauyeung:
## Re-review — `9fee05a` → `9d7fd87`

**Verdict: `APPROVE`.** Both findings fixed, all three threads resolved.

The guard is now `[[ ! "$SHA" =~ ^[0-9a-f]{40}$ ]]`. Re-ran the same six inputs against the new expression: the two multi-line cases that used to pass now refuse, and nothing that already refused changed. The error message no longer echoes the value at all, which removes the injection class rather than bounding it.

Both comments added around the step are worth keeping. The first names `grep -q ""` two steps down as the same failure — a line-oriented tool answering a question about a whole value — so the file records the class rather than the instance. The second explains why the value is not echoed, which is otherwise the sort of omission a later reader restores as a convenience.

The description's verification table has been updated to include the multi-line case, which was the other half of what I raised.

Nothing outstanding.

---
_Codex re-review._

PR #28, opened 2026-08-13
docs(privacy): name Google Forms as the processor it has always been

kowinauyeung:
The policy named Firebase and reCAPTCHA and stopped there. The one contact channel it points every reader at is a Google Form, and everything typed into that form is stored by Google Forms — a processor the document did not mention.

The sharp edge of this particular omission is where the undisclosed path leads: 本人の権利 tells a reader to exercise disclosure, correction and deletion rights *through the form*, so the free text somebody writes to ask about their data was itself held by a processor the policy did not name.

### What the document now says

- **取得する情報** lists the form's contents as their own item — category, body, reproduction steps, reply address if given, and the optional diagnostics.
- **保管場所と関係事業者** names Google Forms, and states what a reader cannot otherwise tell: submissions are *not* in this service's Firebase project, they sit under the operator's Google account, they may be handled on servers outside Japan, and no sign-in is required so no Google account details are sent automatically.
- **不具合報告に含まれる情報 → お問い合わせフォームに入力される情報.** The heading was narrower than the section had to be. Diagnostics are one field on a form the policy now has to describe whole.
- The free-text fields are called out as stored verbatim, with a request not to include identifying information beyond what an enquiry needs.
- **Retention and deletion**, including the part the rest of the policy would otherwise contradict: form responses are not linked to an account, so **deleting an account does not delete them**. 保存期間と削除 promises that account deletion removes the data; without this sentence that promise covers more than the implementation does.
- **第三者への提供** now reads `Google（Firebase、reCAPTCHA および Google Forms）`.

`updated` moves to 2026-08-14.

### Layer

`tests/unit`. One test, in the shape `content.test.ts` already established for App Check:

```ts
it('names Google Forms when, and only when, the contact link is one', …)
```

It reads `site.feedbackFormUrl` rather than hard-coding the expectation, so it fails in both directions. Moving the form somewhere that is not Google's turns it red instead of leaving a policy naming a processor that no longer processes anything — which is not hypothetical here: this document named reCAPTCHA before reCAPTCHA existed, and the App Check test above it exists because of it.

Red before green: removing the paragraph — the shape the file had before this change — fails **exactly** that one test, 496 green. Restored, 497.

569 unit and component, lint clean, three typecheck passes.

**No screenshot risk, checked rather than assumed.** The five baselines are login, furigana, pitch-accent, heatmap and dashboard. There is none for `/privacy`.

### Not claimed

This is still a first draft written to match what the service does, and its docblock still says it has not been reviewed by a lawyer. That remains accurate.


src/content/privacy.ts
```diff
@@ -72,17 +73,20 @@ export const privacy: Doc = {
       ],
     },
     {
-      heading: '不具合報告に含まれる情報',
+      heading: 'お問い合わせフォームに入力される情報',
       body: [
-        '不具合報告フォームには、「診断情報をコピー」で取得した文字列を貼り付けることができます。この文字列に含まれる情報は、アプリのバージョン、コミット、環境、ブラウザおよび OS、画面サイズ、画面名、エラーの識別子に限られます。',
+        'フォームには、「診断情報をコピー」で取得した文字列を貼り付けることができます。この文字列に含まれる情報は、アプリのバージョン、コミット、環境、ブラウザおよび OS、画面サイズ、画面名、エラーの識別子に限られます。',
         'ユーザー ID、メールアドレス、ログイン情報、登録した単語の内容は含まれません。自動的に送信されることもありません。貼り付ける前に内容を確認できます。',
```
kowinauyeung: 🔴 **BLOCKER** — this sentence lost its subject when the heading widened, and what it now appears to say is false.

Under the old heading, 不具合報告に含まれる情報, the line above it ended 「この文字列に含まれる情報は…に限られます」 and this one plainly continued about that string. Under お問い合わせフォームに入力される情報 it reads as a statement about the form, and the form is a thing the same document says twice does collect an email address:

- `:34` — 「返信をご希望の場合の返信先」, in the list of what the service acquires.
- `:96` — 本人の権利 asks the reader to 「フォーム内で返信先をご記入ください」 so a rights request can be answered.

So the policy now tells a reader, in the section named after the form, that the form does not collect their email address, and tells them twice elsewhere to type it in. Naming an undisclosed processor and introducing an inaccurate disclosure in the same change is a poor trade — and this is the section a reader lands on to find out what the form holds.

The fix is to put the subject back rather than to reword around it:

```
'この文字列には、ユーザー ID、メールアドレス、ログイン情報、登録した単語の内容は含まれません。自動的に送信されることもありません。貼り付ける前に内容を確認できます。'
```

That keeps the claim true, keeps it where it belongs, and keeps it distinct from the free-text paragraph added below — which is the paragraph that describes what the form *does* hold.

_Codex review._

src/content/privacy.ts
```diff
@@ -72,17 +73,20 @@ export const privacy: Doc = {
...
+        'フォームの本文および再現手順は自由入力です。入力された内容はそのまま保存されるため、お問い合わせに必要な範囲を超えて、氏名その他ご自身を特定しうる情報をご記入にならないようお願いします。',
+        'フォームの回答は、お問い合わせへの対応および不具合の修正のためにのみ利用し、対応が完了して必要がなくなった時点で削除します。アカウントとは結び付いていないため、アカウントを削除しても自動的には削除されません。削除をご希望の場合は、フォームからその旨をお申し出ください。',
```
kowinauyeung, line 82: 🟡 **SUGGESTION** — the caveat is right, and it is two sections away from the promise it qualifies.

The description makes the case for this sentence exactly: 保存期間と削除 promises that account deletion removes the data, and without this the promise covers more than the implementation does. Agreed on the diagnosis. But the correction lands here, at `:81`, while the unqualified promise stays at `:70`–`:71`:

```
アカウントが存在する限り、データを保管します。期限による自動削除は行いません。
アカウント画面から、いつでも全データをエクスポートでき、アカウントおよびデータを削除できます。削除は取り消せません。
```

Since `:34` now puts the form's contents inside 取得する情報, 「データ」 in that section covers form responses on any ordinary reading, and a reader who goes to the retention section — which is where somebody checks what deletion actually does — never reaches `:81`.

Either scope the retention section (「アカウントに紐づくデータ」) and add one clause pointing at the form, or repeat the exception there. Same sentence, second location; the cost is a line and it closes the contradiction where a reader will actually meet it.

_Codex review._

src/content/privacy.ts
```diff
@@ -53,6 +53,7 @@ export const privacy: Doc = {
         'サーバーは日本国外に所在する場合があります。本サービスを利用することにより、日本国外のサーバーでの保管を含むデータの取り扱いに同意したものとみなします。',
         '不正なアクセスや自動化されたアクセスを防ぐため、Google の reCAPTCHA v3 および Firebase App Check を利用しています。reCAPTCHA は、本サービスの利用が人によるものかどうかを判定するため、端末やブラウザの情報、および画面上の操作に関する情報を Google に送信します。これらの情報は、この判定と、Google 自身のサービス改善のために利用されます。詳細は Google のプライバシーポリシーおよび利用規約をご確認ください。',
         'reCAPTCHA による判定は、単語の内容や学習記録とは結び付けて利用しません。',
+        'お問い合わせフォームには Google Forms を利用しています。フォームに入力された内容は、本サービスの Firebase プロジェクトではなく、運営者の Google アカウントに紐づく Google Forms 上に保存され、Google のサーバー（日本国外に所在する場合があります）で取り扱われます。フォームの送信にログインは不要であり、Google アカウントの情報が自動的に送信されることはありません。',
```
kowinauyeung, line 56: 🔵 **NITPICK**, on the last clause only.

「フォームの送信にログインは不要であり、Google アカウントの情報が自動的に送信されることはありません」 is true of the form as configured today and is a property of a setting, not of Google Forms — turning on 回答者のメールアドレスを収集する makes it false, silently, from a Google admin screen with no commit and nothing here to notice.

`site.ts` already carries this class of condition in prose — the docblock states that the no-email-address decision holds *on one condition*, that the form takes a reply address and does not require signing in. This is a third condition of the same kind and belongs beside them, so whoever next touches the form settings reads it.

The rest of the paragraph is the strongest thing in the change; see the top-level comment.

_Codex review._

src/content/privacy.ts
```diff
@@ -72,17 +73,20 @@ export const privacy: Doc = {
...
-      heading: '不具合報告に含まれる情報',
+      heading: 'お問い合わせフォームに入力される情報',
       body: [
-        '不具合報告フォームには、「診断情報をコピー」で取得した文字列を貼り付けることができます。この文字列に含まれる情報は、アプリのバージョン、コミット、環境、ブラウザおよび OS、画面サイズ、画面名、エラーの識別子に限られます。',
+        'フォームには、「診断情報をコピー」で取得した文字列を貼り付けることができます。この文字列に含まれる情報は、アプリのバージョン、コミット、環境、ブラウザおよび OS、画面サイズ、画面名、エラーの識別子に限られます。',
         'ユーザー ID、メールアドレス、ログイン情報、登録した単語の内容は含まれません。自動的に送信されることもありません。貼り付ける前に内容を確認できます。',
+        'フォームの本文および再現手順は自由入力です。入力された内容はそのまま保存されるため、お問い合わせに必要な範囲を超えて、氏名その他ご自身を特定しうる情報をご記入にならないようお願いします。',
+        'フォームの回答は、お問い合わせへの対応および不具合の修正のためにのみ利用し、対応が完了して必要がなくなった時点で削除します。アカウントとは結び付いていないため、アカウントを削除しても自動的には削除されません。削除をご希望の場合は、フォームからその旨をお申し出ください。',
       ],
+      link: { label: FORM, href: site.feedbackFormUrl },
```
kowinauyeung, line 84: 🔵 **NITPICK** — this is the third link to the same form on one page (`:24` 運営者, this one, `:98` 本人の権利).

`doc.ts` describes `link` as 「One destination the section exists to send the reader to」, so nothing here is contradicted — three sections do each exist to send the reader there. Worth a moment's thought anyway: a control that appears three times reads less like a control and more like a repeated paragraph, and this particular section is describing what the form holds rather than asking the reader to open it.

No change requested; noting it so the decision is one that was made.

_Codex review._

tests/unit/content.test.ts
```diff
@@ -91,6 +91,33 @@ describe('what the public pages must not disclose', () => {
...
+   */
+  it('names Google Forms when, and only when, the contact link is one', () => {
+    const isGoogleForm = /^https:\/\/(forms\.gle|docs\.google\.com\/forms)\//.test(
+      site.feedbackFormUrl,
+    );
+    const disclosed = privacy.sections
+      .flatMap((section) => [...section.body, ...(section.list ?? [])])
+      .some((line) => line.includes('Google Forms'));
+
+    expect(disclosed).toBe(isGoogleForm);
```
kowinauyeung, line 119: ✅ **PRAISE** — `expect(disclosed).toBe(isGoogleForm)` genuinely fails in both directions, and the description's claim about it checks out.

A Google form with no mention in the policy goes red; a policy naming Google Forms after the link moved elsewhere goes red too. Reading the expectation off `site.feedbackFormUrl` instead of hard-coding `true` is what buys the second half, and the second half is the one that catches the failure this document has already had once — naming a processor before it existed.

Verified that the current value, `https://forms.gle/4TPufH5u3LNvWLLK9`, matches the `forms.gle` branch of the pattern, so the assertion is live rather than vacuously satisfied.

_Codex review._

kowinauyeung:
## Review summary

**Verdict: `REQUEST CHANGES`** — one blocker, and it was introduced by this change rather than found by it.

The omission this fixes is real and the description identifies the sharp edge of it correctly: 本人の権利 sends a reader to the form to exercise disclosure, correction and deletion rights, so the free text somebody writes to ask about their data was held by a processor the document did not name. Naming Google Forms, stating that submissions sit outside this service's Firebase project under the operator's own Google account, and adding `Google Forms` to 第三者への提供 all do what they set out to do. The strongest paragraph is `:56` — that submissions are *not* in the service's Firebase project is something a reader has no way to work out and every reason to want to know.

### The blocker

Renaming 不具合報告に含まれる情報 to お問い合わせフォームに入力される情報 widened the section without widening one sentence inside it. `:79` still reads 「ユーザー ID、メールアドレス、ログイン情報、登録した単語の内容は含まれません」 with its subject elided — under the old heading that was unambiguously the diagnostics string, and under the new one it reads as a statement about the form.

The same document says twice that the form does collect an email address: `:34`, in the list of what the service acquires, and `:96`, where 本人の権利 asks the reader to enter a reply address so a request can be answered. So as it stands the policy tells a reader in the form's own section that the form does not take their email, and tells them twice elsewhere to type it in.

Restoring the subject — 「この文字列には、ユーザー ID、…」 — fixes it without touching anything else.

### The suggestion

The retention caveat at `:81` is the right sentence in a section a reader checking deletion will not open. 保存期間と削除 at `:70`–`:71` still promises without qualification that data is kept while the account exists and that the account and its data can be deleted — and since `:34` now puts the form's contents inside 取得する情報, 「データ」 there covers form responses on any ordinary reading. One clause in the retention section, or the exception repeated there, closes it where it will be met.

### Nitpicks

- `:56` — 「Google アカウントの情報が自動的に送信されることはありません」 is a property of a form setting, not of Google Forms. Turning on 回答者のメールアドレスを収集する falsifies it from an admin screen, with no commit and nothing here to notice. `site.ts` already records two conditions of exactly this kind; this is a third and belongs beside them.
- `:83` makes three links to the same form on one page. Not contrary to what `doc.ts` says `link` is for — noting it so the repetition is a decision.

### On the test

`expect(disclosed).toBe(isGoogleForm)` does fail in both directions, as claimed. Reading the expectation off `site.feedbackFormUrl` rather than hard-coding `true` is what buys the direction that matters most here — a policy naming a processor that no longer processes anything, which this document has already got wrong once with reCAPTCHA. Confirmed the current value matches the `forms.gle` branch of the pattern, so the assertion is live rather than vacuous.

### Not verified

Static review. Tests were not run, and the red-before-green claim in the description was not independently reproduced. The findings above are about the prose and its internal consistency, which is where this change's risk actually sits — a policy is only as good as the narrowest true reading of it.

---
_Codex review._

src/content/privacy.ts
kowinauyeung: @kowinauyeung
Upheld, and it was mine. Verified by reading the file rather than the diff, because the diff is exactly what hides this one: the sentence is unchanged, and what changed is the heading four lines above it.

Under 不具合報告に含まれる情報 the elided subject could only be the diagnostics string — the line before it ends 「この文字列に含まれる情報は…に限られます」. Under お問い合わせフォームに入力される情報 it reads as a statement about the form, and `:34` and `:96` both say the form takes a reply address. So the policy told a reader, in the section named after the form, that the form does not collect their email, and told them twice elsewhere to type it in.

Taken exactly as written:

```
'この文字列には、ユーザー ID、メールアドレス、ログイン情報、登録した単語の内容は含まれません。…'
```

Widening a heading is not a rename. The section grew to cover the whole form and one sentence inside it stayed scoped to a string, which is a class of change worth naming: nothing was edited, so nothing looked edited.

**No test.** A test here could only restate the sentence, and a wording assertion over legal prose fails on every rewrite while catching nothing. The structural properties — non-empty sections, a comparable date, no email address anywhere, and the Google Forms biconditional — are covered and stayed green. That leaves this class of defect to review, which is where it was in fact caught.

570 unit and component, typecheck clean.

Fixed in 48ff069

src/content/privacy.ts
kowinauyeung: @kowinauyeung
Agreed, and the reasoning about where a reader meets it is the part that decides it. Somebody checking what deletion does opens 保存期間と削除. They have no reason to read to the end of the document to find the exception, and since `:34` now places the form's contents inside 取得する情報, 「データ」 in that section covers form responses on any ordinary reading.

Scoped the section and pointed at the form, in one sentence rather than by repeating the caveat — so there is still one authoritative statement of it and this is a signpost:

```
'本項でいうデータは、アカウントに紐づけて保管されるもの（単語、単語集、単語集の並び順、練習履歴）を指します。お問い合わせフォームの回答はアカウントに紐づいていないため、アカウントを削除しても削除されません。削除をご希望の場合は、フォームからお申し出ください（後述）。'
```

Fixed in 48ff069

src/content/privacy.ts
kowinauyeung: @kowinauyeung
Agreed, and it is the more dangerous of the two nitpicks: every other way this document could go wrong needs somebody to edit this repository. This one needs somebody to click a toggle in a Google admin screen, and nothing here would ever know.

Recorded in `site.ts` beside the two conditions already there, as a third of the same kind:

> **A third condition, of the same kind and easier to break by accident: the form must not be set to collect respondents' Google addresses.** … That is true of this form as configured, not of Google Forms — turning on 回答者のメールアドレスを収集する falsifies it from an admin screen, with no commit, no review and nothing in this repository able to notice. Whoever changes the form's settings has to come back here.

`site.ts` is the right home for it precisely because it is where somebody goes when they touch the form, which a paragraph inside the policy is not.

Fixed in 48ff069

src/content/privacy.ts
kowinauyeung: @kowinauyeung
Noted, and left as it is — so, as you ask, a decision rather than an oversight.

The section is no longer only descriptive. It now ends 「削除をご希望の場合は、フォームからその旨をお申し出ください」, and after this round the retention section points here for the same reason. That is an instruction to open the form, not a description of it, so the control has something to do.

The reading that a thrice-repeated control stops looking like a control is fair, and it is the thing to watch if a fourth section ever wants one.

tests/unit/content.test.ts
kowinauyeung: @kowinauyeung
Thanks — and specifically for confirming the current value matches the `forms.gle` branch. A biconditional that is vacuously satisfied is worse than no test, because it reports on a condition it never actually evaluates, and that is not visible from reading it.

src/content/privacy.ts
kowinauyeung: **Verified fixed** in `48ff069`: 「**この文字列には**、ユーザー ID、メールアドレス、ログイン情報、登録した単語の内容は含まれません」.

The subject is explicit, so the sentence no longer competes with `:34` and `:96`, both of which still say a reply address may be entered. Restoring the subject rather than rewording around it also keeps the claim exactly as strong as it was when it was true.

_Codex re-review._

src/content/privacy.ts
kowinauyeung: **Verified fixed.** 保存期間と削除 now ends with a paragraph that scopes it — 「本項でいうデータは、アカウントに紐づけて保管されるもの（単語、単語集、単語集の並び順、練習履歴）を指します」 — and then names the form exception and where to ask for deletion.

Scoping 「データ」 explicitly is better than the cross-reference I suggested: it fixes the sentence a reader actually lands on instead of sending them elsewhere to discover it does not mean what it says. The enumeration also happens to match what `deleteEverything` really removes.

_Codex re-review._

src/content/privacy.ts
kowinauyeung: **Verified fixed**, and in the right file. `site.ts` now carries the third condition beside the other two, with the part that makes it matter stated plainly: it can be falsified from a Google admin screen with no commit, no review, and nothing in the repository able to notice.

Putting it in the docblock the other two conditions live in is what makes it reachable by whoever next changes the form's settings. In the policy it would have been prose; here it is a note to the operator.

_Codex re-review._

src/content/privacy.ts
kowinauyeung: Accepted, and the reasoning answers it: the section is no longer only descriptive — it now ends by telling the reader to use the form to request deletion, and the retention section points here for that. A section that asks the reader to act is a section with a reason to carry the control.

Resolving as a decision made rather than a change requested, which is what I asked for.

_Codex re-review._

tests/unit/content.test.ts
kowinauyeung: Unchanged on the new head, and still live — `site.feedbackFormUrl` is still a `forms.gle` URL, so the biconditional is still being evaluated against a true left-hand side rather than satisfied vacuously.

_Codex re-review._

kowinauyeung:
## Re-review — `dc62403` → `48ff069`

**Verdict: `APPROVE`.** The blocker and both actionable nitpicks are fixed; all five threads resolved.

- **Blocker.** 「**この文字列には**、ユーザー ID、メールアドレス、…は含まれません」. The subject is explicit, so the sentence no longer contradicts `:34` and `:96`, which both still say a reply address may be entered.
- **Retention.** 保存期間と削除 now scopes itself — 「本項でいうデータは、アカウントに紐づけて保管されるもの（単語、単語集、単語集の並び順、練習履歴）を指します」 — then names the form exception and where to ask for deletion. Scoping the term is better than the cross-reference I suggested: it fixes the sentence a reader actually lands on rather than sending them elsewhere to find out it does not mean what it says. The enumeration also matches what `deleteEverything` really removes.
- **The Google Forms setting.** Now a third condition in `site.ts`, beside the two already there, with the part that matters stated plainly: it can be falsified from an admin screen with no commit, no review, and nothing in this repository able to notice. The policy would have made it prose; here it is a note to whoever changes the form.
- **The third link.** Kept, with a reason — the section now ends by asking the reader to use the form, and the retention section points at it for that. A decision rather than an oversight, which is what I asked for.

The unit test is unchanged and still live: `site.feedbackFormUrl` is still a `forms.gle` URL, so the biconditional is being evaluated against a true left-hand side rather than satisfied vacuously.

Nothing outstanding.

---
_Codex re-review._

Issue #26, opened 2026-08-13
upload.mjs still writes allowedUsers and reports it as confirming access


kowinauyeung:
Since #21 the security rules gate on the `allowed` custom claim (`firestore.rules:95`). Nothing reads `allowedUsers` any more — `firestore.rules:61` records that it has no match block at all.

`migration/upload.mjs:154` still writes it, and `:157` prints:

```
許可    : allowedUsers/<uid> を確認しました
```

"Confirmed" — of a collection that no longer authorises anything. `:149` still describes the old mechanism in a comment.

### Why this matters more than dead data usually does

The production migration has not been run yet, and this line is on its path. An operator will be told access is confirmed, open the app, and find every screen reporting a load failure — the same wall #22 exists to make legible, reached by following the tooling's own success message.

`migration/README.md:59` and `:165` describe the same replaced mechanism.

### Suggested fix

Drop the write and the line, and replace both with a check that the account carries the claim — or, if that is more than this script should know about, say plainly that granting access is `yarn allow` followed by signing in again.

**Worth doing before the first production migration**, not after.

kowinauyeung:
Moot, and it was already stale before that.

Two separate things:

1. **The write was fixed by #34**, which stopped `upload.mjs` writing `allowedUsers` and stopped it reporting the collection as confirming access. The production run on 2026-08-17 printed only `許可 : email_verified と allowed クレームを確認しました` — no `allowedUsers` line. So the defect described here had not been reachable for some time, and this issue should have been closed then.
2. **The script is now gone.** `migration/` was retired once production was loaded and verified.

What this issue does **not** cover, and what is still outstanding: the legacy `allowedUsers` documents already sitting in `goitei-dev` have not been deleted. Nothing reads them since #21 and nothing writes them since #34, so they are inert — but they are still there, and removing them now needs a one-off Admin SDK call rather than a flag on a script that no longer exists. Production never had any.

PR #29, opened 2026-08-13
fix(rules): allowlist the fields, not just count them

kowinauyeung:
`keys().size()` counts fields and never looks at their names. So this passed every check in the file:

```json
{ "ownerUid": "…", "headword": "x", "junk": "<a megabyte>" }
```

Two keys, both strings the rules name are within their limits, and `junk` invisible to all of them. The size caps were the only thing between an allowed account and a document as large as Firestore permits, and they never looked at the field that was large. A megabyte under a key nobody validates costs the same as a megabyte under `definition`; only one of them was bounded.

### The change

`hasOnly` on each validator, which is the allowlist Firestore documents for this.

It does **not** make anything required — it constrains the key set to a subset, so an absent optional field stays absent. Requiredness is still in the `d.get(field, default)` calls, and `accepts an entry missing every optional field` was green throughout.

| | before | after |
| --- | --- | --- |
| `validEntry` | `keys().size() <= 40` | `hasOnly(entryFields())` — 28 names |
| `validWordSet` | `keys().size() <= 20` | `hasOnly(wordSetFields())` — 11 |
| `validSession` | `keys().size() <= 20` | `hasOnly(sessionFields())` — 8 |
| `progress/entries` | `keys().size() <= 5` | `hasOnly(['ownerUid', 'entries'])` |

**`users/{uid}` loses `create, update` rather than gaining an allowlist.** Nothing in `src/` writes that document — there is no profile and no adapter addresses the path — so the rule described a feature that does not exist while accepting any twenty fields of any names. It is the one place `hasOnly` cannot help, because there is no legitimate field to allow: the honest allowlist for a document nobody writes is empty. `delete` stays; account deletion needs it.

### `migrationKey`, and why the list is not the TypeScript type

`migration/upload.mjs` writes `migrationKey` onto every migrated document as provenance, and its own comment says it is "absent from the domain `Entry`".

An allowlist built by reading `Entry` would therefore have been complete, reviewable, and wrong: `updateDoc` sends the merged document, so **every edit to all 67 migrated words would have been refused — in production only**, where the migration has been run, and after the rules were already live.

The list is read off what is stored. A test pins that.

### Layer

`tests/rules`, per `CLAUDE.md` — the claim is who may write what. Three cases:

- `refuses a field the schema has no name for, however small` — the defect, on all four paths.
- `refuses to create the user document, which nothing writes`.
- `accepts migrationKey, which the migration writes and the domain type does not name` — including the update path, which is the one that actually breaks.

### Red before green, twice, in different directions

Reverting `hasOnly` to the counts, and restoring `create, update` on `users/{uid}`:

```
× refuses a field the schema has no name for, however small
× refuses to create the user document, which nothing writes
74 passing
```

The `migrationKey` case **stayed green**, which is correct — it is a regression guard, not a defect reproduction, so a revert of the fix cannot fail it. It was proved separately by removing `migrationKey` from `entryFields()`, the exact mistake it exists to catch:

```
× accepts migrationKey, which the migration writes and the domain type does not name
75 passing
```

### Gate

569 unit and component, **76 emulator** (up from 73), lint clean, typecheck clean, prettier clean.

### Not in this change

`publicEntries` and `publicSets` still have no shape validation at all. They are next, and the shape of that change is different — `PublicationRepository` has no adapter and no UI, so the answer there is to deny the paths outright rather than to describe fields nothing writes.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -127,6 +127,55 @@ service cloud.firestore {
     //
     // `headword` is the exception and is read directly: an entry without one is
     // not an entry, and the form refuses it too.
+
+    // ---- which fields, not just how many ----
+    //
+    // `keys().size()` bounds the number of fields and says nothing about their
+    // names, so `{ headword: 'x', junk: <a megabyte> }` passed everything
+    // below: the count is two, every string named is within its limit, and
+    // `junk` is invisible to all of them. The size caps were the only thing
+    // standing between an allowed account and a document as large as Firestore
+    // permits, and they never looked at the field that was large.
+    //
+    // `hasOnly` is the allowlist Firestore documents for this. It does **not**
+    // make anything required — it constrains the key set to a subset, so an
+    // absent optional field stays absent. Requiredness is still where it was,
+    // in the `d.get(field, default)` calls further down, and the tests that
+    // assert an entry missing every optional field is accepted still pass.
+    //
+    // **`migrationKey` is on the entry list and is not in the domain `Entry`.**
+    // `migration/upload.mjs` writes it onto every migrated document as
+    // provenance, and an update sends the merged document — so omitting it here
+    // would refuse every edit to all 67 migrated words, in production only,
+    // where the migration has been run. Read the list off what is stored, not
+    // off the TypeScript type.
+    function entryFields() {
+      return [
+        'ownerUid', 'headword', 'reading', 'pitchAccent', 'pos', 'jlpt',
+        'origin', 'style', 'politeness', 'freq', 'citationForm', 'posInfo',
+        'definition', 'definitionSub', 'senses', 'examples', 'related',
+        'source', 'context', 'usage', 'tags', 'learnedOn',
+        'publishedId', 'publishedVersion', 'copiedFrom',
+        'createdAt', 'updatedAt',
+        'migrationKey'
+      ];
+    }
+
+    function wordSetFields() {
+      return [
+        'ownerUid', 'name', 'description', 'entryIds', 'level', 'topics',
+        'publishedId', 'publishedVersion', 'copiedFrom',
+        'createdAt', 'updatedAt'
+      ];
+    }
+
+    function sessionFields() {
+      return [
+        'ownerUid', 'mode', 'filterLabel', 'total', 'correct', 'missed',
+        'startedAt', 'finishedAt'
+      ];
+    }
+
     function str(value, max) {
       return value is string && value.size() <= max;
     }
@@ -156,7 +205,7 @@ service cloud.firestore {
     function validEntry(uid) {
       let d = request.resource.data;
       return ownedBy(uid)
-        && d.keys().size() <= 40
+        && d.keys().hasOnly(entryFields())
         && str(d.headword, 200) && d.headword.size() > 0
         && str(d.get('reading', ''), 200)
         && str(d.get('definition', ''), 20000)
@@ -174,7 +223,7 @@ service cloud.firestore {
     function validWordSet(uid) {
       let d = request.resource.data;
       return ownedBy(uid)
-        && d.keys().size() <= 20
+        && d.keys().hasOnly(wordSetFields())
         && str(d.name, 200) && d.name.size() > 0
         && str(d.get('description', ''), 5000)
         // The one unbounded list in the schema, and the one a set is made of.
@@ -185,7 +234,7 @@ service cloud.firestore {
     function validSession(uid) {
       let d = request.resource.data;
       return ownedBy(uid)
-        && d.keys().size() <= 20
+        && d.keys().hasOnly(sessionFields())
         && str(d.get('filterLabel', ''), 500)
         && d.get('total', 0) is int && d.get('total', 0) >= 0 && d.get('total', 0) <= 2000
         && d.get('correct', 0) is int && d.get('correct', 0) >= 0
@@ -209,14 +258,23 @@ service cloud.firestore {
         return isOwner(uid) && isAllowed();
       }
 
-      allow read, delete: if mine();
-      // Split exactly as the subcollections are, and for the reason
+      // Read and delete only. **Nothing in the application writes this
+      // document** — there is no profile, and no adapter addresses this path —
+      // so the create and update it used to carry described a feature that does
+      // not exist, while permitting any twenty fields of any names. That is the
+      // one place `hasOnly` cannot help, because there is no legitimate field to
+      // allow: the honest allowlist for a document nobody writes is empty.
+      //
+      // Delete stays, and stays split from the rest for the reason
       // `keepsCreatedAt` documents: `write` covers delete, and on a delete
       // `request.resource` is null, so reaching into it is an evaluation error.
       // This document was the one place that kept the combined form, and
       // deleting it therefore failed with a message about null rather than
       // about permission — the failure mode account deletion would have hit.
-      allow create, update: if mine() && request.resource.data.keys().size() <= 20;
+      //
+      // Adding a profile later means adding `create, update` back **with a
+      // `hasOnly` naming its fields**, not restoring what was here.
+      allow read, delete: if mine();
 
       match /entries/{entryId} {
         allow read, delete: if mine();
@@ -247,7 +305,7 @@ service cloud.firestore {
       match /progress/entries {
         allow read, delete: if mine();
         allow create, update: if mine() && ownedBy(uid)
-                              && request.resource.data.keys().size() <= 5;
+                              && request.resource.data.keys().hasOnly(['ownerUid', 'entries']);
       }
     }
 
--- a/firestore.rules
+++ b/firestore.rules
@@ -133,9 +133,21 @@ service cloud.firestore {
     // `keys().size()` bounds the number of fields and says nothing about their
     // names, so `{ headword: 'x', junk: <a megabyte> }` passed everything
     // below: the count is two, every string named is within its limit, and
-    // `junk` is invisible to all of them. The size caps were the only thing
-    // standing between an allowed account and a document as large as Firestore
-    // permits, and they never looked at the field that was large.
+    // `junk` is invisible to all of them.
+    //
+    // **Being on this list is permission to send a field, not a limit on it.**
+    // `hasOnly` closes the unknown-name half and the other half is closed by
+    // the checks in `validEntry` — which is why the short scalars, the two
+    // timestamps and the maps are all bounded there now rather than merely
+    // named here. An allowlisted-but-unchecked field is the same megabyte under
+    // a different key.
+    //
+    // **What still is not bounded, said plainly rather than left to be found.**
+    // `size()` on a map counts keys and `size()` on a list counts elements, so
+    // neither reaches the bytes inside one: a single `context` value, or one
+    // `senses[3].description`, can be as large as a document may be. Firestore
+    // rules cannot see either. Anything depending on per-element limits has to
+    // enforce them where they are visible.
     //
     // `hasOnly` is the allowlist Firestore documents for this. It does **not**
     // make anything required — it constrains the key set to a subset, so an
@@ -217,7 +229,33 @@ service cloud.firestore {
         && list(d.get('senses', []), 100)
         && list(d.get('examples', []), 100)
         && list(d.get('related', []), 100)
-        && (d.get('pitchAccent', null) == null || d.get('pitchAccent', 0) is int);
+        && (d.get('pitchAccent', null) == null || d.get('pitchAccent', 0) is int)
+        // Short, closed-vocabulary scalars. Each is one line and each was
+        // previously unbounded, which `hasOnly` alone does not change: being on
+        // the allowlist is permission to send the field, not a limit on it.
+        && str(d.get('jlpt', ''), 10)
+        && str(d.get('origin', ''), 10)
+        && str(d.get('style', ''), 10)
+        && str(d.get('politeness', ''), 10)
+        && str(d.get('learnedOn', ''), 10)
+        && str(d.get('migrationKey', ''), 200)
+        && (d.get('publishedId', null) == null || str(d.get('publishedId', ''), 200))
+        && d.get('freq', 1) is int && d.get('freq', 1) >= 1 && d.get('freq', 1) <= 5
+        && d.get('publishedVersion', 0) is int && d.get('publishedVersion', 0) >= 0
+        // Timestamps, not strings. Both are written by the adapter from the
+        // server clock, so anything else on this key is a client sending a
+        // value of its own choosing — and before this line that value could be
+        // a string of any length. `keepsCreatedAt` pins createdAt to what is
+        // stored, which says nothing about the create that stored it.
+        && d.get('createdAt', request.time) is timestamp
+        && d.get('updatedAt', request.time) is timestamp
+        // Maps. `size()` counts keys, so this stops a key explosion and **not**
+        // a large value inside one — the same blind spot the note above records
+        // for list elements. Said plainly rather than left to be discovered.
+        && d.get('context', {}).size() <= 20
+        && d.get('usage', {}).size() <= 20
+        && (d.get('posInfo', null) == null || d.get('posInfo', {}).size() <= 20)
+        && (d.get('copiedFrom', null) == null || d.get('copiedFrom', {}).size() <= 10);
     }
 
     function validWordSet(uid) {
@@ -304,8 +342,13 @@ service cloud.firestore {
       // every sibling checks the field as well.
       match /progress/entries {
         allow read, delete: if mine();
+        // `entries` is the whole content of this document and `hasOnly` does not
+        // touch it, so before this bound the only ceiling on one account's
+        // progress map was Firestore's 1 MiB. 5,000 words is far past any
+        // notebook this design expects and far short of that ceiling.
         allow create, update: if mine() && ownedBy(uid)
-                              && request.resource.data.keys().hasOnly(['ownerUid', 'entries']);
+                              && request.resource.data.keys().hasOnly(['ownerUid', 'entries'])
+                              && request.resource.data.get('entries', {}).size() <= 5000;
       }
     }
 
--- a/tests/rules/firestore-rules.test.ts
+++ b/tests/rules/firestore-rules.test.ts
@@ -6,7 +6,9 @@ import {
   type RulesTestEnvironment,
 } from '@firebase/rules-unit-testing';
 import { afterAll, beforeAll, beforeEach, describe, it } from 'vitest';
+import type { PracticeSessionDraft } from '@/domain/practice';
 import { emptyDraft } from '@/lib/draft';
+import { makeWordSet } from '../fixtures/wordSet';
 
 /**
  * Security rules are the only thing protecting this data: the app is pure
@@ -379,9 +381,13 @@ describe('bounds on what an owner may write', () => {
   it('refuses to rewrite createdAt on an existing row', async () => {
     const db = as(ALICE);
     await assertSucceeds(
-      db.doc(`users/${ALICE}/entries/dated`).set(entry(ALICE, { createdAt: '2026-01-01' })),
+      db
+        .doc(`users/${ALICE}/entries/dated`)
+        .set(entry(ALICE, { createdAt: new Date('2026-01-01') })),
     );
-    await assertFails(db.doc(`users/${ALICE}/entries/dated`).update({ createdAt: '2020-01-01' }));
+    await assertFails(
+      db.doc(`users/${ALICE}/entries/dated`).update({ createdAt: new Date('2020-01-01') }),
+    );
     await assertSucceeds(db.doc(`users/${ALICE}/entries/dated`).update({ headword: '別の語' }));
   });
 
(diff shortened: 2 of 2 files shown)


firestore.rules
```diff
@@ -156,7 +205,7 @@ service cloud.firestore {
     function validEntry(uid) {
       let d = request.resource.data;
       return ownedBy(uid)
-        && d.keys().size() <= 40
+        && d.keys().hasOnly(entryFields())
```
kowinauyeung, line 220: 🟡 **SUGGESTION** — this closes the unknown-name half of the hole. The other half is unchanged, and the description reads as though both are gone.

The framing in the description is 「a megabyte under a key nobody validates costs the same as a megabyte under `definition`; only one of them was bounded」. That sentence is still true after this change for 15 of the 28 names now on the list, because being on the list is not the same as being validated. `validEntry` checks `headword`, `reading`, `definition`, `definitionSub`, `source`, `citationForm`, `pos`, `tags`, `senses`, `examples`, `related`, `pitchAccent` and `ownerUid`. It says nothing about:

`jlpt`, `origin`, `style`, `politeness`, `freq`, `posInfo`, `context`, `usage`, `learnedOn`, `publishedId`, `publishedVersion`, `copiedFrom`, `createdAt`, `updatedAt`, `migrationKey`

So `{ ownerUid, headword: 'x', definition: '', context: <a megabyte> }` passes `validEntry` for the same reason `junk` used to — nothing looks at the field that is large. `migrationKey` is worth calling out because this change is what put it on the list: it is now an explicitly permitted, entirely unvalidated string. Read off the rules rather than measured against the emulator; the reasoning is just which names appear in the conjunction.

None of this is a regression — every one of those fields was equally unbounded under `keys().size() <= 40`, and `hasOnly` is strictly stronger than what it replaces. The problem is what the file will read like to the next person. The comment above says the size caps 「were the only thing standing between an allowed account and a document as large as Firestore permits」, in the past tense, immediately above rules under which that is still the case for half the schema.

Two honest ways out, either is fine:

- Bound the cheap ones here. `str(d.get('migrationKey', ''), 200)`, `str(d.get('learnedOn', ''), 10)`, `str(d.get('jlpt', ''), 10)`, `str(d.get('origin', ''), 10)`, `str(d.get('style', ''), 10)`, `str(d.get('politeness', ''), 10)` are one line each and cover most of the surface. `context`, `usage` and `posInfo` are maps and need `.size()` on the map rather than a string cap, which is a larger call.
- Or say the scope out loud — that this change rejects unknown fields and that bounding the known-but-unchecked ones is next — in the same register the block below already uses for `publicEntries`. That block is a good model: it names what is missing instead of leaving it to be discovered.

_Codex review._

firestore.rules
```diff
@@ -247,7 +305,7 @@ service cloud.firestore {
       match /progress/entries {
         allow read, delete: if mine();
         allow create, update: if mine() && ownedBy(uid)
-                              && request.resource.data.keys().size() <= 5;
+                              && request.resource.data.keys().hasOnly(['ownerUid', 'entries']);
```
kowinauyeung: 🟡 **SUGGESTION**, the same point where it bites hardest.

`hasOnly(['ownerUid', 'entries'])` is a real improvement over `keys().size() <= 5` — it stops `progress/junk` fields outright. But `entries` is a map with no bound on how many keys it holds or how large they are, and it is the only field of substance on the document, so what an allowed account can put in this one document is unchanged by this line.

That matters more here than on an entry, because there is exactly one of these documents per account and the ceiling on it is Firestore's 1 MiB rather than anything in this file. `ProgressRepository` already documents accepting that ceiling deliberately, so this is not new — but `d.get('entries', {}).size() <= 5000` is available in rules and would turn an accepted design limit into an enforced one.

_Codex review._

firestore.rules
```diff
@@ -174,7 +223,7 @@ service cloud.firestore {
     function validWordSet(uid) {
       let d = request.resource.data;
       return ownedBy(uid)
-        && d.keys().size() <= 20
+        && d.keys().hasOnly(wordSetFields())
```
kowinauyeung, line 264: 🟡 **SUGGESTION** — the drift guard that protects `entryFields()` does not extend to this list or to `sessionFields()`.

The reason the entry allowlist is safe against a future field is `tests/rules/firestore-rules.test.ts`, in `what the adapters write is what the rules accept`: it builds its payload from `emptyDraft()`, so adding a field to `EntryDraft` and forgetting it here turns `accepts the document entryRepo.create sends` red. That is the single most valuable thing in this change, and it is worth being explicit that it is what makes `hasOnly` a safe move on entries at all.

The word-set and session cases in the same suite are hand-written object literals — `:486` says so directly (「The literal WordSets.tsx sends; there is no factory for it」), and the session payload at `:594` is the same shape. So the exact failure this change introduces, a domain field added and not allowlisted, is caught for entries and not for the other two. It would pass here, pass in review, and be refused only in production — which is precisely the failure mode the `migrationKey` section of the description is about.

`tests/fixtures/` already has `makeWordSet()`. Routing the word-set payload through `toDraft(makeWordSet())` and the session payload through a `PracticeSessionDraft`-typed factory would put all three on the same footing, and `CLAUDE.md` explicitly permits tests to import production constants for this reason — 「duplicating them is how a test keeps passing after the schema moves」.

_Codex review._

firestore.rules
```diff
@@ -127,6 +127,55 @@ service cloud.firestore {
...
+    // in the `d.get(field, default)` calls further down, and the tests that
+    // assert an entry missing every optional field is accepted still pass.
+    //
+    // **`migrationKey` is on the entry list and is not in the domain `Entry`.**
+    // `migration/upload.mjs` writes it onto every migrated document as
+    // provenance, and an update sends the merged document — so omitting it here
+    // would refuse every edit to all 67 migrated words, in production only,
+    // where the migration has been run. Read the list off what is stored, not
+    // off the TypeScript type.
+    function entryFields() {
```
kowinauyeung, line 164: ✅ **PRAISE** — `migrationKey` is the part of this change that is hard to get right, and the reasoning for it is correct.

An allowlist built by reading `Entry` would have been complete, reviewable, and would have refused every edit to all 67 migrated words — green in the emulator, green in review, and broken only where the migration has been run. Reading the list off what is stored rather than off the type is the right instinct, and confirming it independently:

- `migration/upload.mjs` writes `migrationKey` on every document in the batch, and `updateDoc` sends the merged document, so the update path is the one that breaks. The test names that.
- The union of keys across all 67 entries in `migration/output.json` is `id, createdAt, updatedAt, wordSets` — all stripped by `upload.mjs` — plus 21 names, every one of which is on this list.
- The list is otherwise `Entry` minus `id`, which is the document id.

The note about the test staying green on a revert is right too, and worth having written down: a regression guard is not a defect reproduction, so proving it separately by removing `migrationKey` from this list — the exact mistake it exists to catch — is the only thing that could have proved it at all.

_Codex review._

firestore.rules
```diff
@@ -209,14 +258,23 @@ service cloud.firestore {
...
       // `keepsCreatedAt` documents: `write` covers delete, and on a delete
       // `request.resource` is null, so reaching into it is an evaluation error.
       // This document was the one place that kept the combined form, and
       // deleting it therefore failed with a message about null rather than
       // about permission — the failure mode account deletion would have hit.
-      allow create, update: if mine() && request.resource.data.keys().size() <= 20;
+      //
+      // Adding a profile later means adding `create, update` back **with a
+      // `hasOnly` naming its fields**, not restoring what was here.
+      allow read, delete: if mine();
```
kowinauyeung, line 315: ✅ **PRAISE** — 「the honest allowlist for a document nobody writes is empty」 is the right answer, and removing the permission is better than describing fields that do not exist.

Checked independently that nothing writes this document:

- The three Firestore adapters address `entries`, `wordSets`, `practiceSessions` and `progress/entries` and never the parent.
- `UserRepository` is a port with no adapter; `accountData.ts` says as much where it takes `profile` as a parameter rather than through a port.
- `migration/upload.mjs:155` does write a document keyed by uid, but it is `allowedUsers/{uid}`, not `users/{uid}` — and it runs through the Admin SDK, which bypasses rules regardless.

`delete` staying, and staying split from the rest for the `keepsCreatedAt` reason, is correct: account deletion needs it, and `request.resource` is null on a delete.

_Codex review._

kowinauyeung:
## Review summary

**Verdict: `APPROVE`** — no blocker. Two suggestions, both about what the change does *not* cover rather than about what it does.

`keys().size()` counting fields and never reading their names is a real hole, `hasOnly` is the right instrument, and the note that it constrains rather than requires — so an absent optional field stays absent and requiredness stays in the `d.get(field, default)` calls — is the thing most likely to be got wrong here and is got right.

### Allowlist completeness, checked independently

This is the half of the change that fails in production or not at all, so it was worth confirming rather than reading:

| List | Checked against | Result |
| --- | --- | --- |
| `entryFields()` | `Entry` minus `id`, plus `migrationKey` | 28 names, complete |
| | union of keys across all 67 entries in `migration/output.json`, minus what `upload.mjs` strips (`id`, `createdAt`, `updatedAt`, `wordSets`) | every name present |
| `wordSetFields()` | `WordSet` minus `id` | 11 names, complete |
| `sessionFields()` | `PracticeSessionDraft` + `ownerUid` + `finishedAt` | 8 names, complete |
| `progress/entries` | `progressRepo.recordSession`, which merges `{ ownerUid, entries }` | complete |

`users/{uid}` losing `create, update` breaks nothing: the three adapters address the subcollections and never the parent, `UserRepository` has no adapter, and the one uid-keyed document the migration writes is `allowedUsers/{uid}` — a different path, and through the Admin SDK, which bypasses rules anyway.

### The `migrationKey` reasoning is the best thing here

An allowlist built by reading `Entry` would have been complete, reviewable, green in the emulator and green in review, and would have refused every edit to all 67 migrated words in production only. Reading the list off what is stored is the correct instinct and the test pins it. The note that the test stays green on a revert of the fix — because a regression guard is not a defect reproduction, so it had to be proved by removing `migrationKey` from the list instead — is exactly the distinction `CLAUDE.md` asks for, made without being prompted.

### What is not closed

**Being on the allowlist is not the same as being validated.** 15 of the 28 entry names have no type or size check: `jlpt`, `origin`, `style`, `politeness`, `freq`, `posInfo`, `context`, `usage`, `learnedOn`, `publishedId`, `publishedVersion`, `copiedFrom`, `createdAt`, `updatedAt`, `migrationKey`. So `{ ownerUid, headword: 'x', definition: '', context: <a megabyte> }` passes `validEntry` for the same reason `junk` used to. `migrationKey` is worth singling out because this change is what put it there — a newly permitted, entirely unvalidated string.

That is not a regression; all fifteen were equally unbounded under `keys().size() <= 40`, and `hasOnly` is strictly stronger than what it replaces. It is about what the file will read like next: the comment says the size caps "were the only thing standing between an allowed account and a document as large as Firestore permits", in the past tense, directly above rules under which that is still true for half the schema. Either bound the cheap ones — six `str(d.get(…), n)` lines cover most of it — or say the scope out loud, the way the `publicEntries` block below already does so well.

`progress/entries` is the same point at its sharpest: `hasOnly(['ownerUid', 'entries'])` stops stray field names, and `entries` is an unbounded map that is the entire content of the document.

**The drift guard covers entries only.** `what the adapters write is what the rules accept` builds its entry payload from `emptyDraft()`, so a field added to `EntryDraft` and forgotten in `entryFields()` turns that test red — which is what makes `hasOnly` a safe move on entries at all, and is worth stating outright. The word-set and session cases in the same suite are hand-written literals (`:486` says so; `:594` is the same), so the identical mistake on those two lists would pass here and be refused only in production. `tests/fixtures/makeWordSet()` already exists.

### Layer and gate

`tests/rules` is the right layer under `CLAUDE.md` — the claim is who may write what. Tests were not run here; the counts in the description (569 unit and component, 76 emulator) are taken as stated. The allowlist table above was derived by reading the domain types, the adapters and the migration, and by taking the key union over `migration/output.json` directly.

---
_Codex review._

firestore.rules
kowinauyeung: @kowinauyeung
Upheld on both halves — the fifteen unchecked names, and the past tense above them claiming otherwise.

Bounded rather than only documented, because the six lines were as cheap as you said and the remaining scope is small enough to state honestly:

```
str(d.get('jlpt', ''), 10)  ·  origin  ·  style  ·  politeness  ·  learnedOn
str(d.get('migrationKey', ''), 200)
d.get('publishedId', null) == null || str(…, 200)
d.get('freq', 1) is int && >= 1 && <= 5      ·  publishedVersion is int && >= 0
d.get('context', {}).size() <= 20            ·  usage  ·  posInfo  ·  copiedFrom
```

**`createdAt` and `updatedAt` turned out to be the sharpest of the fifteen**, and are now `is timestamp`. `keepsCreatedAt` pins `createdAt` to the stored value on update and says nothing about the create that stored it — so a new entry could carry a string of any length on the one key the rules treat as bookkeeping. Both are written by the adapter from the server clock, so anything else there is a client choosing its own value.

**That check found something.** `accepts the document entryRepo.create sends` went red:

```ts
createdAt: '2026-08-13T00:00:00.000Z',   // a string
```

`entryRepo.create` writes `serverTimestamp()`, which lands as a Timestamp. The test is named *what the adapters write* is what the rules accept, and the two stamps were the one part of the payload hand-written in a shape the adapter never produces. Fixed to `new Date(…)`.

**Scope stated where you suggested, in the register the `publicEntries` block uses.** The comment now says that being on the list is permission to send a field rather than a limit on it, and names what is still unbounded: `size()` on a map counts keys and on a list counts elements, so neither reaches the bytes inside one. A single `context` value, or one `senses[3].description`, can still be as large as a document may be. Rules cannot see either.

Red before green, with `hasOnly` left in place, so these test the bounds and not the allowlist:

```
× refuses an oversized value in a field the allowlist permits
× refuses a progress map holding more words than a notebook has
76 passing
```

78 emulator, 569 unit and component.

Fixed in 373670d

firestore.rules
Agreed — and "the same point where it bites hardest" is the right description, because there is one of these documents per account and `entries` is the whole of it.

```
&& request.resource.data.get('entries', {}).size() <= 5000
```

5,000 words is far past any notebook this design expects and far short of the 1 MiB ceiling, which is what turns an accepted design limit into an enforced one — as you put it.

`ProgressRepository` documents accepting that ceiling deliberately and still does; what it did not document is that nothing enforced anything below it.

`refuses a progress map holding more words than a notebook has` covers it, and goes red with the bound removed and `hasOnly` still in place.

Fixed in 373670d

firestore.rules
kowinauyeung: @kowinauyeung
Agreed, and this is the one of the three that fails in production or not at all — which is the same shape as the `migrationKey` problem, arriving from the other side.

Both payloads now come from something typed, and the two mechanisms differ on purpose:

- **Word set** — `makeWordSet({ ownerUid: ALICE })`, destructured to the stored document. It is typed `WordSet`, so a new domain field arrives in the payload on its own and the emulator refuses it until it is allowlisted.
- **Session** — the literal is annotated `PracticeSessionDraft` rather than routed through a factory, since none exists. A field added to the draft makes it a **type error**, so that drift is caught at `yarn typecheck` — earlier and cheaper than a rules run.

Verified the guard is live rather than assumed, by removing one name from `wordSetFields()`:

```
× accepts the document wordSetRepo.create sends
× accepts the whole-draft update wordSetRepo.update sends over a stored set
76 passing
```

The second one is worth noting: routing the create through the factory tightened the update case too, since it reads from the same source.

Fixed in 373670d

firestore.rules
kowinauyeung: @kowinauyeung
Thank you for taking the key union over `migration/output.json` rather than reading the list against `Entry` — that is the check that would have caught me if I had got it wrong, and it is not the obvious one to run.

Recording one thing your table makes visible that I had not stated: the union being `Entry` minus `id` plus `migrationKey` means the allowlist is now pinned by two independent things — the type, through `emptyDraft()` in the adapter-fidelity test, and the data, through the `migrationKey` case. Neither alone would have been enough.

firestore.rules
kowinauyeung: @kowinauyeung
Thanks — and specifically for checking `migration/upload.mjs:155` rather than assuming. That it writes `allowedUsers/{uid}` and not `users/{uid}` is exactly the near-miss that would have made removing `create` look wrong, and it is one grep away from being read the other way.

Worth adding for the record: that write is itself on the way out. Nothing has read `allowedUsers` since the gate became a custom claim, and the same script still prints 「許可 : allowedUsers/… を確認しました」, which will tell an operator access is confirmed and then leave every screen reporting a load failure. Tracked separately as #26, due before the first production migration.

firestore.rules
kowinauyeung: **Verified fixed**, and this one I ran rather than read, because the fix had a way of being wrong that would only have shown in production.

Six of the new bounds are 10-character caps on closed-vocabulary Japanese fields, and the values in `migration/output.json` are 4 characters but **12 UTF-8 bytes**:

| field | widest stored value | characters | UTF-8 bytes |
| --- | --- | --- | --- |
| `jlpt` | `レベル外` | 4 | 12 |
| `style` | `話し言葉` | 4 | 12 |
| `politeness` | `くだけた` | 4 | 12 |
| `origin` | `外来語` | 3 | 9 |

If `String.size()` counted bytes, `str(d.get('jlpt', ''), 10)` would refuse every entry filed as `レベル外` — and `emptyDraft()` uses it as the default, so it would be every new entry too.

It counts characters. Ran `yarn test:emulator` on `373670d` against JDK 21: **78 passed**, including `accepts the document entryRepo.create sends`, which spreads `emptyDraft()` and therefore writes `jlpt: 'レベル外'` through the new cap. Three further full runs, all 78 green.

The rest checks out against the stored data: `freq` is 2–5 within the new 1–5 bound, `learnedOn` is exactly 10 characters against the 10 cap, `migrationKey` tops out at 15 against 200, and `context`/`usage`/`posInfo` hold 3/3/2 keys against 20.

The two `is timestamp` checks are the sharpest addition and the reasoning is right — `keepsCreatedAt` pins the value on update and says nothing about the create that stored it. One narrow gap worth knowing rather than acting on: no test writes through the real adapter *with rules enforced*, so the `serverTimestamp()` sentinel itself is never evaluated against `is timestamp`. The rules tests hand-build payloads and `tests/integration` bypasses rules by design. Firestore resolves the sentinel before rule evaluation, so this is standard behaviour rather than a risk — but it is the one part of the claim the suite cannot see.

Rewriting the paragraph above to say what is *still* unbounded — `size()` on a map counts keys and on a list counts elements, so neither reaches the bytes inside one — is the part I would have kept if I could keep only one thing.

_Codex re-review._

firestore.rules
kowinauyeung: **Verified fixed.** `request.resource.data.get('entries', {}).size() <= 5000`, with `refuses a progress map holding more words than a notebook has` building a 5,001-key map and asserting the refusal. Green in the run above.

The comment gives the number a reason — far past any notebook this design expects, far short of the 1 MiB ceiling — which is what stops it being an arbitrary constant somebody later raises without knowing what it was protecting.

_Codex re-review._

firestore.rules
kowinauyeung: **Verified fixed**, and by two different mechanisms, which is the right call because the two payloads fail differently.

The word set now goes through `makeWordSet()` and destructures off what the adapter fills, so a field added to `WordSet` arrives in the payload on its own and turns the test red at the emulator. The session is annotated `PracticeSessionDraft`, so a field added there is a **type error** — caught at `yarn typecheck`, before the emulator is even started.

Both were also brought onto real `Date` values instead of ISO strings. That was not in my comment and it is a real correction: the stamps were the one part of those payloads written in a shape `serverTimestamp()` never produces, so a test named 「what the adapters write is what the rules accept」 was not sending what the adapter sends. It matters more now that `createdAt` and `updatedAt` are checked with `is timestamp` — as ISO strings those assertions would have been failing for the right reason and passing for the wrong one.

All three drift guards now sit on the same footing as the entry one.

_Codex re-review._

firestore.rules
kowinauyeung: Still holds. `migrationKey` is now bounded as well as permitted — `str(d.get('migrationKey', ''), 200)`, against a stored maximum of 15 characters — so it is no longer the one allowlisted-but-unchecked field the earlier comment singled out.

_Codex re-review._

firestore.rules
kowinauyeung: Unchanged on the new head and still correct. Nothing in `src/` writes `users/{uid}`, and the migration's uid-keyed write is still `allowedUsers/{uid}` through the Admin SDK.

_Codex re-review._

kowinauyeung:
## Re-review — `986156c` → `373670d`

**Verdict: `APPROVE`.** All three suggestions fixed; all five threads resolved. This one I ran rather than read, because the fix had a failure mode that would only have appeared in production.

### The characters-versus-bytes question, settled

Six of the new bounds are 10-character caps on closed-vocabulary Japanese fields, and the stored values are 4 characters but **12 UTF-8 bytes**:

| field | widest stored value | characters | UTF-8 bytes |
| --- | --- | --- | --- |
| `jlpt` | `レベル外` | 4 | 12 |
| `style` | `話し言葉` | 4 | 12 |
| `politeness` | `くだけた` | 4 | 12 |
| `origin` | `外来語` | 3 | 9 |

Had `String.size()` counted bytes, `str(d.get('jlpt', ''), 10)` would have refused every entry filed as `レベル外` — and since `emptyDraft()` uses it as the default, every new entry as well.

It counts characters. `yarn test:emulator` on `373670d` under JDK 21: **78 passed**, including `accepts the document entryRepo.create sends`, which spreads `emptyDraft()` and so writes `jlpt: 'レベル外'` through the new cap. Three further full runs, all 78 green.

The rest agrees with the stored data: `freq` is 2–5 inside the new 1–5 bound, `learnedOn` is exactly 10 characters against the 10 cap, `migrationKey` tops out at 15 against 200, and `context`/`usage`/`posInfo` carry 3/3/2 keys against 20.

### The rest

- **The fifteen unchecked fields are bounded**, and the paragraph that claimed otherwise in the past tense now says what is still not covered: `size()` on a map counts keys and on a list counts elements, so neither reaches the bytes inside one. Naming the remaining blind spot is better than silently half-closing the hole.
- **`progress.entries` is capped at 5,000**, with a test that builds a 5,001-key map, and a comment giving the number a reason rather than leaving it an arbitrary constant.
- **The drift guards now cover all three payloads**, by two mechanisms suited to each: the word set through `makeWordSet()` so a new domain field turns the emulator test red, the session annotated `PracticeSessionDraft` so a new field is a *type* error at `yarn typecheck`. Both were also moved onto real `Date` values — not something I raised, and a real correction: the stamps were the one part of those payloads written in a shape `serverTimestamp()` never produces, in tests named 「what the adapters write is what the rules accept」.

### Two things to know, neither blocking

**The `serverTimestamp()` sentinel is never evaluated against the new `is timestamp` checks.** The rules tests hand-build payloads and `tests/integration` bypasses rules by design, so nothing writes through the real adapter with rules enforced. Firestore resolves the sentinel before rule evaluation, so this is standard behaviour rather than a risk — but it is the one part of the claim the suite cannot see.

**There is an intermittent failure in the suite, and it is not this pull request's.** On the first of four full runs:

```
× tests/integration/entryRepo.test.ts > list — pagination > orders newest first
  expected [ '二番目', '三番目', '一番目' ] to deeply equal [ '三番目', '二番目', '一番目' ]
```

Green on the next three full runs and on five isolated runs. That file states 「Rules are deliberately bypassed here (`mockUserToken: 'owner'`)」, so a rules change cannot reach it. The mechanism looks like three sequential `serverTimestamp()` writes landing in the same millisecond under full-suite load, leaving `orderBy(documentId(), 'desc')` to break the tie by random auto-id — which is the same collision the test two below it is built around, whose comment says 「serverTimestamp() cannot be made to collide」. Worth its own issue; nothing to do here.

---
_Codex re-review._

PR #30, opened 2026-08-13
fix(rules): close the publishing paths until publishing exists

kowinauyeung:
`publicEntries` and `publicSets` carried create, update and delete with **no shape or size validation of any kind** — no `hasOnly`, no length caps, nothing. `PublicationRepository` is a port in `src/domain/ports.ts` with no adapter behind it and no interface in front of it, so nothing in the application has ever written to either.

That combination is the worst way round: the only collections an allowed account could write anything at all to were the two nothing writes.

### Deleted rather than disabled

Since the recursive wildcard went, an unlisted path is denied. So closing these is removing the blocks, not adding `if false` — and `setPath()` and `ownsParentSet()` go with them, having no other caller.

### Two things this also fixes early

Both would have been wrong on the day publishing shipped, and both are far cheaper now:

- **`deleteEverything()` does not touch published copies.** The privacy policy promises account deletion removes the user's data; that promise would have been false for precisely the data other people can see.
- **`exportEverything()` does not include them either.**

### The design is kept, as properties rather than as rules

The rules file keeps a block naming what has to come back **with** the feature, and the new `describe` docblock points at it:

- `ownerUid` never reassigned on an existing copy, never somebody else's on a new one.
- The parent set checked with `getAfter`, not `get` — a first publish writes the set and its first entries in one batch, and `get` sees only committed state, so it would deny every first publish.
- Delete on a snapshot entry deliberately exempt from that parent check, so an orphan from a half-finished unpublish stays cleanable by its owner.
- Every mutation gated on `isAllowed()`, not only create.
- `hasOnly` validation, which never existed here.
- Deletion and export covering published copies before the feature is reachable.

### One test changed direction instead of being deleted

`revokes update and delete on already-published copies` protected a real gap the first review round found: revoking an account has to take away update and delete on what it already wrote, not only create.

The path is gone; the property is not. It is now `takes away update and delete on what a revoked account already wrote`, against the private notebook. Not a duplicate of the existing stranger test, which writes to a path holding no document — this one edits and deletes an existing row.

### Layer

`tests/rules`. Four cases replacing eleven: read, create, update-or-delete, and the one-batch first publish. The batch case is named separately because it is the shape most likely to slip past a rule written only against single writes.

**Reads are denied to the owner too, and that is the point.** Ownership was the entire basis of the rules being removed, so a check that let the owner through would leave the surface exactly as wide as it was. The seeded documents are written with rules disabled — as a partly-built feature or an earlier deploy would have left them — so being unreadable is a fact about the rules and not about the fixture.

### Red before green

`git checkout develop -- firestore.rules`, restoring the publishing rules exactly:

```
× refuses to read a published copy, to its owner as much as to anyone else
× refuses to create one, which is the write nothing in the app makes
× refuses to update or delete one, including by its own owner
× refuses the one-batch first publish the old getAfter check existed for
61 passing
```

569 unit and component, 65 emulator, lint clean, typecheck clean, prettier clean. 59 lines net removed.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -251,8 +251,28 @@ service cloud.firestore {
     //  * `exportEverything()` does not include them either.
     //
     // **What has to come back with the feature**, restored as properties rather
-    // than as rules — the tests that covered them are named in
-    // `tests/rules/firestore-rules.test.ts` where this block used to be tested:
+    // than as rules. The twelve tests that covered them were removed with the
+    // rules and are not recoverable from this list — they are in pull request
+    // #30, which is where to read them, and their names are the specification.
+    //
+    // Reads first, because they are the access model rather than a detail:
+    //
+    //  * **A published copy is readable by any allowed account, not only its
+    //    owner.** That is the whole point of publishing and the first decision
+    //    somebody rebuilding this has to make; the old rule was
+    //    `allow read: if isAllowed()` on both collections.
+    //  * **A snapshot entry is readable only while its parent set exists** —
+    //    `allow read: if isAllowed() && exists(setPath(publicSetId))`. Firestore
+    //    does not cascade deletes, so without it an unpublish that fails
+    //    part-way leaves a snapshot readable from a path somebody already noted.
+    //    This is the half of the orphan story that fails **open**; the delete
+    //    exemption below is the half that fails closed.
+    //  * **The owner must be able to publish, republish and unpublish.** A list
+    //    of prohibitions with nothing that must succeed is how rules nobody can
+    //    publish through get written — which is exactly what the `getAfter`
+    //    point below exists to prevent.
+    //
+    // Then the writes:
     //
     //  * ownerUid may never be reassigned on an existing copy, and may never be
     //    somebody else's on a new one.
@@ -266,7 +286,12 @@ service cloud.firestore {
     //    account must lose the ability to edit what it has already put in front
     //    of other people.
     //  * Shape and size validation, which never existed here and must, on the
-    //    same `hasOnly` footing as the private collections above.
+    //    same footing as the private collections above — which by then means
+    //    `hasOnly` plus per-field bounds. That arrives in #29, not here: this
+    //    branch still bounds the private collections with `keys().size()`, and
+    //    the intended order is #29 first. The two merge cleanly either way; it
+    //    is this sentence that stops being true if they land the other way
+    //    round.
     //  * Deletion and export have to cover published copies before the feature
     //    is reachable, not after.
 
--- a/firestore.rules
+++ b/firestore.rules
@@ -1,8 +1,9 @@
 rules_version = '2';
 
 // Private data lives under users/{uid} and is reachable by that account alone.
-// Published copies are separate, deliberately duplicated documents in their own
-// top-level collections.
+// **That is currently all of it.** Publishing is designed and unbuilt, and the
+// collections it would use are closed rather than waiting open — see the block
+// further down for what has to come back with the feature.
 //
 // Two properties are worth stating because they are what make this simple:
 //
@@ -22,29 +23,6 @@ service cloud.firestore {
 
     function isOwner(uid) {
       return request.auth != null && request.auth.uid == uid;
-    }
-
-    function setPath(publicSetId) {
-      return /databases/$(database)/documents/publicSets/$(publicSetId);
-    }
-
-    // Carrying ownerUid on a snapshot entry answers "who wrote this row", but
-    // not "may this row be in this set". Without checking the parent, any
-    // allowlisted account could create an entry under somebody else's published
-    // set — or under a set id that does not exist — and it would be served as
-    // part of that set to everyone reading or adopting it.
-    //
-    // getAfter, not get: a first publish writes the set and its first entries in
-    // one batch, and get() only sees committed state, so the child would be
-    // evaluated against a parent that does not exist yet and every first publish
-    // would be denied. getAfter evaluates the batch's end state, which is what
-    // "does this row belong here once this write lands" actually means. It still
-    // returns null for a set nobody is creating, so an entry under an unrelated
-    // or non-existent set stays denied.
-    //
-    // One document read per call, paid only on publish.
-    function ownsParentSet(publicSetId) {
-      return getAfter(setPath(publicSetId)).data.ownerUid == request.auth.uid;
     }
 
     // The single gate that keeps this a private notebook.
@@ -251,49 +229,46 @@ service cloud.firestore {
       }
     }
 
-    // A published copy is an immutable snapshot. Republishing overwrites it
-    // wholesale and bumps its version; unpublishing deletes it. Only the owner
-    // writes, and ownerUid can never be reassigned to somebody else.
+    // ---- publishing: closed until it exists ----
     //
-    // Every mutation is gated on isAllowed(), not just create. Removing an
-    // account from allowedUsers has to revoke it completely — otherwise a
-    // revoked account keeps the ability to edit what it has already put in front
-    // of other people, which is the one thing revocation is for. The cost is
-    // that it can no longer delete its own published copies either; admin
-    // tooling cleans those up, which is the right way round for a ban.
-    match /publicEntries/{publicEntryId} {
-      allow read: if isAllowed();
-      allow create: if isAllowed() && isOwner(request.resource.data.ownerUid);
-      allow update: if isAllowed() && isOwner(resource.data.ownerUid)
-                    && request.resource.data.ownerUid == resource.data.ownerUid;
-      allow delete: if isAllowed() && isOwner(resource.data.ownerUid);
-    }
-
-    match /publicSets/{publicSetId} {
-      allow read: if isAllowed();
-      allow create: if isAllowed() && isOwner(request.resource.data.ownerUid);
-      allow update: if isAllowed() && isOwner(resource.data.ownerUid)
-                    && request.resource.data.ownerUid == resource.data.ownerUid;
-      allow delete: if isAllowed() && isOwner(resource.data.ownerUid);
-
-      match /entries/{entryId} {
-        // Deleting publicSets/{id} does NOT delete this subcollection —
-        // Firestore has no cascading delete. Requiring the parent to exist means
-        // an unpublish that fails part-way through cannot leave a snapshot
-        // readable from a path somebody already knows. PublicationRepository
-        // still has to delete the children first; this is the backstop, not the
-        // mechanism.
-        allow read: if isAllowed() && exists(setPath(publicSetId));
-        allow create: if isAllowed() && isOwner(request.resource.data.ownerUid)
-                      && ownsParentSet(publicSetId);
-        allow update: if isAllowed() && isOwner(resource.data.ownerUid)
-                      && request.resource.data.ownerUid == resource.data.ownerUid
-                      && ownsParentSet(publicSetId);
-        // Deliberately no parent check: orphans left by a half-finished
-        // unpublish must stay deletable by their owner.
-        allow delete: if isAllowed() && isOwner(resource.data.ownerUid);
-      }
-    }
+    // `publicEntries` and `publicSets` had full create/update/delete rules and
+    // **no shape or size validation of any kind** — no `hasOnly`, no length
+    // caps, nothing. Meanwhile `PublicationRepository` is a port in
+    // `src/domain/ports.ts` with no adapter behind it and no interface in front
+    // of it, so nothing in the application has ever written to either.
+    //
+    // That combination is the worst way round. The paths nothing writes were
+    // the only ones an allowed account could write anything at all to, and an
+    // account is only ever one compromise away from being that account.
+    //
+    // Two further things would have been wrong on the day publishing shipped,
+    // and both are cheaper to fix while the feature is unbuilt:
+    //
+    //  * `deleteEverything()` removes entries, word sets, progress and
+    //    sessions. It does not touch published copies — so "アカウントおよび
+    //    データを削除できます" would have been false for exactly the data other
+    //    people can see.
+    //  * `exportEverything()` does not include them either.
+    //
+    // **What has to come back with the feature**, restored as properties rather
+    // than as rules — the tests that covered them are named in
+    // `tests/rules/firestore-rules.test.ts` where this block used to be tested:
+    //
+    //  * ownerUid may never be reassigned on an existing copy, and may never be
+    //    somebody else's on a new one.
+    //  * A snapshot entry's parent set must be owned by the writer, checked
+    //    with `getAfter` and not `get` — a first publish writes the set and its
+    //    first entries in one batch, and `get` sees only committed state, so it
+    //    denies every first publish.
+    //  * Delete on a snapshot entry deliberately skips that parent check, so an
+    //    orphan left by a half-finished unpublish stays cleanable by its owner.
+    //  * Every mutation gates on `isAllowed()`, not only create: a revoked
+    //    account must lose the ability to edit what it has already put in front
+    //    of other people.
+    //  * Shape and size validation, which never existed here and must, on the
+    //    same `hasOnly` footing as the private collections above.
+    //  * Deletion and export have to cover published copies before the feature
+    //    is reachable, not after.
 
     // Anything not listed above stays denied. The old top-level /entries,
     // /entryProgress, /practiceSessions and /wordSets collections are gone: they
--- a/tests/rules/firestore-rules.test.ts
+++ b/tests/rules/firestore-rules.test.ts
@@ -176,12 +176,6 @@ describe('the allowedUsers gate', () => {
   });
 
   /**
-   * The gap the first review round found: update and delete were gated on
-   * ownership alone, so revoking an account left it able to keep editing what
-   * it had already put in front of other people — the one thing revocation is
-   * for.
-   */
-  /**
    * Revocation is the claim going away, not a document being deleted.
    *
    * That is a real change in what "revoked" means: the claim lives in the ID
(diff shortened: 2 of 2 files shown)


firestore.rules
```diff
@@ -251,49 +229,46 @@ service cloud.firestore {
...
+    //    sessions. It does not touch published copies — so "アカウントおよび
+    //    データを削除できます" would have been false for exactly the data other
+    //    people can see.
+    //  * `exportEverything()` does not include them either.
+    //
+    // **What has to come back with the feature**, restored as properties rather
+    // than as rules — the tests that covered them are named in
+    // `tests/rules/firestore-rules.test.ts` where this block used to be tested:
+    //
+    //  * ownerUid may never be reassigned on an existing copy, and may never be
```
kowinauyeung, line 277: 🟡 **SUGGESTION** — the list is the whole point of the change, and it drops every read property.

Six bullets, all of them about writes: `ownerUid` reassignment, the `getAfter` parent check, the delete exemption, `isAllowed()` on every mutation, shape validation, deletion and export. Nothing about reads. Three of the twelve deleted tests were about reads, and they are not recoverable from this block:

- `lets an allowlisted user read anyone's published content` — that reads are open to any allowed account and not only the owner is the *entire* access model of publishing. It is the first thing somebody rebuilding this has to decide and the list does not say what was decided.
- `stops serving snapshot entries once the parent set is gone` — the `exists(setPath(publicSetId))` condition on read. This is a security property, not a convenience: Firestore does not cascade deletes, so without it a half-finished unpublish leaves a snapshot readable from a path somebody already noted. The list keeps the *delete* half of the orphan story (bullet three) and drops the *read* half, which is the half that fails open.
- `lets the owner publish, republish and unpublish` — the positive path. A list of prohibitions with no statement of what must succeed is the shape that produces rules nobody can publish through, which is exactly the failure the `getAfter` bullet is about.

The deleted rules had `allow read: if isAllowed()` on both collections and `allow read: if isAllowed() && exists(setPath(publicSetId))` on the subcollection. Two more bullets carry all three.

_Codex review._

firestore.rules
```diff
@@ -251,49 +229,46 @@ service cloud.firestore {
...
+    //
+    //  * `deleteEverything()` removes entries, word sets, progress and
+    //    sessions. It does not touch published copies — so "アカウントおよび
+    //    データを削除できます" would have been false for exactly the data other
+    //    people can see.
+    //  * `exportEverything()` does not include them either.
+    //
+    // **What has to come back with the feature**, restored as properties rather
+    // than as rules — the tests that covered them are named in
+    // `tests/rules/firestore-rules.test.ts` where this block used to be tested:
```
kowinauyeung: 🟡 **SUGGESTION** — this pointer does not resolve, and it points in a circle.

「the tests that covered them are named in `tests/rules/firestore-rules.test.ts` where this block used to be tested」 — but at this ref that file no longer names them. What it has is a docblock listing the same properties, which points back here. A reader following the breadcrumb in either direction arrives at the other end of it and never reaches the twelve test names, which now exist only in history.

The names are the useful artefact — `refuses a create under a set that does not exist`, `stops serving snapshot entries once the parent set is gone` and the rest are a specification somebody can rebuild against. Point at the commit or at this pull request, which is a durable address for exactly that content, or list the names here.

_Codex review._

firestore.rules
```diff
@@ -251,49 +229,46 @@ service cloud.firestore {
...
+    //    with `getAfter` and not `get` — a first publish writes the set and its
+    //    first entries in one batch, and `get` sees only committed state, so it
+    //    denies every first publish.
+    //  * Delete on a snapshot entry deliberately skips that parent check, so an
+    //    orphan left by a half-finished unpublish stays cleanable by its owner.
+    //  * Every mutation gates on `isAllowed()`, not only create: a revoked
+    //    account must lose the ability to edit what it has already put in front
+    //    of other people.
+    //  * Shape and size validation, which never existed here and must, on the
+    //    same `hasOnly` footing as the private collections above.
```
kowinauyeung: 🟡 **SUGGESTION** — this forward-references a change that is not in this branch.

「on the same `hasOnly` footing as the private collections above」 — at this ref there is no `hasOnly` anywhere in the file. The private collections use `keys().size()` with field checks. `hasOnly` arrives in #29, whose own description ends 「`publicEntries` and `publicSets` still have no shape validation at all. They are next」, so the intended order is #29 then #30.

It is worth saying that intent out loud, because nothing else records it and the sentence quietly stops making sense if this merges first.

On the mechanical side there is no obstacle: merging `origin/fix/rules-close-unbuilt-publishing` into `origin/fix/rules-allowlist-the-fields` auto-merges both files with no conflict, and the result is right — nine `hasOnly` call sites, one copy of this block, no `publicEntries` or `publicSets` match blocks left. So this is about the comment reading true on its own, not about the branches fighting.

_Codex review._

firestore.rules
```diff
@@ -251,49 +229,46 @@ service cloud.firestore {
       }
     }
 
-    // A published copy is an immutable snapshot. Republishing overwrites it
-    // wholesale and bumps its version; unpublishing deletes it. Only the owner
-    // writes, and ownerUid can never be reassigned to somebody else.
+    // ---- publishing: closed until it exists ----
```
kowinauyeung, line 232: ✅ **PRAISE** — deleting the blocks rather than adding `if false` is right, and it is right for a stated reason.

Since the recursive wildcard went, an unlisted path is denied, so closing these *is* removing them. `setPath()` and `ownsParentSet()` going with them follows, having no other caller. A disabled-but-present rule is a rule somebody re-enables without re-deriving why it was disabled; an absent one has to be written again on purpose.

Confirmed independently that nothing in the application is affected: the three Firestore adapters address `entries`, `wordSets`, `practiceSessions` and `progress/entries` only, and `PublicationRepository` at `src/domain/ports.ts:103` is a port with no adapter and no caller. The two references to `publicEntries` in `src/` are a docblock and a field name on `Entry`, neither of which is a path.

_Codex review._

tests/rules/firestore-rules.test.ts
```diff
@@ -193,117 +187,89 @@ describe('the allowedUsers gate', () => {
...
    * directly, so the window itself is not reachable from here.
+   *
+   * The published-copy version of this went with the publishing rules. The
+   * property survives it: revocation has to take away update and delete on what
+   * the account already wrote, not only create. Gating `create` alone would
+   * leave a revoked account editing its own notebook indefinitely, which is the
+   * gap the first review round found.
    */
-  it('revokes update and delete on already-published copies, not just create', async () => {
+  it('takes away update and delete on what a revoked account already wrote', async () => {
```
kowinauyeung, line 197: ✅ **PRAISE** — the claim that this is not a duplicate of the existing stranger test holds up, and it is a distinction worth having made.

Checked both. The existing one uses `asStranger(MALLORY)` against `users/${MALLORY}/entries/e1`, a path `beforeEach` never seeds — so it asserts denial where there is no document. This one uses `asStranger(ALICE)` against `users/${ALICE}/entries/e1`, which *is* seeded, so it asserts denial of update, delete and read on an existing row belonging to the very account making the request.

Those are different properties, and only the second one is the property the deleted published-copy test was carrying: revocation has to take away what the account already wrote, not only what it might write next. Carrying it across to the private notebook rather than deleting it with the path is the right instinct — the path was incidental to the property.

_Codex review._

tests/rules/firestore-rules.test.ts
```diff
@@ -193,117 +187,89 @@ describe('the allowedUsers gate', () => {
...
+ * validation, and deletion and export covering published copies before the
+ * feature is reachable.
+ */
+describe('publishing, which is not built', () => {
+  /**
+   * The seeded documents exist — `beforeEach` writes them with rules disabled,
+   * as a partly-built feature or an earlier deploy would have left them. Being
+   * unreadable is therefore a fact about the rules and not about the fixture.
+   */
+  it('refuses to read a published copy, to its owner as much as to anyone else', async () => {
```
kowinauyeung, line 234: ✅ **PRAISE** — denying the owner too, and saying why, is what makes this suite worth having.

「Ownership was the entire basis of the rules being removed, so a check that let the owner through would leave the surface exactly as wide as it was」 is exactly right, and it is the sort of thing that gets softened into an owner exemption by the next person unless the reason is written next to it.

The note about the fixture earns its place as well: seeding through `withSecurityRulesDisabled` means being unreadable is a fact about the rules rather than about the setup, which is the difference between this test proving something and it passing vacuously.

_Codex review._

kowinauyeung:
## Review summary

**Verdict: `APPROVE`** — no blocker. Three suggestions, all about the comment block that is doing the preservation work.

Closing paths that nothing writes and that carry no validation is straightforwardly right, and the order of the two facts is what makes it urgent rather than tidy: `publicEntries` and `publicSets` were the *only* collections an allowed account could write anything at all to, and they were the two nothing writes. Deleting the blocks rather than disabling them follows correctly from the recursive wildcard being gone, and taking `setPath()` and `ownsParentSet()` with them is right for the same reason.

Confirmed independently that nothing breaks: the three Firestore adapters address `entries`, `wordSets`, `practiceSessions` and `progress/entries` only; `PublicationRepository` at `src/domain/ports.ts:103` has no adapter and no caller; the two `publicEntries` references in `src/` are a docblock and a field name on `Entry`, neither of them a path.

### The comment block is load-bearing, and it has three gaps

This change deliberately converts twelve tests into prose, so the prose has to carry them. Inline in detail; in short:

- **Every read property is missing.** The list is six bullets, all about writes. Three deleted tests were about reads, and one of them — `stops serving snapshot entries once the parent set is gone`, the `exists(setPath(...))` condition — is a security property that fails *open*: Firestore does not cascade deletes, so without it a half-finished unpublish leaves a snapshot readable from a known path. The list keeps the delete half of the orphan story and drops the read half.
- **The pointer at `:255` is circular.** It says the tests 「are named in `tests/rules/firestore-rules.test.ts`」; that file now lists the same properties and points back here. Neither end has the twelve names.
- **`:269` forward-references `hasOnly`,** which does not exist at this ref. It arrives in #29, whose description ends 「They are next」 — so the intended order is #29 then #30, and nothing records that.

On the mechanics of that ordering: I merged this branch into #29's locally and it auto-merges both files with no conflict, producing the right result — nine `hasOnly` call sites, one copy of this block, no publishing match blocks. So the dependency is on the sentence reading true, not on the branches colliding.

### Verified in the description

- The claim that the surviving renamed test is not a duplicate is correct. The existing stranger test uses `MALLORY` against a path `beforeEach` never seeds; this one uses `asStranger(ALICE)` against the seeded `users/${ALICE}/entries/e1`, so it asserts denial of update, delete and read on an existing row. Different properties, and only the second is the one the deleted test carried.
- The test arithmetic holds. Rules `it()` blocks go 40 → 32; the emulator total 73 → 65 is consistent once the 33 adapter integration tests are added back. Both match the stated 65 and the 61 in the red-before-green run.

### One correction to the description

🔵 「Four cases replacing eleven」 — it is twelve. Five in `published copies`, seven in `snapshot entries under a published set`. The net of −8 that the counts above confirm is −12 + 4, so every other number in the description is right and only this word is off.

### Not verified

Static review plus a local trial merge. Tests were not run; the counts above were taken by counting `it(` blocks across the three refs, not by executing the suite.

---
_Codex review._

firestore.rules
kowinauyeung: @kowinauyeung
Upheld. Six bullets, all writes — and the framing that makes it obvious is yours:

> The list keeps the *delete* half of the orphan story and drops the *read* half, which is the half that fails open.

That is the one I would least have wanted to lose. `allow read: if isAllowed() && exists(setPath(publicSetId))` is a security property, and I had kept the delete exemption that exists *because* of it while dropping the condition itself. Somebody rebuilding from my list would have written the exemption and never known what it was an exemption from.

Three read bullets added, and placed **before** the writes, because who may read a published copy is the access model rather than a detail:

- A published copy is readable by any allowed account, not only its owner — quoting the old rule, so the decision is recoverable and not merely referred to.
- A snapshot entry is readable only while its parent set exists, with the non-cascading delete spelled out and the two halves of the orphan story named as failing open and closed respectively.
- The owner must be able to publish, republish and unpublish. Your reason for this one is better than any I would have given: a list of prohibitions with nothing that must succeed is how rules nobody can publish through get written, which is what the `getAfter` bullet exists to prevent.

Fixed in 0fe8265

firestore.rules
Upheld, and it is a circle I built in one sitting without noticing, because each end was written while looking at the other.

Both ends now point at pull request #30, which is a durable address holding the twelve names. The test docblock says so explicitly — 「This docblock points at the rules and the rules point at that pull request, so neither is a step in a circle」 — so the next person to add a pointer here has the shape written down rather than has to notice it.

Agreed that the names are the useful artefact. `refuses a create under a set that does not exist` and `stops serving snapshot entries once the parent set is gone` are a specification in a way that a paraphrase is not, which is why the bullets now quote the deleted rules where a rule is short enough to quote.

Fixed in 0fe8265

firestore.rules
Upheld — the sentence describes a file that does not exist at this ref.

Rewritten to say the ordering out loud rather than to assume it:

> …on the same footing as the private collections above — which by then means `hasOnly` plus per-field bounds. That arrives in #29, not here: this branch still bounds the private collections with `keys().size()`, and the intended order is #29 first. The two merge cleanly either way; it is this sentence that stops being true if they land the other way round.

Thank you for testing the merge rather than reasoning about it. That it auto-merges to nine `hasOnly` call sites, one copy of this block and no `publicEntries` or `publicSets` match blocks is exactly what I would have wanted to know and had not checked.

One thing has changed under you since: #29 now also bounds the fields the allowlist only permitted, so "the same footing" is `hasOnly` **plus** per-field bounds rather than `hasOnly` alone. The sentence above is written to that.

Fixed in 0fe8265

firestore.rules
kowinauyeung: @kowinauyeung
Thanks — and the distinction you draw is the one I would want kept:

> A disabled-but-present rule is a rule somebody re-enables without re-deriving why it was disabled; an absent one has to be written again on purpose.

Also for checking that the two `publicEntries` references in `src/` are a docblock and a field name rather than paths. That is precisely the check that would have stopped this if it had gone the other way, and it is not one a green test suite performs.

tests/rules/firestore-rules.test.ts
kowinauyeung: @kowinauyeung
Thanks for running both rather than taking the claim. The difference is exactly as you describe: the existing test denies where `beforeEach` seeds nothing, and this one denies update, delete and read on a seeded row belonging to the account making the request.

「the path was incidental to the property」 is a better statement of why it was carried across than the one I wrote, and it is the reason I did not simply delete it with the rules: a test named after a path goes with the path, and a test named after a property outlives it.

tests/rules/firestore-rules.test.ts
kowinauyeung: @kowinauyeung
Thanks. The owner exemption is the specific thing I expected to be asked for, which is why the reason sits next to the assertion rather than in the pull request description — a description is not there six months later when somebody is looking at a failing test and reaching for the smallest change that makes it pass.

Your point about the fixture is the one I would underline. Seeding through `withSecurityRulesDisabled` is what makes the assertion mean something: if the documents were absent, every one of these would pass while proving nothing, and it would look identical.

firestore.rules
kowinauyeung: **Verified fixed** in `0fe8265`, and structured better than I suggested: reads are now a labelled group placed **first**, 「because they are the access model rather than a detail」, with the writes after them.

All three are there, each with the rule it came from — `allow read: if isAllowed()` on both collections, `allow read: if isAllowed() && exists(setPath(publicSetId))` on the subcollection, and the owner's publish/republish/unpublish path. Naming the fail-open half against the fail-closed one in the same bullet is what makes the orphan story readable as a pair rather than as two unrelated notes.

The test docblock was updated to match, so the two ends now list the same set.

_Codex re-review._

firestore.rules
kowinauyeung: **Verified fixed.** Both ends now point at pull request #30, and the test docblock says so explicitly — 「This docblock points at the rules and the rules point at that pull request, so neither is a step in a circle」.

The rules comment also states plainly that the twelve tests 「are not recoverable from this list」, which is the honest framing: the list is a specification to rebuild against, not a substitute for what was deleted.

_Codex re-review._

firestore.rules
kowinauyeung: **Verified fixed**, and it now says more than I asked for. Rather than removing the forward reference it keeps it and dates it: `hasOnly` plus per-field bounds 「arrives in #29, not here」, this branch still uses `keys().size()`, the intended order is #29 first, and — the part worth having — 「The two merge cleanly either way; it is this sentence that stops being true if they land the other way round」.

That matches what I measured: merging the two branches auto-merges both files with no conflict. A reader now gets the ordering, the reason for it, and the fact that ignoring it costs a stale sentence rather than a broken deploy.

_Codex re-review._

firestore.rules
kowinauyeung: Unchanged on the new head. Still holds.

_Codex re-review._

tests/rules/firestore-rules.test.ts
kowinauyeung: Unchanged on the new head. Still holds.

_Codex re-review._

tests/rules/firestore-rules.test.ts
kowinauyeung: Unchanged on the new head. Still holds.

_Codex re-review._

kowinauyeung:
## Re-review — `fdd3b25` → `0fe8265`

**Verdict: `APPROVE`.** All three suggestions fixed; all six threads resolved.

- **The read properties are back**, and organised better than I suggested: a labelled group placed **first**, 「because they are the access model rather than a detail」, with the writes after. Each bullet carries the rule it came from — `allow read: if isAllowed()` on both collections, `allow read: if isAllowed() && exists(setPath(publicSetId))` on the subcollection, and the owner's publish/republish/unpublish path. Naming the fail-open half against the fail-closed one in the same breath is what makes the orphan story readable as a pair.
- **The circular pointer is gone.** Both ends now address pull request #30, and the test docblock says so outright: 「This docblock points at the rules and the rules point at that pull request, so neither is a step in a circle.」 The rules comment also states that the twelve tests 「are not recoverable from this list」, which is the honest framing — a specification to rebuild against, not a substitute.
- **The `hasOnly` reference is dated rather than removed**, which is more useful: it says the per-field bounds arrive in #29, that this branch still uses `keys().size()`, that the intended order is #29 first, and that 「The two merge cleanly either way; it is this sentence that stops being true if they land the other way round」. That matches what I measured — the branches auto-merge with no conflict — so the reader gets the ordering, the reason, and the cost of ignoring it.

### One thing left, and it is only the description

🔵 「Four cases replacing eleven」 is still in the body. It is twelve: five in `published copies`, seven in `snapshot entries under a published set`. The rules `it()` count going 40 → 32 is −12 + 4, and the emulator total 73 → 65 confirms it — so every other number in the description is right and only this word is off. Not worth a push on its own; worth folding into the next one that touches this branch.

---
_Codex re-review._

PR #31, opened 2026-08-13
fix(account): prove the session first, and notice a stuck walk on page two

kowinauyeung:
Two failures on the same two paths, both of which cost more than they should when they go wrong.

## Deleting could succeed at everything irreversible and fail at the end

`deleteEverything()` removes word sets, entries and practice history, then deletes the account. **That order is correct and is unchanged**: delete the account first and there is no session left to delete the data with.

What the order cannot survive on its own is the provider refusing the last step. Firebase raises `auth/requires-recent-login` on a stale session, so a delete run from an old-enough one succeeded at every destructive write and failed only at the one needing proof. The user pressed delete and was left holding an account with nothing in it — both halves of the outcome they did not want, and unrecoverable.

`AuthPort` gains `reauthenticate()`, called as the **first statement** of `deleteEverything()`. The failure moves to the one point where nothing has happened yet, and costs a popup.

`reauthenticateWithPopup`, not `signInWithPopup`: the latter succeeds for a *different* Google account, and the caller is about to delete everything belonging to the current one.

### The two messages now describe two situations

A cancelled or refused popup leaves **nothing deleted**, and `Account.tsx` says so. The previous wording warned that some data might already be gone — true when it was written, a false alarm now. The part-way message stays for everything else, because everything else can still fail part-way.

## A stuck walk cost a million reads

`drain()` gave up after 10,000 pages at 100 records each. A cursor that stops advancing therefore spent **1,000,000 document reads** — twenty times the free daily allowance — establishing something the second page already showed.

A `Set` of cursors already handed out catches it on the second read. The page count stays as a backstop rather than as the mechanism.

**`deleteEverything` drains twice as well**, so this was reachable from the delete button and not only from export.

## Layer

`tests/unit`, per `CLAUDE.md`. Both claims are functions of their inputs — no DOM, no clock, no emulator. Two tests:

- `refuses to delete anything at all when the session cannot be proved` — asserts all four removal ports were untouched, which is the property; asserting only that it threw would pass with the data already gone.
- `stops on the second page rather than after ten thousand` — asserts the call count is exactly 2. Asserting only that it threw would pass with the old code, since the old code threw too, a million reads later.

## Red before green, each fix separately

```
reauthenticate call removed:
  × refuses to delete anything at all when the session cannot be proved      497 passing

cursor guard removed, page cap alone:
  × stops on the second page rather than after ten thousand                  497 passing
```

Removing either leaves the other green, so neither test is standing in for the other.

## Also changed

`src/lib/backend.e2e.ts` implements `reauthenticate` as a resolved promise, for the same reason it never refuses `deleteAccount`: the end-to-end suite cannot age a session, and a fake that refuses covers nothing.

571 unit and component (up from 569), lint clean, both typecheck passes clean, prettier clean. Nothing in `tests/` was pinned to the previous wording — checked.

### DIFF
--- a/src/infra/firebase/authAdapter.ts
+++ b/src/infra/firebase/authAdapter.ts
@@ -1,5 +1,11 @@
 import type { User } from 'firebase/auth';
-import { deleteUser, onAuthStateChanged, signInWithPopup, signOut } from 'firebase/auth';
+import {
+  deleteUser,
+  onAuthStateChanged,
+  reauthenticateWithPopup,
+  signInWithPopup,
+  signOut,
+} from 'firebase/auth';
 import type { AuthPort, AuthUser } from '@/domain/ports';
 import { auth, googleProvider } from './client';
 
@@ -25,6 +31,18 @@ export const firebaseAuth: AuthPort = {
   },
   signOut: async () => {
     await signOut(auth);
+  },
+  /**
+   * The same popup as signing in, against the account already signed in.
+   *
+   * `reauthenticateWithPopup` rather than `signInWithPopup`: the latter would
+   * succeed for a *different* Google account, and the caller is about to delete
+   * everything belonging to the current one. Firebase refuses the swap here.
+   */
+  reauthenticate: async () => {
+    const user = auth.currentUser;
+    if (!user) throw new Error('ログインしていません。');
+    await reauthenticateWithPopup(user, googleProvider);
   },
   /**
    * Firebase refuses this on a session that is not recent, raising
--- a/src/lib/accountData.ts
+++ b/src/lib/accountData.ts
@@ -45,10 +45,25 @@ const PAGE = 100;
 async function drain<T>(read: (q: PageQuery) => Promise<Page<T>>): Promise<T[]> {
   const all: T[] = [];
   let cursor: string | null = null;
+  // A cursor already handed out means the walk is not advancing. Recognising
+  // that is the guard; the page count below is only a backstop.
+  //
+  // It used to be the other way round, and the cost of that was the whole
+  // point: a stuck cursor ran the loop to its 10,000-page limit at 100 records
+  // a page, so the failure mode of "this query stopped moving" was a million
+  // document reads — twenty times the free daily allowance, spent proving
+  // something the second page already showed. Worse, `deleteEverything` drains
+  // twice too, so it was reachable from the delete button as well as from
+  // export.
+  const seen = new Set<string>();
   for (let page = 0; page < 10_000; page += 1) {
     const result: Page<T> = await read({ limit: PAGE, cursor });
     all.push(...result.items);
     if (!result.cursor) return all;
+    if (seen.has(result.cursor)) {
+      throw new Error('エクスポートが終わりませんでした。時間をおいて再度お試しください。');
+    }
+    seen.add(result.cursor);
     cursor = result.cursor;
   }
   throw new Error('エクスポートが終わりませんでした。時間をおいて再度お試しください。');
@@ -134,6 +149,20 @@ export async function deleteEverything({
   progress: ProgressRepository;
   auth: AuthPort;
 }): Promise<{ entries: number; wordSets: number }> {
+  // **First, and before anything irreversible.**
+  //
+  // The order below — data, then the account — is deliberate and stays: delete
+  // the account first and there is no session left to delete the data with.
+  // What that order cannot survive is the provider refusing the last step. On a
+  // session old enough for `auth/requires-recent-login`, every destructive write
+  // succeeded and only the account remained, so a user who pressed delete was
+  // left with an account holding nothing and a message telling them some of it
+  // was already gone.
+  //
+  // Proving the session up front costs a popup and moves that failure to the
+  // one place where nothing has happened yet.
+  await auth.reauthenticate();
+
   const setList = await drain((q) => wordSets.list(q));
   for (const set of setList) await wordSets.remove(set.id);
 
--- a/src/lib/accountData.ts
+++ b/src/lib/accountData.ts
@@ -42,6 +42,13 @@ const PAGE = 100;
  * The guard is a page count rather than a timeout: a cursor that stops moving
  * would otherwise spin forever, and 10,000 pages is far past any real notebook.
  */
+/**
+ * One literal, because `deleteEverything` drains as well — so this string is
+ * already reachable from a path it does not describe. Two copies would drift,
+ * and the copy that drifts is the one nobody is looking at.
+ */
+const WALK_STOPPED = 'エクスポートが終わりませんでした。時間をおいて再度お試しください。';
+
 async function drain<T>(read: (q: PageQuery) => Promise<Page<T>>): Promise<T[]> {
   const all: T[] = [];
   let cursor: string | null = null;
@@ -60,13 +67,11 @@ async function drain<T>(read: (q: PageQuery) => Promise<Page<T>>): Promise<T[]> 
     const result: Page<T> = await read({ limit: PAGE, cursor });
     all.push(...result.items);
     if (!result.cursor) return all;
-    if (seen.has(result.cursor)) {
-      throw new Error('エクスポートが終わりませんでした。時間をおいて再度お試しください。');
-    }
+    if (seen.has(result.cursor)) throw new Error(WALK_STOPPED);
     seen.add(result.cursor);
     cursor = result.cursor;
   }
-  throw new Error('エクスポートが終わりませんでした。時間をおいて再度お試しください。');
+  throw new Error(WALK_STOPPED);
 }
 
 export async function exportEverything({
@@ -138,6 +143,33 @@ export function exportFilename(now = new Date()): string {
  * decision worth making on its own rather than as a side effect of adding a
  * delete button.
  */
+/**
+ * The session could not be proved, so **nothing was deleted**.
+ *
+ * This exists because the caller was inferring that from a provider error code
+ * and could not. `auth/requires-recent-login` is the clearest case: it is
+ * `CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, which the backend returns for a *sensitive
+ * operation* — here `deleteAccount()`, the last statement below. Reading it as
+ * "the re-authentication was refused" inverts the truth exactly: by the time it
+ * can be raised, every word set, entry, session and the progress map are gone.
+ *
+ * `reauthenticateWithPopup` raises `auth/user-mismatch`, `auth/popup-blocked`
+ * and the cancellation codes; `authAdapter` can also throw with no `code` at
+ * all. That list is provider-defined and moves between SDK versions. This
+ * function is the only place that knows which half it got through, so it says
+ * so and the caller stops guessing.
+ */
+export class NothingDeleted extends Error {
+  /** What the provider actually raised, for the console. */
+  readonly reason: unknown;
+
+  constructor(reason: unknown) {
+    super('reauthentication failed, so nothing was deleted');
+    this.name = 'NothingDeleted';
+    this.reason = reason;
+  }
+}
+
 export async function deleteEverything({
   entries,
   wordSets,
@@ -161,7 +193,11 @@ export async function deleteEverything({
   //
   // Proving the session up front costs a popup and moves that failure to the
   // one place where nothing has happened yet.
-  await auth.reauthenticate();
+  try {
+    await auth.reauthenticate();
+  } catch (cause) {
+    throw new NothingDeleted(cause);
+  }
 
   const setList = await drain((q) => wordSets.list(q));
   for (const set of setList) await wordSets.remove(set.id);
--- a/src/routes/Account.tsx
+++ b/src/routes/Account.tsx
@@ -80,13 +80,23 @@ export function Component() {
       await signOutUser();
     } catch (cause) {
       console.error(cause);
-      // `requires-recent-login` is not a failure the user can act on without
-      // being told what it is: the provider refuses to delete an account on a
-      // stale session, and the fix is to sign in again.
+      // The re-authentication now runs before the first delete, so these two
+      // messages describe genuinely different situations rather than the same
+      // half-finished one.
+      //
+      // A cancelled or refused popup happens with **nothing deleted**, and the
+      // message says so plainly — the previous wording warned that some data
+      // might already be gone, which was true then and would be a false alarm
+      // now. Anything else still fails part-way, and still has to say so.
       const code = (cause as { code?: string }).code;
+      const cancelled =
+        code === 'auth/popup-closed-by-user' ||
+        code === 'auth/cancelled-popup-request' ||
+        code === 'auth/user-cancelled' ||
+        code === 'auth/requires-recent-login';
       setError(
-        code === 'auth/requires-recent-login'
-          ? 'セキュリティのため、一度ログインし直してから削除してください。データの一部はすでに削除されている可能性があります。'
+        cancelled
+          ? 'セキュリティのため、削除の前に本人確認が必要です。データは削除されていません。もう一度お試しください。'
           : // Deliberately not "failed": some of it may be gone. Saying so is
             // the difference between a user retrying and a user assuming their
             // data is intact when half of it is not.
--- a/src/routes/Account.tsx
+++ b/src/routes/Account.tsx
@@ -2,7 +2,12 @@ import { useMemo, useState } from 'react';
 import { Link } from 'react-router-dom';
 import { ConfirmDialog } from '@/components/ConfirmDialog';
 import { CopyDiagnostics } from '@/components/CopyDiagnostics';
-import { deleteEverything, exportEverything, exportFilename } from '@/lib/accountData';
+import {
+  deleteEverything,
+  exportEverything,
+  exportFilename,
+  NothingDeleted,
+} from '@/lib/accountData';
 import { useAuth } from '@/lib/auth';
 import { authPort } from '@/lib/backend';
 import { appVersion, buildLine } from '@/lib/build';
@@ -80,22 +85,20 @@ export function Component() {
       await signOutUser();
     } catch (cause) {
       console.error(cause);
-      // The re-authentication now runs before the first delete, so these two
-      // messages describe genuinely different situations rather than the same
-      // half-finished one.
+      // Branching on what happened, not on which code the provider chose.
       //
-      // A cancelled or refused popup happens with **nothing deleted**, and the
-      // message says so plainly — the previous wording warned that some data
-      // might already be gone, which was true then and would be a false alarm
-      // now. Anything else still fails part-way, and still has to say so.
-      const code = (cause as { code?: string }).code;
-      const cancelled =
-        code === 'auth/popup-closed-by-user' ||
-        code === 'auth/cancelled-popup-request' ||
-        code === 'auth/user-cancelled' ||
-        code === 'auth/requires-recent-login';
+      // The list this replaces was wrong in both directions, and one of the
+      // errors was the dangerous kind. `auth/requires-recent-login` is
+      // `CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, raised for a *sensitive operation* —
+      // which here is `deleteAccount()`, the last thing `deleteEverything`
+      // does. Treating it as "the popup was refused" meant telling a user
+      // 「データは削除されていません」 at the one moment everything already had
+      // been. It also missed `auth/user-mismatch`, `auth/popup-blocked` and the
+      // adapter's own error, which carries no `code` at all.
+      //
+      // `NothingDeleted` is thrown by the only code that knows.
       setError(
-        cancelled
+        cause instanceof NothingDeleted
           ? 'セキュリティのため、削除の前に本人確認が必要です。データは削除されていません。もう一度お試しください。'
           : // Deliberately not "failed": some of it may be gone. Saying so is
             // the difference between a user retrying and a user assuming their
--- a/tests/unit/accountData.test.ts
+++ b/tests/unit/accountData.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it, vi } from 'vitest';
 import type { Page, PageQuery } from '@/domain/ports';
-import { deleteEverything, exportEverything, exportFilename } from '@/lib/accountData';
+import {
+  deleteEverything,
+  exportEverything,
+  exportFilename,
+  NothingDeleted,
+} from '@/lib/accountData';
 
 /** A repository holding `count` rows and serving them `PAGE` at a time. */
 const paged = <T>(items: T[]) => {
@@ -160,16 +165,48 @@ describe('deleteEverything', () => {
    */
   it('refuses to delete anything at all when the session cannot be proved', async () => {
     const p = ports(5, 5, 5);
+    // A code `reauthenticateWithPopup` actually raises. It is the one the
+    // adapter chose that call for: the user picked a different Google account,
+    // and the caller is about to delete everything belonging to this one.
     (p.auth.reauthenticate as ReturnType<typeof vi.fn>).mockRejectedValue(
-      Object.assign(new Error('stale'), { code: 'auth/requires-recent-login' }),
+      Object.assign(new Error('different account'), { code: 'auth/user-mismatch' }),
     );
 
-    await expect(deleteEverything(p)).rejects.toThrow('stale');
+    await expect(deleteEverything(p)).rejects.toBeInstanceOf(NothingDeleted);
 
     expect((p.entries.remove as ReturnType<typeof vi.fn>).mock.calls).toHaveLength(0);
     expect((p.wordSets.remove as ReturnType<typeof vi.fn>).mock.calls).toHaveLength(0);
     expect((p.progress.removeAll as ReturnType<typeof vi.fn>).mock.calls).toHaveLength(0);
     expect((p.auth.deleteAccount as ReturnType<typeof vi.fn>).mock.calls).toHaveLength(0);
+  });
+
+  /**
+   * The state a user must never be told is safe.
+   *
+   * `auth/requires-recent-login` is `CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, which the
+   * backend returns for a sensitive operation — and the only sensitive
+   * operation here is `deleteAccount()`, which runs last. So this code can only
+   * reach a caller *after* every word set, entry, session and the progress map
+   * are already gone. Reading it as "the re-authentication was refused" and
+   * reporting 「データは削除されていません」 is the exact inverse of what
+   * happened, and it is unrecoverable: the reader retries later into an account
+   * holding nothing.
+   *
+   * The window is real rather than theoretical — the removals are one sequential
+   * `await` each, and nothing bounds how long that takes.
+   */
+  it('does not report a late refusal as nothing having been deleted', async () => {
+    const p = ports(5, 5, 5);
+    (p.auth.deleteAccount as ReturnType<typeof vi.fn>).mockRejectedValue(
+      Object.assign(new Error('stale'), { code: 'auth/requires-recent-login' }),
+    );
+
+    const failure = await deleteEverything(p).catch((cause: unknown) => cause);
+
+    expect(failure).not.toBeInstanceOf(NothingDeleted);
+    // ...because plenty was.
+    expect((p.entries.remove as ReturnType<typeof vi.fn>).mock.calls).toHaveLength(5);
+    expect((p.progress.removeAll as ReturnType<typeof vi.fn>).mock.calls).toHaveLength(1);
   });
 });
 
--- a/tests/unit/accountData.test.ts
+++ b/tests/unit/accountData.test.ts
@@ -24,7 +24,10 @@ const ports = (entryCount: number, setCount: number, sessionCount: number) => {
     listSessions: paged(rows(sessionCount, 'p')),
     removeAll: vi.fn(() => Promise.resolve()),
   };
-  const auth = { deleteAccount: vi.fn(() => Promise.resolve()) };
+  const auth = {
+    reauthenticate: vi.fn(() => Promise.resolve()),
+    deleteAccount: vi.fn(() => Promise.resolve()),
+  };
   // Only the methods under test are implemented; the ports are wider.
   return {
     entries,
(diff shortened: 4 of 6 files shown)


src/routes/Account.tsx
```diff
@@ -80,13 +80,23 @@ export function Component() {
...
+      // A cancelled or refused popup happens with **nothing deleted**, and the
+      // message says so plainly — the previous wording warned that some data
+      // might already be gone, which was true then and would be a false alarm
+      // now. Anything else still fails part-way, and still has to say so.
       const code = (cause as { code?: string }).code;
+      const cancelled =
+        code === 'auth/popup-closed-by-user' ||
+        code === 'auth/cancelled-popup-request' ||
+        code === 'auth/user-cancelled' ||
+        code === 'auth/requires-recent-login';
```
kowinauyeung: 🔴 **BLOCKER** — this is the one code in the list that, after this change, can only mean the opposite of what the message says.

`reauthenticateWithPopup` does not raise `auth/requires-recent-login`. Grounded in the copy of `@firebase/auth` in this repo rather than from memory:

- `_reauthenticate`, which backs `reauthenticateWithPopup`, parses the returned token and asserts `user.uid === localId`, raising **`auth/user-mismatch`** when they differ. That is the failure mode this call was chosen for.
- `auth/requires-recent-login` is `AuthErrorCode.CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, mapped from the backend's response to a *sensitive operation* — 「This operation is sensitive and requires recent authentication」. In this file the only sensitive operation is `deleteAccount()`, and `deleteAccount()` is the last statement of `deleteEverything()`.

So after this change `auth/requires-recent-login` reaching this handler means the popup succeeded, every word set was deleted, every entry was deleted, the progress map and every session were deleted, and the account itself then refused. At that exact moment this branch tells the user 「データは削除されていません」.

That is the same false statement this pull request is removing from the other message, pointed the other way and made worse: the old wording over-warned, this one reassures. A user who reads it retries later, is signed in to an account holding nothing, and has no way to tell that the deletion already happened.

And the window is real rather than theoretical, which is what makes this reachable rather than merely wrong. Between the two calls, `deleteEverything` drains word sets and entries a page at a time and then removes them **one sequential `await` each** — `for (const entry of entryList) await entries.remove(entry.id)`. Nothing bounds how long that takes; Firebase's recent-login window does not extend to cover it. A large enough notebook, or a slow enough connection, walks straight into it.

Removing `auth/requires-recent-login` from this list fixes the false reassurance. It does not fix what the user is then told, which is the suggestion on `accountData.ts`.

_Codex review._

src/routes/Account.tsx
```diff
@@ -80,13 +80,23 @@ export function Component() {
...
+      // The re-authentication now runs before the first delete, so these two
+      // messages describe genuinely different situations rather than the same
+      // half-finished one.
+      //
+      // A cancelled or refused popup happens with **nothing deleted**, and the
+      // message says so plainly — the previous wording warned that some data
+      // might already be gone, which was true then and would be a false alarm
+      // now. Anything else still fails part-way, and still has to say so.
       const code = (cause as { code?: string }).code;
+      const cancelled =
```
kowinauyeung: 🟡 **SUGGESTION** — the set is under-inclusive as well as over-inclusive, and one of the omissions is the case the adapter was written for.

Failures that leave **nothing deleted** and are not on this list:

- **`auth/user-mismatch`** — the user picked a different Google account in the popup. This is precisely the case the docblock on `authAdapter.ts` cites as the reason for choosing `reauthenticateWithPopup` over `signInWithPopup`, and the person who hits it gets told their data may be half gone.
- **`auth/popup-blocked`** — the browser refused to open the popup. Common, entirely recoverable, nothing touched.
- **`auth/network-request-failed`** — could be either, but on the reauth call it is nothing.
- **No code at all.** `authAdapter.ts:44` throws `new Error('ログインしていません。')` with no `code` property, so `code` is `undefined`, `cancelled` is false, and a user who was never signed in is told a deletion failed part way.

Enumerating provider error codes is a losing game — the list is provider-defined, changes between SDK versions, and is wrong in both directions here on first attempt. See the suggestion on `accountData.ts:164` for the version of this that does not need the list at all.

_Codex review._

src/lib/accountData.ts
```diff
@@ -134,6 +149,20 @@ export async function deleteEverything({
...
+  // the account first and there is no session left to delete the data with.
+  // What that order cannot survive is the provider refusing the last step. On a
+  // session old enough for `auth/requires-recent-login`, every destructive write
+  // succeeded and only the account remained, so a user who pressed delete was
+  // left with an account holding nothing and a message telling them some of it
+  // was already gone.
+  //
+  // Proving the session up front costs a popup and moves that failure to the
+  // one place where nothing has happened yet.
+  await auth.reauthenticate();
```
kowinauyeung: 🟡 **SUGGESTION** — the caller knows which phase failed; making it say so removes the error-code taxonomy entirely.

`deleteEverything` is the only place that knows whether `reauthenticate()` returned. `Account.tsx` is reduced to inferring it from a provider error code, and the two findings above are both consequences of that inference being unreliable — over-inclusive on `auth/requires-recent-login`, under-inclusive on `auth/user-mismatch`, `auth/popup-blocked` and the adapter's own uncoded error.

The fact is available for free right here:

```ts
try {
  await auth.reauthenticate();
} catch (cause) {
  throw new NothingDeletedError(cause);   // or { phase: 'reauthenticate' }
}
```

Everything after that point is genuinely part-way and keeps the existing message. `Account.tsx` then branches on one thing it can be certain of instead of on a list it has to keep in step with the SDK.

The second reason to move it is coverage. There is no test for `Account.tsx` — `tests/component/` has twelve files and none of them is this route — so this branch, which decides whether a user is told their data is intact, is the one piece of this change nothing exercises. Moved into `accountData.ts` it is a function of its inputs and belongs in `tests/unit` alongside the two excellent tests this pull request already adds.

_Codex review._

tests/unit/accountData.test.ts
```diff
@@ -143,4 +146,57 @@ describe('deleteEverything', () => {
...
+   * succeeded at every irreversible step and failed only at the one that needed
+   * proving. The user pressed delete and was left holding an account with
+   * nothing in it — the worst of both outcomes, and unrecoverable.
+   *
+   * Proving the session first turns that into a popup.
+   */
+  it('refuses to delete anything at all when the session cannot be proved', async () => {
+    const p = ports(5, 5, 5);
+    (p.auth.reauthenticate as ReturnType<typeof vi.fn>).mockRejectedValue(
+      Object.assign(new Error('stale'), { code: 'auth/requires-recent-login' }),
```
kowinauyeung: 🟡 **SUGGESTION** — this mock is the same mistaken assumption as the blocker, written into the suite.

`reauthenticate()` is made to reject with `auth/requires-recent-login`, which is the one code that call does not produce: it is `CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, raised on the sensitive operation — `deleteAccount()` — not on the reauthentication that exists to prevent it.

The test passes regardless, because `deleteEverything` rethrows whatever it caught and the assertions are about the four removal ports being untouched, which is the right property. But it now stands as the file's statement of what failure looks like here, and the next reader takes the code from it. `CLAUDE.md` asks that a test name say what breaks in production; this one's *fixture* says something that cannot happen.

Two changes, and the second is the one that would have caught the blocker:

- Reject with `auth/popup-closed-by-user` (or `auth/user-mismatch`) — a code that call really raises.
- Add the case the blocker is about: `deleteAccount()` rejecting with `auth/requires-recent-login` *after* the data is gone, asserting that the removal ports **were** called. That is the state the user is currently told does not exist.

_Codex review._

src/infra/firebase/authAdapter.ts
```diff
@@ -26,6 +32,18 @@ export const firebaseAuth: AuthPort = {
...
+   * The same popup as signing in, against the account already signed in.
+   *
+   * `reauthenticateWithPopup` rather than `signInWithPopup`: the latter would
+   * succeed for a *different* Google account, and the caller is about to delete
+   * everything belonging to the current one. Firebase refuses the swap here.
+   */
+  reauthenticate: async () => {
+    const user = auth.currentUser;
+    if (!user) throw new Error('ログインしていません。');
+    await reauthenticateWithPopup(user, googleProvider);
```
kowinauyeung, line 45: ✅ **PRAISE** — `reauthenticateWithPopup` over `signInWithPopup` is the right call and the docblock gives the right reason.

「the latter would succeed for a *different* Google account, and the caller is about to delete everything belonging to the current one」 — confirmed in the SDK: `_reauthenticate` asserts `user.uid` against the `sub` claim of the returned token and refuses the swap. `signInWithPopup` would have proved that *somebody* is present, which is not the question being asked before an irreversible delete.

This is a distinction that is easy to get wrong and invisible once it is wrong, since both calls open the same popup and both look like they worked.

_Codex review._

src/lib/accountData.ts
```diff
@@ -45,10 +45,25 @@ const PAGE = 100;
...
+  // that is the guard; the page count below is only a backstop.
+  //
+  // It used to be the other way round, and the cost of that was the whole
+  // point: a stuck cursor ran the loop to its 10,000-page limit at 100 records
+  // a page, so the failure mode of "this query stopped moving" was a million
+  // document reads — twenty times the free daily allowance, spent proving
+  // something the second page already showed. Worse, `deleteEverything` drains
+  // twice too, so it was reachable from the delete button as well as from
+  // export.
+  const seen = new Set<string>();
```
kowinauyeung, line 65: ✅ **PRAISE**, with one small thing.

A repeated cursor is the right signal — it is the definition of the walk not advancing, and it is available on the second read rather than after the cap. Demoting the page count to a backstop instead of removing it is also right: it still catches a cursor that keeps changing without terminating, which the set cannot see.

The arithmetic in the description is what makes the case, and it is worth having done: 10,000 pages × 100 records is twenty times the free daily read allowance, spent establishing what page two already showed, and `deleteEverything` drains twice so the delete button reached it too.

🔵 The message literal is now duplicated on lines 64 and 69. Worth a `const` — two copies of a user-visible string drift, and this one is about to be read in a context it does not describe: `deleteEverything` drains as well, so a stuck walk during deletion throws 「エクスポートが終わりませんでした」. Harmless today, because the delete handler in `Account.tsx` discards `cause.message` and shows a fixed string — checked — but it is a trap for whoever next decides to surface the real message there.

_Codex review._

kowinauyeung:
## Review summary

**Verdict: `REQUEST CHANGES`** — one blocker, in the part of the change that decides what a user is told about whether their data still exists.

Both defects are real and both diagnoses are right. Deleting the data before the account is the correct order and the description is right to say so and keep it; what that order cannot survive on its own is the provider refusing the last step, and moving the proof to the front is the right answer. The `drain` arithmetic is right too — 10,000 pages of 100 is twenty times the free daily read allowance, spent establishing what page two already showed — and demoting the page cap to a backstop rather than removing it keeps the case the cursor set cannot see.

### The blocker

`auth/requires-recent-login` is in the `cancelled` set, and after this change it is the one code there that can only mean the opposite of what the message says.

Grounded in the copy of `@firebase/auth` in this repository rather than from memory:

- `_reauthenticate`, which backs `reauthenticateWithPopup`, asserts `user.uid` against the `sub` claim of the returned token and raises **`auth/user-mismatch`** when they differ.
- `auth/requires-recent-login` is `AuthErrorCode.CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, mapped from the backend's response to a *sensitive operation*. In this file the only sensitive operation is `deleteAccount()` — the last statement of `deleteEverything()`.

So that code reaching this handler now means the popup succeeded, every word set, entry, session and the progress map are gone, and only the account remains. At that exact moment the user is told 「データは削除されていません」. The old wording over-warned; this one reassures, which is the worse direction and is the failure this pull request set out to remove.

The window is reachable rather than theoretical: between the two calls the removals are sequential, one `await` per document, over an unbounded collection. Nothing extends Firebase's recent-login window to cover that.

### The rest, in short

- The `cancelled` set is under-inclusive as well: `auth/user-mismatch` — the case `reauthenticateWithPopup` was *chosen* for — plus `auth/popup-blocked`, `auth/network-request-failed`, and `authAdapter.ts:44`'s uncoded `new Error('ログインしていません。')`. All leave nothing deleted and all get the part-way message.
- **The underlying problem is the inference.** `deleteEverything` is the only place that knows whether `reauthenticate()` returned; `Account.tsx` is reduced to guessing it from a provider error code, and that guess is wrong in both directions on first attempt. Throwing a typed error from the one place that knows removes the taxonomy entirely — and makes the branch unit-testable, which matters because `tests/component/` has twelve files and none of them is this route, so nothing currently exercises it.
- `tests/unit/accountData.test.ts:164` mocks `reauthenticate()` rejecting with `auth/requires-recent-login` — the same mistaken assumption, written into the suite. It passes anyway, because the assertions are about the removal ports being untouched, which is the right property. Worth rejecting with a code that call really raises, and adding the case the blocker is about: `deleteAccount()` failing *after* the data is gone.

### What is done well

- `reauthenticateWithPopup` over `signInWithPopup`, for the stated reason. Confirmed in the SDK. Easy to get wrong and invisible once wrong, since both open the same popup.
- Both new tests assert the property rather than the throw, and the description says why each weaker assertion would have passed with the bug present — `toHaveLength(0)` on the four removal ports, and `toHaveLength(2)` on the stuck read where asserting only that it threw would pass with the old code a million reads later. That is the distinction `CLAUDE.md` asks for, made without prompting.
- Removing each fix separately and confirming the other stayed green is the right way to show the two tests are not standing in for each other.

### Not verified

Static review. Tests were not run and the red-before-green runs were not reproduced. The claims about SDK error codes were checked against `node_modules/@firebase/auth` in this repository; the claim that the delete window can exceed the recent-login window is reasoning from the sequential loop in `deleteEverything`, not a measurement.

---
_Codex review._

src/routes/Account.tsx
kowinauyeung: @kowinauyeung
Upheld. Confirmed against the copy of `@firebase/auth` in this repository before changing anything:

```
CREDENTIAL_TOO_OLD_LOGIN_AGAIN: 'auth/requires-recent-login'
  → 'This operation is sensitive and requires recent authentication'
  → mapped from ServerError.CREDENTIAL_TOO_OLD_LOGIN_AGAIN
USER_MISMATCH: 'auth/user-mismatch'
  → 'The supplied credentials do not correspond to the previously signed in user'
```

So it is the backend's answer to a sensitive operation, and the only sensitive operation in this file is `deleteAccount()` — the last statement of `deleteEverything`. After my change that code can only reach the handler once every word set, entry, session and the progress map are gone.

**I moved a false statement rather than removing one, and made it worse in the process.** The old wording over-warned; mine reassures, at the one moment reassurance is unrecoverable — the reader retries later into an account holding nothing and has no way to learn that the deletion already happened.

Your point about the window is what makes it reachable rather than merely wrong. The removals are `for (const entry of entryList) await entries.remove(entry.id)`, one sequential await each, and nothing bounds how long that takes. Firebase's recent-login window does not stretch to cover it.

Fixed by the `accountData.ts` suggestion rather than by editing this list — see the reply there. The list is gone.

Fixed in f24fb83

src/routes/Account.tsx
kowinauyeung: @kowinauyeung
Agreed, and the omission you name first is the one that stings: `auth/user-mismatch` is the case the adapter's own docblock cites as the reason for choosing `reauthenticateWithPopup`. I wrote that sentence and then left the code it produces off the list.

`authAdapter.ts:44` throwing `new Error('ログインしていません。')` with no `code` is the same failure in its purest form — the branch is `code === …`, `code` is `undefined`, and a user who was never signed in is told a deletion failed part way.

> Enumerating provider error codes is a losing game — the list is provider-defined, changes between SDK versions, and is wrong in both directions here on first attempt.

Taken. The list is deleted rather than corrected; the fix is in `accountData.ts`.

Fixed in f24fb83

src/lib/accountData.ts
kowinauyeung: @kowinauyeung
Taken, and it is the right shape: the fact is free where it happens and unknowable where I was reading it.

```ts
try {
  await auth.reauthenticate();
} catch (cause) {
  throw new NothingDeleted(cause);
}
```

`Account.tsx` is now `cause instanceof NothingDeleted`. Everything past that point is genuinely part-way and keeps the existing message.

The class docblock records why it exists rather than only what it is — that `auth/requires-recent-login` is `CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, returned for the sensitive operation, so reading it as a refused re-authentication inverts the truth exactly. Without that, the next person has the same reason to reach for the code that I did.

**The coverage argument is the half that decides it.** There is no component test for this route, so the branch deciding whether a user is told their data is intact was the one thing here nothing exercised. In `accountData.ts` it is a function of its inputs, and `tests/unit` now carries both directions:

- `refuses to delete anything at all when the session cannot be proved` — now rejecting with `auth/user-mismatch`, and asserting `NothingDeleted` rather than a message.
- `does not report a late refusal as nothing having been deleted` — `deleteAccount()` rejecting with `auth/requires-recent-login` after the data is gone, asserting the failure is **not** `NothingDeleted` and that the removal ports **were** called. That is the state a user was about to be told did not exist.

Red before green: reverting to the bare `await auth.reauthenticate()` fails exactly the first, 498 passing.

Fixed in f24fb83

tests/unit/accountData.test.ts
kowinauyeung: @kowinauyeung
Agreed — the fixture was the assumption, and a passing test stating something that cannot happen is how the assumption survives review.

> `CLAUDE.md` asks that a test name say what breaks in production; this one's *fixture* says something that cannot happen.

Both changes made:

- The rejection is now `auth/user-mismatch`, which that call really raises and which is the case `authAdapter` chose it for.
- `does not report a late refusal as nothing having been deleted` added: `deleteAccount()` rejects with `auth/requires-recent-login` *after* the walk, and it asserts the failure is not `NothingDeleted` **and** that `entries.remove` was called five times and `progress.removeAll` once. Asserting only the error type would have passed with the data still there, which is the wrong half.

You are right that it would have caught the blocker. Writing the late case first would have made the early case's fixture obviously wrong, because the two cannot both use that code.

Fixed in f24fb83

src/infra/firebase/authAdapter.ts
kowinauyeung: @kowinauyeung
Thanks — and for going to `_reauthenticate` rather than taking the docblock's word. That the assertion is against the `sub` claim of the returned token is the detail that makes it a guarantee rather than a convention.

「both calls open the same popup and both look like they worked」 is the part worth keeping. It is why this could not have been caught by using the feature.

src/lib/accountData.ts
kowinauyeung: @kowinauyeung
Thanks, and the nitpick is taken — with your reason rather than for tidiness.

`WALK_STOPPED` is now one literal. The trap you describe is the real argument: `deleteEverything` drains twice, so a stuck walk during deletion already throws 「エクスポートが終わりませんでした」 from a path that has nothing to do with export. It is harmless only because `Account.tsx` discards `cause.message` there — which you checked, and which is exactly the kind of thing that stops being true the moment somebody decides to surface the real message.

The constant's comment says so, so the next person meets the problem before deciding to solve it.

Fixed in f24fb83

src/routes/Account.tsx
kowinauyeung: **Verified fixed** in `f24fb83`. The code list is gone entirely and the branch is now `cause instanceof NothingDeleted`, so the message no longer depends on guessing which half of `deleteEverything` failed from a provider string.

The comment left in its place is worth having: it records that `auth/requires-recent-login` is `CREDENTIAL_TOO_OLD_LOGIN_AGAIN`, raised for a sensitive operation, so the next reader knows why the taxonomy was abandoned rather than merely that it was.

_Codex re-review._

src/routes/Account.tsx
kowinauyeung: **Verified fixed**, and dissolved rather than corrected — which is the better outcome. `auth/user-mismatch`, `auth/popup-blocked`, `auth/network-request-failed` and the adapter's uncoded `new Error('ログインしていません。')` are now all handled, not because they were added to a list but because there is no list: anything thrown by `reauthenticate()` is wrapped, and nothing else is.

That is the version that stays correct across SDK upgrades.

_Codex re-review._

src/lib/accountData.ts
kowinauyeung: **Verified fixed**, exactly at the seam suggested. `deleteEverything` wraps only the `reauthenticate()` call and throws `NothingDeleted`; everything after it is genuinely part-way and keeps the existing message.

Two details better than what I proposed. The class carries `reason`, so the provider's error survives for the console instead of being swallowed by the wrapper. And the docblock explains the inversion — 「Reading it as 'the re-authentication was refused' inverts the truth exactly」 — which is the sentence that stops somebody re-deriving the original mistake.

One small thing, not worth a change on its own: `Account.tsx` logs `console.error(cause)`, which prints the wrapper. The original provider error is one level down on `.reason`. `super(msg, { cause })` would put it where DevTools renders it natively, if this is ever touched again.

The coverage argument also resolved itself. The decision moved out of the untested route and into `accountData.ts`, where it is a function of its inputs — so `tests/component` having no `Account` test no longer leaves it unexercised.

_Codex re-review._

tests/unit/accountData.test.ts
kowinauyeung: **Verified fixed on both counts.** The mock now rejects with `auth/user-mismatch` — a code that call really raises, and the one the adapter was chosen for — and `does not report a late refusal as nothing having been deleted` adds the case the blocker was about.

Resolving this thread, because both things asked for are here. The new test has a gap of its own, which I have raised separately rather than reopening this one, since it is a property of the new test rather than of the mock: **it stays green if `deleteEverything` swallows the `deleteAccount` failure.** Measured, one line to close. See the new comment on `tests/unit/accountData.test.ts`.

_Codex re-review._

src/infra/firebase/authAdapter.ts
kowinauyeung: Unchanged on the new head. Still holds.

_Codex re-review._

src/lib/accountData.ts
kowinauyeung: **Verified fixed** — one `WALK_STOPPED` constant, both throw sites using it.

The docblock added to it is the reason I raised it, written better than I raised it: 「this string is already reachable from a path it does not describe. Two copies would drift, and the copy that drifts is the one nobody is looking at.」

_Codex re-review._

tests/unit/accountData.test.ts
```diff
@@ -143,4 +151,89 @@ describe('deleteEverything', () => {
...
+   */
+  it('does not report a late refusal as nothing having been deleted', async () => {
+    const p = ports(5, 5, 5);
+    (p.auth.deleteAccount as ReturnType<typeof vi.fn>).mockRejectedValue(
+      Object.assign(new Error('stale'), { code: 'auth/requires-recent-login' }),
+    );
+
+    const failure = await deleteEverything(p).catch((cause: unknown) => cause);
+
+    expect(failure).not.toBeInstanceOf(NothingDeleted);
```
kowinauyeung, line 213: 🟡 **SUGGESTION** — this test cannot see the defect it is named after. Measured, not reasoned.

`not.toBeInstanceOf(NothingDeleted)` is satisfied by *anything that is not that class* — including `deleteEverything` returning normally. When it resolves, `failure` is the `{ entries, wordSets }` result object, which is not a `NothingDeleted`, and the two call-count assertions below hold either way because the removals ran regardless.

So the failure mode this test exists to prevent — a late refusal being reported as a safe one — has a neighbour it cannot distinguish: the account refusing to delete and `deleteEverything` swallowing it.

Injected exactly that on `f24fb83`, changing nothing else:

```ts
try { await auth.deleteAccount(); } catch { /* swallowed */ }
```

```
Tests  12 passed (12)
```

Twelve of twelve, including this one. The account now fails to delete, the user is told the deletion succeeded, and the suite is silent.

One line closes it, and `expect(failure).toBeInstanceOf(Error)` is enough because the result object is not one:

```ts
const failure = await deleteEverything(p).catch((cause: unknown) => cause);

expect(failure).toBeInstanceOf(Error);
expect(failure).not.toBeInstanceOf(NothingDeleted);
```

Verified in both directions with that line added, and only in the direction it should move:

```
no defect      → Tests  12 passed (12)
defect present → × does not report a late refusal as nothing having been deleted
                 Tests  1 failed | 11 passed (12)
```

Exactly one test goes red, and it is this one — so it fails for its own reason rather than taking others with it.

Worth saying that the sibling test does not have this problem: `refuses to delete anything at all when the session cannot be proved` uses `rejects.toBeInstanceOf(NothingDeleted)`, which cannot be satisfied by a resolution. It is only the negative form that needs the extra assertion, which is the sort of asymmetry that is invisible until something is injected against it — and `CLAUDE.md` asks for exactly that before a test is claimed to pass.

_Codex re-review._

kowinauyeung:
## Re-review — `db1388d` → `f24fb83`

**Verdict: `NEEDS DISCUSSION`** — the blocker is fixed and all six original threads are resolved. One new finding, on a test added in this round.

### The blocker is properly gone

Not corrected but **dissolved**. The error-code list is removed entirely and `Account.tsx` branches on `cause instanceof NothingDeleted`, thrown by `deleteEverything` around the `reauthenticate()` call and nothing else. That fixes the over-inclusive case (`auth/requires-recent-login`) and the under-inclusive ones (`auth/user-mismatch`, `auth/popup-blocked`, `auth/network-request-failed`, and the adapter's uncoded error) in one move, and it stays correct across SDK upgrades in a way an enumeration cannot.

The docblock on `NothingDeleted` explains the inversion rather than just recording the rule, which is what stops the original mistake being re-derived. The coverage concern resolved itself too: the decision moved out of the untested route into `accountData.ts`, where it is a function of its inputs.

`WALK_STOPPED` is one constant with both throw sites using it, and the comment on it says why the duplicate mattered.

### The new finding

`does not report a late refusal as nothing having been deleted` cannot see the defect it is named after. `not.toBeInstanceOf(NothingDeleted)` is satisfied by anything that is not that class — including `deleteEverything` returning normally — and the call-count assertions hold either way.

Injected the neighbouring defect on `f24fb83`, changing nothing else:

```ts
try { await auth.deleteAccount(); } catch { /* swallowed */ }
```

```
Tests  12 passed (12)
```

The account fails to delete, the user is told it succeeded, and the suite is silent. Adding `expect(failure).toBeInstanceOf(Error)` closes it — verified green without the defect and exactly one test red with it. Details inline.

The sibling test does not have this problem: `rejects.toBeInstanceOf(NothingDeleted)` cannot be satisfied by a resolution. It is only the negative form that needs the extra line.

### One note, not worth a change on its own

`Account.tsx` logs `console.error(cause)`, which prints the `NothingDeleted` wrapper; the provider's error is one level down on `.reason`. `super(msg, { cause })` would put it where DevTools renders it natively, if this is touched again.

---
_Codex re-review._

tests/unit/accountData.test.ts
kowinauyeung: @kowinauyeung
Upheld, and reproduced before agreeing rather than after. Injecting exactly the swallow you describe on `f24fb83`, changing nothing else:

```
Tests  12 passed (12)
```

Twelve of twelve, including the test named for the failure. So the test I wrote to stop a user being told their data is intact could not tell that case apart from the account silently failing to delete — and the second one is arguably worse, because the user is told the deletion *succeeded*.

The asymmetry is the part worth keeping:

> It is only the negative form that needs the extra assertion.

`rejects.toBeInstanceOf(NothingDeleted)` cannot be satisfied by a resolution, so the sibling test was never at risk. `not.toBeInstanceOf` is satisfied by everything else in the world, including the `{ entries, wordSets }` result object — and the two call-count assertions I added as the real property hold in both cases, because the removals genuinely ran. Three assertions, none of which could see it.

Taken as written, and measured in both directions after adding it:

```
no defect      → Tests  12 passed (12)
defect present → × does not report a late refusal as nothing having been deleted
                 Tests  1 failed | 11 passed (12)
```

One test red, and it is this one.

The comment beside it says why the first line is not redundant, since `expect(failure).toBeInstanceOf(Error)` immediately above `expect(failure).not.toBeInstanceOf(NothingDeleted)` reads like belt and braces and is not.

572 unit and component, typecheck clean.

Fixed in 1e11ad3

kowinauyeung: **Verified fixed** in `1e11ad3`, and I re-ran it against your commit rather than against my patch — the two are not the same thing, and a fix that works in the reviewer's copy is not evidence about the one that will merge.

```
A. as pushed, no defect                → Tests  12 passed (12)
B. deleteAccount failure swallowed     → × does not report a late refusal as nothing having been deleted
                                          Tests  1 failed | 11 passed (12)
```

One test red under the injected defect, and it is the one named for it.

I also ran a third case you did not claim, because "one test red" is only half of "only the tests you expect". Reverting the *other* fix — removing the `try`/`catch` that wraps `reauthenticate()` in `NothingDeleted` — while leaving `deleteAccount` alone:

```
C. NothingDeleted wrapper removed      → × refuses to delete anything at all when the session cannot be proved
                                          Tests  1 failed | 11 passed (12)
```

A different single test, and the right one. So the two guards are independent: each defect turns exactly its own test red and neither stands in for the other. That is the property that was actually at risk here, since the whole finding was one test being unable to distinguish two neighbouring failures.

The comment beside the assertion is doing the work it needs to. `expect(failure).toBeInstanceOf(Error)` immediately above `expect(failure).not.toBeInstanceOf(NothingDeleted)` reads like belt and braces, and the next person to tidy this file will delete it unless something explains that the second line is satisfied by the `{ entries, wordSets }` result object. It does.

`yarn test:unit`: 35 files, **572 passed**. `yarn typecheck`: clean across all three passes. Both match what you reported.

_Codex re-review._

kowinauyeung:
## Re-review — `f24fb83` → `1e11ad3`

**Verdict: `APPROVE`.** The last open finding is fixed and verified; all seven threads are now resolved.

`expect(failure).toBeInstanceOf(Error)` is exactly the line, and it was reproduced before being agreed to rather than after — which matters here more than usual, because the whole finding was that a passing test proves less than it appears to.

Re-ran it against `1e11ad3` rather than against my own patch:

| case | result |
| --- | --- |
| as pushed, no defect | `Tests 12 passed (12)` |
| `deleteAccount` failure swallowed | `× does not report a late refusal…` — 1 failed, 11 passed |
| `NothingDeleted` wrapper removed | `× refuses to delete anything at all…` — 1 failed, 11 passed |

The third case is one I added rather than one that was claimed. It matters because "one test goes red" is only half of what `CLAUDE.md` asks for — the other half is that **only** the expected test goes red. Each defect now turns exactly its own test red, and neither test stands in for the other. Given the finding was precisely that one test could not tell two neighbouring failures apart, that independence is the property worth having measured.

`yarn test:unit`: 35 files, 572 passed. `yarn typecheck`: clean across all three passes. Both match the reported figures.

### Where this pull request ended up

Both defects are real, both fixes are at the right seam, and the blocker was resolved by removing the thing that caused it rather than by correcting it — the provider error-code list is gone, not fixed, so it cannot drift back out of step with the SDK. `NothingDeleted` puts the decision in the only code that knows which half ran, which also moved it somewhere a unit test can reach.

One note carried over from a resolved thread, still not worth a change on its own: `Account.tsx` logs `console.error(cause)`, which prints the wrapper; the provider's error is one level down on `.reason`. `super(msg, { cause })` would put it where DevTools renders it natively, if this is ever touched again.

Nothing outstanding.

---
_Codex re-review._

PR #34, opened 2026-08-14
fix(migration): check the claim instead of writing a collection nothing reads

### LINKED ISSUE #26: upload.mjs still writes allowedUsers and reports it as confirming access

kowinauyeung:
Closes #26.

`migration/upload.mjs` wrote `allowedUsers/{uid}` and printed:

```
許可    : allowedUsers/<uid> を確認しました
```

Nothing has read that collection since the gate became a custom claim. The write went on succeeding and the line went on saying "confirmed" — a confirmation of something that had stopped being true, printed by the tool an operator trusts most at the moment they most need it to be right.

The failure that produces is not subtle and is not fixed by retrying: the script reports access confirmed, the operator opens the app, and every screen reports that it could not load. That is the wall #22 exists to make legible, reached by following the tooling's own success message — and by then 67 documents have been written.

### Checked, not deleted

Deleting the write alone would leave the script silent about the one thing most likely to go wrong. It now checks the claim instead, and refuses:

```js
if (owner.customClaims?.allowed !== true) {
  console.error(`許可    : ${ownerEmail} に allowed クレームがありません。`);
  console.error(`          yarn allow ${ownerEmail}${env === 'prod' ? ' prod' : ''}`);
  console.error('          を実行し、サインアウトしてからサインインし直してください。');
  process.exit(1);
}
```

It costs nothing: `getUserByEmail` at `:133` already returned the record. And it fails **before** the 67 documents are written rather than after.

Same field on both sides, checked rather than assumed — `admin/allow-user.ts` writes `{ ...user.customClaims, allowed: true }`, this reads `owner.customClaims?.allowed`.

### What it cannot check, said rather than implied

A claim reaches a client only in a freshly minted ID token, so an account granted access while signed in still sees nothing until it signs out and back in. Nothing server-side can observe that. The script does not pretend to — it prints the caveat under the success line, because that is the half most likely to be hit.

### Documentation

`migration/README.md` described the replaced mechanism in two places.

The production rollout also went from five steps to seven. It never mentioned `yarn allow` or signing in again at all — the two steps without which the rest cannot work — and it ended at "verify the count is 67". Using the app afterwards is now step 7, because everything above it runs through the Admin SDK, which bypasses rules entirely: nothing before that step proves the paths, `ownerUid` and the deployed rules agree.

### Tests

**None, and the reason is not that it seemed unnecessary.** `CLAUDE.md`'s layer table covers `src`, not the one-shot migration scripts, and `upload.mjs` has never had a test — `tests/unit/migrationOutput.test.ts` covers the field mapping, not the script's control flow. Exercising this condition needs Auth Admin, which the Firestore emulator does not provide, against a real project.

What was done instead: `node --check`, a repository-wide grep confirming the only remaining `allowedUsers` references are the two historical explanations, and reading the claim's shape off `allow-user.ts` rather than assuming it.

`eslint` clean (`migration/**/*.mjs` is linted), prettier clean, 500 unit passing.

### Base

`release/0.1.0`, not `develop`. This is needed for the 0.1.0 rollout itself: the migration follows the first production deploy, and this line sits directly on that path.

### DIFF
--- a/migration/README.md
+++ b/migration/README.md
@@ -164,14 +164,21 @@ production has been loaded and verified.
 
 Production has never been written to and has no rules deployed. In order:
 
-1. `yarn rules:prod` — deploy `firestore.rules` first, so the collection is
-   never briefly writable by anyone but the owner. (Or run the production deploy
-   workflow, which deploys rules before hosting for the same reason.)
+1. Run _Deploy (production)_. It deploys rules **then** hosting, which is the
+   order that matters: a client briefly older than its rules fails closed, and
+   the other way round fails open.
+
+   **Hosting is what makes step 2 possible.** On a project that has never been
+   deployed there is no app to sign into, so this cannot be shortened to
+   `yarn rules:prod` — that deploys rules only, and leaves you at step 2 with
+   nothing to open. It is the right command later, when hosting is already live
+   and only the rules have changed.
+
 2. Sign into the production app once, so the account exists and `--owner` can
    resolve to a uid.
 3. `yarn allow you@example.com prod`, **then sign out and sign back in.** The
    rules gate on a custom claim, and a claim reaches a client only in a freshly
-   minted ID token — an open session stays denied for up to an hour. Step 4
+   minted ID token — an open session stays denied for up to an hour. Step 5
    refuses to run until the claim is set, but it cannot tell whether you have
    signed in again since.
 4. Fill the accents — see above. Cheap now, 67 hand edits later.
--- a/migration/README.md
+++ b/migration/README.md
@@ -55,10 +55,18 @@ node migration/upload.mjs prod --owner you@example.com --confirm      # goitei
 
 `--owner` is mandatory. Entries live at `users/{uid}/entries/{autoId}`, and the
 uid is resolved from that address through the Admin SDK rather than guessed, so
-the account must have signed into the app at least once. The uploader also
-ensures `allowedUsers/{uid}` exists — the security rules gate every path on it
-and no client can write that collection, so uploading a notebook its owner would
-then be denied access to is the easy mistake to make.
+the account must have signed into the app at least once.
+
+**The uploader checks that the owner carries the `allowed` custom claim and
+refuses to run without it.** It used to _write_ `allowedUsers/{uid}`, back when
+the rules read that collection; they gate on the claim now and nothing reads it,
+so the write had stopped granting anything while still reporting that it had.
+Uploading a notebook its owner is then denied access to is the easy mistake, and
+the check is what stops it — before 67 documents are written rather than after.
+
+What it cannot check is the other half: a claim reaches a client only in a
+freshly minted ID token, so an account granted access while signed in still sees
+nothing until it signs out and back in. Nothing server-side can observe that.
 
 `upload.mjs` uses `set()` without merge, so a document is fully replaced and
 stale fields disappear. It is still idempotent, but by a different mechanism:
@@ -157,12 +165,21 @@ production has been loaded and verified.
 Production has never been written to and has no rules deployed. In order:
 
 1. `yarn rules:prod` — deploy `firestore.rules` first, so the collection is
-   never briefly writable by anyone but the owner.
+   never briefly writable by anyone but the owner. (Or run the production deploy
+   workflow, which deploys rules before hosting for the same reason.)
 2. Sign into the production app once, so the account exists and `--owner` can
    resolve to a uid.
-3. Fill the accents — see above. Cheap now, 67 hand edits later.
-4. `node migration/upload.mjs prod --owner you@example.com --confirm` — loads
-   the 67 entries and adds the `allowedUsers` record.
-5. Verify the count is 67 and spot-check one entry with a 📌 context section.
+3. `yarn allow you@example.com prod`, **then sign out and sign back in.** The
+   rules gate on a custom claim, and a claim reaches a client only in a freshly
+   minted ID token — an open session stays denied for up to an hour. Step 4
+   refuses to run until the claim is set, but it cannot tell whether you have
+   signed in again since.
+4. Fill the accents — see above. Cheap now, 67 hand edits later.
+5. `node migration/upload.mjs prod --owner you@example.com --confirm` — loads
+   the 67 entries.
+6. Verify the count is 67 and spot-check one entry with a 📌 context section.
+7. **Then use the app** — add, edit and delete a word. Everything above runs
+   through the Admin SDK, which bypasses rules entirely, so nothing before this
+   proves the paths, `ownerUid` and the deployed rules agree.
 
 Step 1 before step 2 is the whole point of the ordering; do not reverse it.
--- a/migration/upload.mjs
+++ b/migration/upload.mjs
@@ -165,13 +165,28 @@ console.log(`件数  : ${entries.length}`);
 // minted ID token, so an account granted access while signed in still cannot
 // read anything until it signs in again. Nothing server-side can observe that,
 // which is why the message says it rather than checking it.
+//
+// **Both halves, because the gate is both.** `isAllowed()` is
+// `signedIn() && request.auth.token.allowed == true`, and `signedIn()` requires
+// `email_verified`. Standing in for one of them would report an account ready
+// when the rules will still refuse it — which is the same failure this block
+// replaces, in a smaller size.
+//
+// Low impact and worth saying so: `client.ts` registers Google as the only
+// provider and a Google account arrives verified. This is the check matching the
+// rule, not a case anyone is likely to hit.
+if (owner.emailVerified !== true) {
+  console.error(`許可    : ${ownerEmail} のメールアドレスが未確認です。`);
+  console.error('          ルールは email_verified を要求します。');
+  process.exit(1);
+}
 if (owner.customClaims?.allowed !== true) {
   console.error(`許可    : ${ownerEmail} に allowed クレームがありません。`);
   console.error(`          yarn allow ${ownerEmail}${env === 'prod' ? ' prod' : ''}`);
   console.error('          を実行し、サインアウトしてからサインインし直してください。');
   process.exit(1);
 }
-console.log(`許可    : allowed クレームを確認しました`);
+console.log('許可    : email_verified と allowed クレームを確認しました');
 console.log('          （付与直後の場合は、一度サインアウトして入り直す必要があります）');
 
 // One read of the collection buys idempotency: without the key -> id map a
--- a/migration/upload.mjs
+++ b/migration/upload.mjs
@@ -146,15 +146,33 @@ console.log(`認証  : ${how}`);
 console.log(`所有者: ${ownerEmail} (${owner.uid})`);
 console.log(`件数  : ${entries.length}`);
 
-// The rules gate every path on allowedUsers/{uid} existing, and no client can
-// write that collection, so uploading a notebook the owner would then be denied
-// access to is the obvious way to get this wrong. Done here rather than by hand
-// because the uid is only resolvable through the Admin SDK anyway.
-await db
-  .collection('allowedUsers')
-  .doc(owner.uid)
-  .set({ email: ownerEmail, addedAt: Timestamp.now() }, { merge: true });
-console.log(`許可    : allowedUsers/${owner.uid} を確認しました`);
+// **This checks. It used to grant, and the difference went unreported.**
+//
+// Until the claim gate landed, the rules read `allowedUsers/{uid}` and this
+// script wrote it, so an owner could not end up locked out of a notebook it had
+// just uploaded. Nothing has read that collection since. The write went on
+// succeeding and the line below went on saying 「確認しました」 — a confirmation
+// of something that had stopped being true, printed by the tool an operator
+// trusts most at the one moment they most need it to be right.
+//
+// The failure that produces is not subtle and not recoverable by retrying: the
+// script reports access confirmed, the operator opens the app, and every screen
+// says it could not load. Checking the claim instead costs nothing —
+// `getUserByEmail` above already returned it — and it fails *before* 67
+// documents are written rather than after.
+//
+// It cannot see the other half. A claim reaches a client only in a freshly
+// minted ID token, so an account granted access while signed in still cannot
+// read anything until it signs in again. Nothing server-side can observe that,
+// which is why the message says it rather than checking it.
+if (owner.customClaims?.allowed !== true) {
+  console.error(`許可    : ${ownerEmail} に allowed クレームがありません。`);
+  console.error(`          yarn allow ${ownerEmail}${env === 'prod' ? ' prod' : ''}`);
+  console.error('          を実行し、サインアウトしてからサインインし直してください。');
+  process.exit(1);
+}
+console.log(`許可    : allowed クレームを確認しました`);
+console.log('          （付与直後の場合は、一度サインアウトして入り直す必要があります）');
 
 // One read of the collection buys idempotency: without the key -> id map a
 // second run cannot tell an already-uploaded note from a new one.


migration/README.md
```diff
@@ -157,12 +165,21 @@ production has been loaded and verified.
...
+   workflow, which deploys rules before hosting for the same reason.)
 2. Sign into the production app once, so the account exists and `--owner` can
    resolve to a uid.
-3. Fill the accents — see above. Cheap now, 67 hand edits later.
-4. `node migration/upload.mjs prod --owner you@example.com --confirm` — loads
-   the 67 entries and adds the `allowedUsers` record.
-5. Verify the count is 67 and spot-check one entry with a 📌 context section.
+3. `yarn allow you@example.com prod`, **then sign out and sign back in.** The
+   rules gate on a custom claim, and a claim reaches a client only in a freshly
+   minted ID token — an open session stays denied for up to an hour. Step 4
```
kowinauyeung: **Stale cross-reference after the renumbering.** This says "Step 4 refuses to run until the claim is set", but step 4 is now "Fill the accents" — the upload that refuses is step 5.

Worth fixing precisely because of what this pull request is about. The defect it removes is a line of operator output that went on describing a mechanism after that mechanism had been replaced; this is the same shape, one document over, in the list an operator follows step by step during a one-shot production migration.

```suggestion
   minted ID token — an open session stays denied for up to an hour. Step 5
```

migration/README.md
```diff
@@ -157,12 +165,21 @@ production has been loaded and verified.
 Production has never been written to and has no rules deployed. In order:
 
 1. `yarn rules:prod` — deploy `firestore.rules` first, so the collection is
-   never briefly writable by anyone but the owner.
+   never briefly writable by anyone but the owner. (Or run the production deploy
+   workflow, which deploys rules before hosting for the same reason.)
```
kowinauyeung: Step 2 asks the operator to sign into the production app, but the primary path here — `yarn rules:prod` — deploys rules only. On a project that "has never been written to and has no rules deployed", there is no app to sign into yet, so step 2 cannot happen until hosting is deployed.

The parenthetical added here is what actually makes step 2 reachable, since the production deploy workflow deploys hosting after rules. But it reads as an alternative way to accomplish step 1 rather than as a prerequisite of step 2, so an operator taking the shorter-looking path arrives at step 2 with nothing to sign into.

Suggest making the dependency explicit — either fold hosting into step 1, or note under step 2 that it requires the production deploy to have run.

migration/upload.mjs
```diff
@@ -146,15 +146,33 @@ console.log(`認証  : ${how}`);
...
+// script reports access confirmed, the operator opens the app, and every screen
+// says it could not load. Checking the claim instead costs nothing —
+// `getUserByEmail` above already returned it — and it fails *before* 67
+// documents are written rather than after.
+//
+// It cannot see the other half. A claim reaches a client only in a freshly
+// minted ID token, so an account granted access while signed in still cannot
+// read anything until it signs in again. Nothing server-side can observe that,
+// which is why the message says it rather than checking it.
+if (owner.customClaims?.allowed !== true) {
```
kowinauyeung, line 183: The gate is two conditions, and this checks one of them. `firestore.rules:73` reads `signedIn() && request.auth.token.allowed == true`, and `signedIn()` at `:21` is `request.auth != null && request.auth.token.email_verified == true`.

The argument in the comment above applies unchanged to the other half: `getUserByEmail` already returned `owner.emailVerified`, so checking it costs nothing either.

Low impact in practice, and worth saying so plainly: `src/infra/firebase/client.ts:111` registers `GoogleAuthProvider` as the only provider, and a Google account arrives with its email verified. So this is about the check matching the rule it stands in for, not about a failure anyone is likely to hit.

migration/upload.mjs
```diff
@@ -146,15 +146,33 @@ console.log(`認証  : ${how}`);
...
+  console.error(`許可    : ${ownerEmail} に allowed クレームがありません。`);
+  console.error(`          yarn allow ${ownerEmail}${env === 'prod' ? ' prod' : ''}`);
+  console.error('          を実行し、サインアウトしてからサインインし直してください。');
+  process.exit(1);
+}
+console.log(`許可    : allowed クレームを確認しました`);
```
kowinauyeung: Template literal with nothing interpolated — a plain quoted string reads the same and says less about what it is.

migration/upload.mjs
```diff
@@ -146,15 +146,33 @@ console.log(`認証  : ${how}`);
...
+// `getUserByEmail` above already returned it — and it fails *before* 67
+// documents are written rather than after.
+//
+// It cannot see the other half. A claim reaches a client only in a freshly
+// minted ID token, so an account granted access while signed in still cannot
+// read anything until it signs in again. Nothing server-side can observe that,
+// which is why the message says it rather than checking it.
+if (owner.customClaims?.allowed !== true) {
+  console.error(`許可    : ${ownerEmail} に allowed クレームがありません。`);
+  console.error(`          yarn allow ${ownerEmail}${env === 'prod' ? ' prod' : ''}`);
```
kowinauyeung, line 185: Deriving the `prod` suffix from `env` rather than printing a fixed command is the right amount of care here. `admin/allow-user.ts` rejects anything other than `prod` and `--revoke` precisely because `--prod` would silently grant on dev, so a remediation line that gets this wrong sends the operator into exactly the failure that script guards against.

migration/README.md
```diff
@@ -157,12 +165,21 @@ production has been loaded and verified.
...
+3. `yarn allow you@example.com prod`, **then sign out and sign back in.** The
+   rules gate on a custom claim, and a claim reaches a client only in a freshly
+   minted ID token — an open session stays denied for up to an hour. Step 4
+   refuses to run until the claim is set, but it cannot tell whether you have
+   signed in again since.
+4. Fill the accents — see above. Cheap now, 67 hand edits later.
+5. `node migration/upload.mjs prod --owner you@example.com --confirm` — loads
+   the 67 entries.
+6. Verify the count is 67 and spot-check one entry with a 📌 context section.
+7. **Then use the app** — add, edit and delete a word. Everything above runs
```
kowinauyeung, line 188: This step closes a real gap rather than adding ceremony. Every step above it runs through the Admin SDK, which bypasses rules entirely, so a rollout that ended at "verify the count is 67" verified that documents exist and nothing about whether the deployed rules, the paths and `ownerUid` agree — which is the one thing the operator is about to depend on.

kowinauyeung:
## Review summary

**Verdict: approve.** No blockers. Two documentation findings are worth clearing before this runs against production, since the runbook is the artefact an operator follows line by line during a migration that happens once.

The change itself is right, and right for the reason given: an operator tool that prints a confirmation is making a claim, and this one had been making a claim about a mechanism that no longer existed. Replacing the write with a check — rather than deleting the write and leaving the script silent — keeps the script saying something true about the thing most likely to go wrong, and moves the failure ahead of the 67 documents instead of after them.

### Findings

| Severity | Location | Finding |
| --- | --- | --- |
| Suggestion | `migration/README.md:174` | "Step 4 refuses to run until the claim is set" is stale after the renumbering — the upload is step 5, step 4 is filling the accents. |
| Suggestion | `migration/README.md:167-171` | Step 2 requires a deployed app, but step 1's primary path deploys rules only. The parenthetical is what makes step 2 reachable, phrased as an alternative rather than a prerequisite. |
| Nitpick | `migration/upload.mjs:168` | The rule gates on `email_verified` **and** `allowed`; this checks `allowed`. `owner.emailVerified` is already in hand. Near-unreachable in practice — Google is the only provider. |
| Nitpick | `migration/upload.mjs:174` | Template literal with no interpolation. |

The first is worth naming for what it is: the same class of defect this pull request exists to remove, in the neighbouring file. A line that describes a mechanism the surrounding text has already changed.

### What holds up well

- The check runs before `collection.get()` and before any batch, so the claim in the description — fails before the documents, not after — is true of the code as written.
- `!== true` matches the rules' `== true`, so a claim set to the string `"true"` is refused on both sides rather than passing here and failing there.
- The remediation line derives its `prod` suffix from `env`. `admin/allow-user.ts` rejects unknown arguments specifically because `--prod` would grant on dev and say so only in its last line, so a printed command that got this wrong would walk the operator into the failure that script was written to prevent.
- The caveat about token freshness is printed as a caveat rather than folded into the success line. The distinction between what was verified and what cannot be verified is the whole point of the change, and the output keeps it.
- Step 7 of the rollout closes a genuine gap: everything above it runs through the Admin SDK, which bypasses rules, so a rollout ending at "verify the count is 67" proved that documents exist and nothing about whether the deployed rules and `ownerUid` agree.

### Coverage against #26

| Item in #26 | This pull request | Closed |
| --- | --- | --- |
| `upload.mjs:154` writes `allowedUsers` | Removed, replaced by a claim check | Yes |
| `:157` prints a confirmation of nothing | Printed only after a check that can fail | Yes |
| `:149` comment describes the replaced mechanism | Rewritten | Yes |
| `migration/README.md:59` describes it too | Rewritten | Yes |
| `migration/README.md:165` production rollout | Rewritten as seven steps | Yes, with the stale step reference above |
| Suggested fix: check the claim, **or** state that access is `yarn allow` plus signing in again | Both | Yes |

### Verification performed during review

`node --check migration/upload.mjs` passes and `prettier --check` is clean on both files. A repository-wide grep confirms the only surviving `allowedUsers` references are the four historical explanations (`firestore.rules:39`, `admin/allow-user.ts:12`, `migration/upload.mjs:151`, `migration/README.md:61`) plus the rules tests that exist to prove the collection is unreachable from any client. The emulator and end-to-end suites were not run — this touches no rule and no adapter.

On the absence of tests: the reasoning given holds. The layer table in `CLAUDE.md` addresses `src`, not one-shot migration scripts, and exercising this branch needs Auth Admin, which the Firestore emulator does not provide.

---

_Cross-checked against a second model, which surfaced no findings beyond these._

migration/README.md
kowinauyeung: @kowinauyeung
Upheld, and the framing is the part I want on the record:

> The defect it removes is a line of operator output that went on describing a mechanism after that mechanism had been replaced; this is the same shape, one document over.

I renumbered five steps into seven and did not re-read what pointed at them — in the list an operator follows once, in production, without a second run to notice it. The same class of mistake, made while fixing an instance of it.

Now "Step 5", which is the upload.

Fixed in bd8cb0e

migration/README.md
kowinauyeung: @kowinauyeung
Upheld, and this is the more serious of the two. Verified rather than accepted: `package.json:34` is `firebase deploy --only firestore:rules --project prod`. Rules only. The section opens 「Production has never been written to and has no rules deployed」, so at step 2 there is genuinely nothing to open.

You are right about why the parenthetical did not fix it. I wrote it as an alternative way to do step 1, and it is a prerequisite of step 2 — so the reader taking the shorter-looking path arrives at a step they cannot perform, with nothing on the page explaining why.

Restructured rather than annotated:

```
1. Run *Deploy (production)*. It deploys rules **then** hosting, which is the
   order that matters: a client briefly older than its rules fails closed, and
   the other way round fails open.

   **Hosting is what makes step 2 possible.** On a project that has never been
   deployed there is no app to sign into, so this cannot be shortened to
   `yarn rules:prod` — that deploys rules only, and leaves you at step 2 with
   nothing to open. It is the right command later, when hosting is already live
   and only the rules have changed.
```

`yarn rules:prod` is kept and given its actual scope rather than removed, since it is the right command once hosting exists.

Fixed in bd8cb0e

migration/upload.mjs
kowinauyeung: @kowinauyeung
Upheld. Read off the file rather than from the description:

```
firestore.rules:73   isAllowed() = signedIn() && request.auth.token.allowed == true
firestore.rules:21   signedIn()  = request.auth != null && request.auth.token.email_verified == true
```

And you are right that the argument transfers unchanged — `getUserByEmail` already returned `owner.emailVerified`, so the second half is as free as the first.

Both halves are now checked, as separate conditions with separate messages, because the remedies differ: `yarn allow` does nothing for an unverified address.

**The reason for the change is worth separating from its impact, and I have written both down.** Standing in for one half of a two-part gate reports an account ready when the rules will still refuse it — which is the same defect this block replaces, one size smaller. Your note that `client.ts` registers Google as the only provider and a Google account arrives verified is in the comment too, so the next reader does not mistake a correctness fix for a live hazard.

Fixed in bd8cb0e

migration/upload.mjs
kowinauyeung: @kowinauyeung
Agreed — the backticks were left over from a version of the line that did interpolate. Plain quotes now.

The three neighbouring lines that keep their template literals all interpolate something, so the file no longer implies a variable that is not there.

Fixed in bd8cb0e

migration/upload.mjs
kowinauyeung: @kowinauyeung
Thanks — and for connecting it to `admin/allow-user.ts` rejecting `--prod`. That is exactly the reasoning: the script refuses the misspelling because it would silently grant on dev, so a remediation line that prints the wrong command sends the operator into the failure that refusal exists to prevent.

A printed command is as much a part of the interface as an argument parser, and only one of the two had been thought about before this.

migration/README.md
kowinauyeung: @kowinauyeung
Thanks. 「verified that documents exist and nothing about whether the deployed rules, the paths and `ownerUid` agree」 is a better statement of the gap than the one I wrote in the step.

It is also the step most likely to be dropped as ceremony, which is why the reason sits in the list rather than in a pull request description — the reader is an operator part-way through a one-shot migration, looking for a reason to stop.

migration/README.md
kowinauyeung: Verified against `bd8cb0e` rather than the claim: `migration/README.md:181` now reads "Step 5", and `:185` is the `node migration/upload.mjs prod` line — so the reference and the step it points at agree. The remaining numbering (1 deploy, 2 sign in, 3 allow, 4 accents, 5 upload, 6 count, 7 use the app) is contiguous.

Resolving.

migration/upload.mjs
kowinauyeung: Verified in `bd8cb0e`. Both halves are now checked as separate conditions with separate messages — `owner.emailVerified !== true` at `:178`, `owner.customClaims?.allowed !== true` at `:183` — and the verified half is checked first, so an unverified address is reported as an unverified address rather than as a missing claim.

Both still sit ahead of `collection.get()` and every batch, so the fail-before-the-documents property survives the addition.

`:189` now names both: 「email_verified と allowed クレームを確認しました」. The success line and the checks say the same thing, which is the property this pull request is about.

One thing I checked and am satisfied with: the `email_verified` message states the requirement without printing a remediation command, unlike the `allowed` one. That is correct rather than incomplete — no command in this repository verifies an address, and inventing one would be the failure mode this change exists to remove.

Resolving.

migration/upload.mjs
kowinauyeung: Verified in `bd8cb0e`: `:189` uses plain quotes.

I also checked the claim about the neighbours rather than taking it — grepping the file for backtick-delimited strings with no `${` leaves only comment prose and one single-quoted string that happens to contain backticks (`:120`). No non-interpolating template literal remains in the file.

Resolving.

migration/README.md
kowinauyeung: Verified in `bd8cb0e`. Step 1 is now _Deploy (production)_, and `:171-175` states the dependency in the direction that matters — hosting is what makes step 2 possible — rather than leaving it to be inferred from a parenthetical. Naming `yarn rules:prod` as the command for later, when hosting is live and only the rules changed, keeps it useful instead of merely removing it.

One leftover, and explicitly not a reason to reopen this: `:192` still says

> Step 1 before step 2 is the whole point of the ordering; do not reverse it.

That sentence guarded the previous shape, where step 1 was a manual `yarn rules:prod` and reversing it meant signing in against a project with no rules deployed. Under the new step 1 the rules-before-hosting order lives inside the workflow, and an operator cannot reach step 2 before step 1 at all — there is nothing to sign into. So the warning now guards a mistake that is no longer available.

Harmless, and worth a moment because of what this pull request is: a line that describes a hazard in the shape it used to have. Either point it at the ordering that still matters — everything before step 7 bypasses rules — or drop it, since `:167-169` already carries the rules-before-hosting reasoning.

Resolving the finding as raised; the above is a note, not a request.

migration/upload.mjs
kowinauyeung: Agreed on the framing — a printed command is part of the interface, and the argument parser was the only half that had been designed. Nothing further here.

Resolving.

migration/README.md
kowinauyeung: Right about where the reason belongs. An operator part-way through a one-shot migration, reading a list, is exactly the reader who will skip a step whose justification lives in a pull request description they are not looking at.

Resolving.

kowinauyeung:
## Re-review — `bd8cb0e`

**Verdict: approve.** All four findings verified fixed against the commit rather than the replies. Every thread resolved.

| Finding | Fix | Verified |
| --- | --- | --- |
| `README:174` stale "Step 4" | Now "Step 5" at `:181`, and `:185` is the upload command | Read the file; numbering 1–7 contiguous |
| `README:167-171` hosting prerequisite | Step 1 is now _Deploy (production)_, with `:171-175` stating hosting is what makes step 2 possible; `yarn rules:prod` kept as the later command | Read the file |
| `upload.mjs:168` half a gate | `owner.emailVerified` at `:178`, `owner.customClaims?.allowed` at `:183`, separate messages, verified-half first | Read the file; both still precede `collection.get()` and every batch |
| `upload.mjs:174` template literal | Plain quotes at `:189` | Grepped the file — no non-interpolating template literal remains |

`node --check` passes and `prettier --check` is clean on both files at `bd8cb0e`. Emulator and end-to-end not run; still nothing here touches a rule or an adapter.

### One note, carried in the thread rather than as a finding

`migration/README.md:192` — "Step 1 before step 2 is the whole point of the ordering; do not reverse it" — guarded the previous shape, where step 1 was a manual `yarn rules:prod`. Under the new step 1 an operator cannot reach step 2 first at all, because there is nothing to sign into, so the warning now names a mistake that is no longer available. Harmless, and left to your judgement; `:167-169` already carries the rules-before-hosting reasoning it was standing in for.

### On the fix to the second finding

Worth saying plainly, because it is the part a reviewer usually does not get: the response did not do the minimum. The minimum was to move the parenthetical into step 2's prerequisite. What is there instead keeps `yarn rules:prod` and says when it is the right command — later, when hosting is live and only the rules have changed — so the shorter path is documented rather than deleted. A runbook that only forbids things teaches an operator less than one that says when each command is correct.

Nothing outstanding from my side.

PR #35, opened 2026-08-14
fix(rules): bound every validator, and require the fields every listing orders by

kowinauyeung:
Three of the four blockers from the production-readiness review. The fourth is legal wording and is left for the author to decide.

## The allowlist argument was only applied to entries

`hasOnly` closed the unknown-name half of the hole, and a later round bounded the fields `validEntry` permits. **`validWordSet` and `validSession` were left naming fields and not reading them** — which is the same megabyte under a different key, one collection over.

Reproduced before fixing, against the emulator, and it is exactly as reported:

```
× refuses an oversized value in a word set or session field too
  70 passing
```

`level`, `publishedId`, `publishedVersion`, `copiedFrom`, `mode`, `startedAt` and both timestamp pairs are now bounded. Green afterwards.

The comment I wrote in the round that introduced `hasOnly` says *"being on this list is permission to send a field, not a limit on it"* — and then bounds one validator of three. That is worth naming: the reasoning was written down and applied incompletely, which is harder to catch than reasoning that is absent.

## Optional in the rules, ordered-by in the adapter

`{ ownerUid, headword }` passed every check. All three adapters order their listings by a timestamp — `entryRepo` and `wordSetRepo` by `createdAt`, `progressRepo` by `finishedAt` — and **Firestore omits a document that lacks the field it is ordering by**.

So a row written without one is invisible to paging: it is never returned, therefore never exported, therefore never deleted. `Account.tsx` offers "export everything" and "delete everything", and the privacy policy says account deletion removes the user's data. Neither was true for such a row.

`createdAt`, `updatedAt` and `finishedAt` are now required on create, alongside `headword`, and the rules say why the reason is different: this combination does not break a query, it breaks a promise.

**Six existing tests went red, and that is the second finding.** The `entry()`, `wordSet()` and `session()` fixtures never sent timestamps — documents no write path in the application can produce. The same shape as the ISO-string `createdAt` an earlier round found in the adapter-fidelity test. The fixtures now send what the adapters send.

### What is deliberately not in this change

The review also asked for a full-collection scan in export and delete, deletion of `users/{uid}`, and an offboarding flow with token revocation.

- **The scan** protects rows already written without timestamps. Production has no data at all, and after this change none can be written, so there is nothing for it to find. It matters only for dev.
- **`users/{uid}`** cannot exist: nothing writes it, and its `create, update` was removed when the field allowlist landed.
- **Offboarding and revocation** is real and is a multi-account concern. Worth an issue, not a release blocker while there is one account.

## A red light on the release, on the one pull request where every light gets read

`deploy-dev.yml` now runs on pull requests into `develop` only.

The deploy job checks out the pull request's **base** on purpose — it holds a credential, so it must not run the pull request's own code. That assumes the base carries the toolchain, and a release pull request into `main` broke the assumption rather than the rule: `main` had no `.nvmrc`, so `setup-node` failed four seconds in.

It was never a required check and never blocked anything. That is what made it worth fixing: a red light nobody is meant to act on, on a release. A dev preview has no business there.

## The liability clause said the opposite of the comment above it

`terms.ts` excluded liability except for intent and gross negligence. Read the other way round, that excluded **everything** for ordinary negligence — and 消費者契約法 8条1項1号 is about that shape: a clause wholly excluding a business's liability, not only one excluding it for intent and gross negligence.

The file's own docblock claimed **"there is no blanket disclaimer … The limitation below is bounded instead"**. The carve-out that made the comment feel true was for the wrong half.

Now bounded by the kind of damage rather than removed:

```
本サービスは無償で提供しています。
故意または重過失 → その全部について責任を負います。
軽過失          → 通常生じうる損害の範囲に限り責任を負います。
                  特別損害、逸失利益、間接損害については負いません。
```

**Deliberately no monetary cap.** The service is free, so a limit at "the amount paid" is a total exclusion wearing a number. That is written into the docblock, because it is the thing most likely to be added back as a tidy-up.

The docblock also now records that it had not described the text beneath it. A comment asserting something the code does not do is worse than no comment: the next reader believes it.

## The operator could not be identified

`privacy.ts` named the operator as `Kowin` and offered no way to obtain a legal name or address. Added:

> 運営者の氏名および住所は、お問い合わせフォームからのご請求に応じて、遅滞なくお知らせします。

**No address is published.** APPI accepts a state in which the person can obtain the information on request, so this satisfies it the same way the no-email-address decision does — the form is the channel, and it stays revocable.

## Still not legal advice

Both documents' docblocks still say they have not been reviewed by a lawyer, and that remains true. What changed is that each now does what the comment above it says it does, using a structure that is the conventional one — which is not the same as being enforceable.

## Gate

72 emulator (from 70), 573 unit and component, lint clean, prettier clean.


firestore.rules
```diff
@@ -105,6 +105,15 @@ service cloud.firestore {
...
     // not an entry, and the form refuses it too.
+    //
+    // **`createdAt`, `updatedAt` and `finishedAt` are the other exceptions, and
+    // the reason is not schema tidiness.** Every `list()` in the adapters orders
+    // by one of them, and Firestore omits a document that lacks the field it is
+    // ordering by. So a row written without one is invisible to paging — which
+    // means it is never exported and never deleted, while the account screen and
+    // the privacy policy both promise otherwise. Optional here and ordered-by
+    // there is a combination that quietly breaks a legal claim, so they are
+    // required at the only point that can enforce it.
```
kowinauyeung, line 117: 🟡 **The justification is true for two of the three fields, and this file has a test suite named after comments being true.**

"Every `list()` in the adapters orders by one of them" holds for `createdAt` (`entryRepo.ts:51`, `wordSetRepo.ts:43`) and for `finishedAt` (`progressRepo.ts:101`). **Nothing orders by `updatedAt`** — grepping `orderBy` across `src/` returns those three call sites and no fourth.

Requiring `updatedAt` is still right: both adapters always send it, `hasOnly` already permits it, and a document without it is one no write path produces. But that is a different argument from the invisible-to-paging one, and as written the paragraph lends the export/delete reasoning to a field it does not apply to. A reader auditing this later checks whether a listing orders by `updatedAt`, finds none, and has to work out whether the comment or the rule is wrong.

Worth splitting the sentence: `createdAt` and `finishedAt` because a listing orders by them, `updatedAt` because it is written on every create and every update and its absence means the document did not come from this application.

firestore.rules
```diff
@@ -225,8 +234,8 @@ service cloud.firestore {
         // value of its own choosing — and before this line that value could be
         // a string of any length. `keepsCreatedAt` pins createdAt to what is
         // stored, which says nothing about the create that stored it.
-        && d.get('createdAt', request.time) is timestamp
-        && d.get('updatedAt', request.time) is timestamp
+        && d.createdAt is timestamp
+        && d.updatedAt is timestamp
```
kowinauyeung, line 247: ✅ This pair of lines is the sharpest change in the pull request and the description undersells it.

`d.get('createdAt', request.time) is timestamp` was not a weaker check — for an **absent** field it was no check at all. The default it falls back to is itself a timestamp, so a create with no `createdAt` satisfied the predicate through the default and never touched the document. It only ever caught a present-but-wrongly-typed value.

So the fix is not "tighten a bound", it is "the guard was reading its own default". That is the kind of thing that survives review indefinitely because the line looks like it is doing the work.

firestore.rules
```diff
@@ -244,7 +253,17 @@ service cloud.firestore {
...
+        // The rest of what `wordSetFields()` permits. Naming a field and never
+        // reading it is the hole `hasOnly` was added to close, and closing it
+        // for entries alone left the same megabyte reachable one collection over
+        // — measured, not inferred: `level` took 900,000 characters.
+        && str(d.get('level', ''), 10)
+        && (d.get('publishedId', null) == null || str(d.get('publishedId', ''), 200))
+        && d.get('publishedVersion', 0) is int && d.get('publishedVersion', 0) >= 0
+        && (d.get('copiedFrom', null) == null || d.get('copiedFrom', {}).size() <= 10)
+        && d.createdAt is timestamp
+        && d.updatedAt is timestamp;
```
kowinauyeung, line 275: ✅ Confirmed against every write path rather than by reading the allowlist: `wordSetRepo.create` (`wordSetRepo.ts:74-84`) sends `publishedId: null`, `publishedVersion: 0`, `copiedFrom: null` and both stamps, `wordSetRepo.update` (`:88-92`) re-sends `updatedAt` and relies on the merged resource for `createdAt`, and `migration/upload.mjs:228-229` writes both as `Timestamp`. Every new bound clears real data by a wide margin — `level` is `JlptLevel | ''`, so at most two characters against a cap of ten.

tests/rules/firestore-rules.test.ts
```diff
@@ -417,6 +431,29 @@ describe('bounds on what an owner may write', () => {
...
+  /**
+   * The half of the allowlist argument that was only applied to entries.
+   *
+   * `validEntry` bounds every scalar it permits. `validWordSet` and
+   * `validSession` name theirs and check some — so `level`, `publishedId`,
+   * `copiedFrom`, `mode`, `startedAt`, `finishedAt` and both timestamp pairs
+   * were permitted and unread, which is the same megabyte under a different
+   * key that `hasOnly` was introduced to stop.
+   */
+  it('refuses an oversized value in a word set or session field too', async () => {
```
kowinauyeung, line 443: 🟡 **The docblock above names seven fields; the assertions reach four of them.**

Named: `level`, `publishedId`, `copiedFrom`, `mode`, `startedAt`, `finishedAt`, both timestamp pairs. Asserted: `level`, `mode`, `startedAt`, and word-set `createdAt`.

I reverted `firestore.rules` to the base and re-ran the emulator suite to see what this test actually holds down. It goes red, so it is a real test — but `publishedId`, `publishedVersion`, `copiedFrom` and `finishedAt`'s type have **never been red**, and by this repository's own standard that makes their failure mode unknown. They are also the fields most likely to be quietly dropped in a later edit, precisely because nothing fails when they are.

Four more `assertFails` lines in this same test would close it: an oversized word-set `publishedId`, a negative `publishedVersion`, a `copiedFrom` map past ten keys, and a session `finishedAt` sent as a string.

(Independently raised by a second reviewer pass, which is some evidence the gap is visible from outside.)

tests/rules/firestore-rules.test.ts
```diff
@@ -653,7 +691,38 @@ describe('what the comments in the rules claim', () => {
...
+  });
+
+  /**
+   * The combination that breaks a promise rather than a query.
+   *
+   * Optional in the rules and ordered-by in the adapter is invisible data: it
+   * passes every check, never appears in a page, and so is never exported and
+   * never deleted. `Account.tsx` and the privacy policy both say otherwise.
+   */
+  it('refuses a row with no createdAt, which no listing would ever return', async () => {
```
kowinauyeung: 🔵 The name says `createdAt`, and two of the three assertions are about something else: the word-set case is missing both stamps, and the session case is missing `finishedAt`.

The property is "a row without the field its listing orders by", which is stronger and is the one the rules comment argues for. Something like `refuses a row with no ordered timestamp, which no listing would ever return` names the defect rather than one instance of it.

.github/workflows/deploy-dev.yml
```diff
@@ -3,7 +3,19 @@ name: Deploy (dev)
...
+  # toolchain. A release or hotfix pull request into `main` breaks the
+  # assumption rather than the rule: `main` had no `.nvmrc`, so `setup-node`
+  # failed four seconds in and put a red check on the release.
+  #
+  # It is not a required check there, so it never blocked anything. It was worse
+  # than that: a red light nobody is meant to act on, on the one pull request
+  # where somebody should be reading every light. A dev preview has no business
+  # on a release in the first place.
   pull_request:
+    branches: [develop]
```
kowinauyeung: 🟡 **The cause was a pull request into `main`; the filter also catches stacked pull requests, which never had the problem.**

A pull request based on another feature branch now gets no preview channel. Its base does carry `.nvmrc`, so the deploy job's assumption holds there and the check was green — it is collateral, not the thing being fixed.

`ci.yml:4-7` argues the opposite case in the same repository, and in so many words:

> Every pull request, whatever it targets. Filtering on the base branch would leave a pull request stacked on another feature branch unverified until it was retargeted, which is exactly when it is easiest to merge something red.

The two files can legitimately differ — verification and a preview deploy are not the same obligation — but they now hold opposite positions on base-branch filtering and only one of them explains why. `branches-ignore: [main, 'release/**']` would remove the red light from releases without touching the stacked case; whichever way it goes, the comment here should say that stacked pull requests were considered, since the file three directories over says they were the reason not to filter.

Separately: **`README.md:181` now states something false.** "every pull request gets its own Hosting preview channel that expires after seven days" was accurate before this line and is not after it. The README is the public description of this repository's deploy model, and this is the release pull request.

kowinauyeung:
## Review

**Verdict: APPROVE.** No blockers. Everything below is a comment, a document or a test that has never been red — nothing that changes what this branch does.

### What I verified rather than read

I ran the emulator suite locally (JDK 21): **72 passed**, matching the description.

Then the part that mattered more. The repository requires a test to have been red before it is claimed to pass, so I reverted `firestore.rules` to `release/0.1.0`, kept the new tests, and re-ran:

```
× bounds on what an owner may write > refuses an oversized value in a word set or session field too
× what the comments in the rules claim > refuses a row with no createdAt, which no listing would ever return
  Tests  2 failed | 70 passed (72)
```

Exactly the two new tests, and **only** those two. Restoring the rules returns 72/72. The claim in the description is accurate as stated.

I also traced the new bounds against every write path rather than against the allowlist: `entryRepo.create/update`, `wordSetRepo.create/update`, `progressRepo.recordSession`, and `migration/upload.mjs`. All three adapters send both stamps; the migration writes them as `Timestamp` via the Admin SDK. `level` is `JlptLevel | ''` against a cap of ten, `mode` is one of two words against twenty, `startedAt` is a 24-character ISO instant against forty. Nothing real is close to a bound, and no existing path is now refused.

### The finding worth reading first

The change from `d.get('createdAt', request.time) is timestamp` to `d.createdAt is timestamp` is described here as tightening a bound. It is more than that: for an **absent** field the old line was no check at all, because the default it fell back to is itself a timestamp. It only ever caught a present-but-wrongly-typed value. A guard that reads its own default is the sort of line that survives review forever, because it looks like it is working.

### What is left open

| # | Where | What |
|---|---|---|
| 1 | `firestore.rules:109-116` | "Every `list()` orders by one of them" is true for `createdAt` and `finishedAt`, and false for `updatedAt` — nothing orders by it. The rule is right; the reason given for it is borrowed. |
| 2 | `README.md:181` | "every pull request gets its own Hosting preview channel" stops being true with this change, on the release pull request. |
| 3 | `tests/rules/firestore-rules.test.ts:435-443` | The docblock names seven fields, the assertions reach four. `publishedId`, `publishedVersion`, `copiedFrom` and `finishedAt`'s type have never been red. |
| 4 | `.github/workflows/deploy-dev.yml:18` | The cause was a pull request into `main`; the filter also removes previews from pull requests stacked on a feature branch, which `ci.yml:4-7` explicitly argues against filtering out. |
| 5 | description | The full-collection scan is deferred on "production has no data at all", while `firestore.rules:145-150` says the migration has been run there — 67 words. |

On (5): the deferral is still correct, but for a different reason than the one given. Those 67 documents are not at risk because `migration/upload.mjs:228-229` writes `createdAt` and `updatedAt` as `Timestamp`, so none of them is invisible to a listing and none of them will be refused on its next edit. That is worth stating in the description in place of "no data", because the next person to weigh a backfill will weigh it against whichever sentence is written down.

### On the two things deliberately left out

The liability clause and the operator identity are correctly the author's call and I am not second-guessing either. One observation only, since this is the release: the argument made here is not about what the law requires but about `terms.ts` not doing what its own docblock says it does. That is a self-consistency defect in a public document, and it is decidable without a lawyer.

### Cross-check

Run independently by a second model. It surfaced the `README.md` drift, which I confirmed by reading the file; its other findings overlapped with mine. It did not run the suite — every test result above is from my own run.

firestore.rules
kowinauyeung: @kowinauyeung
Upheld. `grep -rn "orderBy(" src/` returns three call sites — `entryRepo.ts:51` and `wordSetRepo.ts:43` on `createdAt`, `progressRepo.ts:101` on `finishedAt` — and no fourth. Nothing orders by `updatedAt`.

The consequence you name is the one that decides it: an auditor goes looking for the listing that orders by `updatedAt`, does not find it, and is then entitled to conclude that either the comment or the rule is wrong. In a file with a suite named `what the comments in the rules claim`, a paragraph that borrows a justification for a field it does not cover is a defect in its own right.

Split, with the distinction called load-bearing so it does not get merged back:

- `createdAt` and `finishedAt` — a listing orders by them, and Firestore omits a document lacking the field it orders by, so such a row is never exported and never deleted.
- `updatedAt` — nothing orders by it. Both adapters write it on every create and every update, so a document without one did not come from this application. Worth refusing on its own; not the paging argument.

Fixed in 60efaec

firestore.rules
kowinauyeung: @kowinauyeung
This is a better statement of the defect than the one in the description, and I had not seen it that sharply.

> for an **absent** field it was no check at all. The default it falls back to is itself a timestamp, so a create with no `createdAt` satisfied the predicate through the default and never touched the document.

`request.time` is a timestamp, so `d.get('createdAt', request.time) is timestamp` is a tautology whenever the key is missing. It only ever caught a present-but-wrongly-typed value — which is the case nothing produces, since no write path sends a wrong type. The guard was reading its own default and reporting success.

I had described this as making an optional field required. It is closer to removing a line that looked like a check and never was, which is exactly why it survived: `is timestamp` on the right-hand side reads like the work is being done.

firestore.rules
kowinauyeung: @kowinauyeung
Thanks — and specifically for checking each bound against the write path rather than against the allowlist. `level` being `JlptLevel | ''` and therefore at most two characters against a cap of ten is the kind of margin worth having confirmed rather than assumed, because a bound that is merely plausible is one somebody tightens later on a hunch.

tests/rules/firestore-rules.test.ts
kowinauyeung: @kowinauyeung
Upheld, and thank you for reverting `firestore.rules` to base and re-running rather than reading the assertions. That is the check that distinguishes "this test is real" from "these four lines are real".

You are right on both halves: the docblock named seven and the assertions reached four, and the four it missed had **never been red**. By this repository's own standard that makes their failure mode unknown — and they are exactly the ones a later edit drops quietly, because nothing fails when they go.

Four `assertFails` added in the same test: an oversized word-set `publishedId`, a negative `publishedVersion`, a `copiedFrom` map of eleven keys, and a session `finishedAt` sent as an ISO string.

Then measured, because adding them would otherwise reproduce the problem you are describing. Removing **only** those four bounds and leaving the rest of the change intact:

```
× refuses an oversized value in a word set or session field too
× refuses a row with no ordered timestamp, which no listing would ever return
70 passing
```

Both are now red for their own reasons rather than green by construction.

Fixed in 60efaec

tests/rules/firestore-rules.test.ts
kowinauyeung: @kowinauyeung
Agreed — the name described one of the three assertions and the weakest one.

`refuses a row with no ordered timestamp, which no listing would ever return`. That is the property the rules comment argues for, and it covers the word-set and session cases that were already in the body.

A test whose name is narrower than its assertions is the shape that gets trimmed to match its name.

Fixed in 60efaec

.github/workflows/deploy-dev.yml
kowinauyeung: @kowinauyeung
Upheld on both, and the second is the more serious.

**The filter was wider than the defect.** What broke was a base without the toolchain, which is `main` and nothing else; a stacked pull request's base carries `.nvmrc` and its deploy was green throughout. `branches: [develop]` took a preview away from a case that never had the problem.

`branches-ignore: [main, 'release/**']` now. The comment records that stacked pull requests were considered and quotes the position `ci.yml` takes — filtering on the base leaves a stacked pull request unchecked until it is retargeted, which is when it is easiest to merge something red. Verification and a preview deploy are different obligations, but not different enough for two files in one repository to reach opposite conclusions without either of them saying so.

**`README.md:181` was made false by my change, on the release pull request.** 「every pull request gets its own Hosting preview channel」 was accurate before it and not after. Corrected to state the exception and why it exists — the deploy job checks out the base, so it needs a base that carries the toolchain.

That one is worth naming for what it is: this pull request removes a comment that outlived the mechanism it described, and I introduced another one three directories away in the same change.

Fixed in 60efaec

firestore.rules
kowinauyeung: Verified in `60efaec`. The paragraph now stops at `createdAt` and `finishedAt`, names which repository orders by which, and gives `updatedAt` its own paragraph with its own reason.

The sentence that closes it — a reader looks for the listing that orders by `updatedAt`, finds none, and is entitled to conclude the comment or the rule is wrong — is a better statement of the problem than the one I opened with. Resolving.

firestore.rules
kowinauyeung: Nothing outstanding here. Resolving.

firestore.rules
kowinauyeung: Nothing outstanding. Resolving.

tests/rules/firestore-rules.test.ts
kowinauyeung: Verified, and I did not take the measurement on trust — the point of the finding was that these four had never been red, so a claim that they now are is the one thing worth re-running.

Removing all four bounds at once reproduces exactly what you reported: the same two tests red, 70 passing. But that only establishes that *at least one* of the four is load-bearing, which is the same weakness as the docblock naming seven fields. So I neutralised them one at a time, four runs:

```
removed publishedId       -> Tests  1 failed | 71 passed (72)
removed publishedVersion  -> Tests  1 failed | 71 passed (72)
removed copiedFrom        -> Tests  1 failed | 71 passed (72)
removed finishedAt        -> Tests  2 failed | 70 passed (72)
```

Each of the four is individually red-capable. `finishedAt` taking down two is correct rather than surprising: `d.finishedAt is timestamp` is also what refuses the session in the ordered-timestamp test.

Resolving.

tests/rules/firestore-rules.test.ts
kowinauyeung: Verified renamed in `60efaec`, and the three assertions in the body are now all instances of the name. Resolving.

.github/workflows/deploy-dev.yml
kowinauyeung: Both verified.

**The filter.** `branches-ignore: [main, 'release/**']` — and I checked it against this pull request rather than against the syntax, which is the only way to be sure a filter pattern does what it reads like. `gh pr checks 35` on this branch returns `e2e`, `verify`, `emulator` and no `deploy-dev`: the base is `release/0.1.0`, so it is correctly excluded, while a stacked pull request keeps its preview.

**The README.** Corrected, and it now states the exception and the mechanism behind it rather than just the exception.

One thing left in each, small enough that I have raised them as fresh comments rather than reopening this: the YAML says `release/**` while the comment above it and `README.md:182` both say `release/*`, and under GitHub's filter-pattern rules those are different patterns. Resolving this thread.

src/content/terms.ts
```diff
@@ -74,8 +88,9 @@ export const terms: Doc = {
     {
       heading: '責任について',
       body: [
-        '本サービスは無償で提供しています。運営者は、本サービスの利用により生じた損害について、運営者に故意または重過失がある場合を除き、責任を負いません。',
-        '運営者の故意または重過失により生じた損害については、前記の免責にかかわらず責任を負います。',
+        '本サービスは無償で提供しています。',
+        '運営者の故意または重過失により生じた損害については、運営者がその全部について責任を負います。',
+        '運営者の軽過失により生じた損害については、通常生じうる損害の範囲に限り責任を負います。特別の事情によって生じた損害、逸失利益および間接損害については、責任を負いません。',
```
kowinauyeung, line 93: ✅ **Reviewed properly this time — my first pass never saw this commit, and I have said so in the summary.**

The structure is right, and one line of the docblock is the reason:

> Deliberately no monetary cap — the service is free, so a cap at "the amount paid" is a total exclusion wearing a number.

That is the trap most free-service terms walk into, and naming it is worth more than the clause it justifies.

**I checked for the same hole under a different heading**, since that is what this whole pull request is about. If another section had said 責任を負いません about data loss, the total exclusion would be back by another route. It does not: バックアップ (`:84`) and 内容の正確性 (`:69`) both say 保証しません — a no-warranty statement, not a liability exclusion — so 責任について is the only place liability is allocated, and the docblock now describes it.

Not a lawyer, and the file says the same about itself. What I can say is that the comment and the text agree now, which is the defect that was actually reported.

src/content/privacy.ts
```diff
@@ -20,6 +20,7 @@ export const privacy: Doc = {
       body: [
         `語彙庭は、${site.operator}が個人で運営するサービスです。法人による運営ではありません。`,
         '本ポリシーに関するお問い合わせは、お問い合わせフォームよりご連絡ください。返信をご希望の場合は、フォーム内で返信先をご記入ください。',
+        '運営者の氏名および住所は、お問い合わせフォームからのご請求に応じて、遅滞なくお知らせします。',
```
kowinauyeung, line 23: 🟡 **This commit changed the policy text and did not touch `updated` — and it is only correct because of what day it is.**

`privacy.updated` was already `2026-08-14` on the base, and this commit landed on 2026-08-14, so the date happens to be right. Land the same diff tomorrow and the policy ships changed text under a stale date, with nothing red.

That is not a hypothetical standard I am importing. **The documents promise it themselves** — `terms.ts:99` and `privacy.ts:105` both say 「このページの更新日を改めます」. So there is a commitment in the product text with nothing enforcing it, which is the exact shape this pull request exists to close in `firestore.rules`: a promise made in one place and kept only by whoever remembers.

`content.test.ts` currently asserts `updated` *matches a date format*, not that it *moves when the body does*. The mechanical version is the same trick that file already plays twice — it reads `client.ts` for App Check and `site.feedbackFormUrl` for Google Forms rather than hard-coding either. Hashing each doc's body and pinning the hash beside its `updated` would make an unbumped date go red instead of relying on the edit landing before midnight.

Suggestion, not a blocker: the date is right today, and a release is not the moment to add a test that pins content. Worth an issue if it does not fit here.

.github/workflows/deploy-dev.yml
```diff
@@ -3,7 +3,28 @@ name: Deploy (dev)
 on:
   push:
     branches: [develop]
+  # Every pull request except those into `main` and `release/*`.
```
kowinauyeung: 🔵 The comment says `release/*` and the configuration two dozen lines down says `release/**`. Under GitHub's filter-pattern rules those are different: `*` does not match `/`, `**` does. So `release/2026/q1` is ignored by the filter and not by the sentence describing it.

`release/**` is the one you want — it is the pattern that covers `release/0.1.0` — so this is the comment to change, not the config. `README.md:182` says `release/*` as well.

Raising it because it is the same class of defect the rest of this pull request is fixing: a description that is very slightly wider or narrower than the mechanism it describes, in a file whose whole argument is that those diverge silently.

.github/workflows/deploy-dev.yml
```diff
@@ -3,7 +3,28 @@ name: Deploy (dev)
...
+  # requests are why.** A pull request based on another feature branch has a
+  # base that does carry the toolchain; it was never broken, and an allowlist
+  # would have taken its preview away as collateral. `ci.yml` argues the same
+  # case for verification in so many words — filtering on the base leaves a
+  # stacked pull request unchecked until it is retargeted, which is exactly when
+  # it is easiest to merge something red. Verification and a preview deploy are
+  # different obligations, but not different enough to reach opposite
+  # conclusions here without saying so.
   pull_request:
+    branches-ignore: [main, 'release/**']
```
kowinauyeung, line 31: ✅ Verified against the pull request rather than against the syntax, which is the only check that settles what a filter pattern actually does: `gh pr checks 35` returns `e2e`, `verify` and `emulator`, and **no `deploy-dev`**. The base here is `release/0.1.0`, so the exclusion fires; a stacked pull request keeps the preview it never should have lost.

The comment above is also the right length now — it records that stacked pull requests were considered and quotes the position `ci.yml` takes, so the two files no longer reach opposite conclusions in silence.

README.md
```diff
@@ -177,9 +177,12 @@ split — `develop` for dev, `main` for production. Both are protected: changes
...
 
-Pushing to `develop` deploys hosting and Firestore rules to `goitei-dev`, and
-every pull request gets its own Hosting preview channel that expires after seven
-days. Preview channels are hosting-only: rules are project-wide, so deploying
+Pushing to `develop` deploys hosting and Firestore rules to `goitei-dev`, and a
+pull request gets its own Hosting preview channel that expires after seven days
+— every pull request except one into `main` or a `release/*` branch, which get
+CI and no preview. The deploy job checks out the pull request's base, so it
+needs a base that carries the toolchain, and a release into `main` is the one
+case that does not. Preview channels are hosting-only: rules are project-wide, so deploying
```
kowinauyeung: 🔵 This line is 91 characters where the rest of the paragraph is hand-wrapped at 75–80. Prettier will not catch it: `.prettierrc` sets `printWidth: 100` but leaves `proseWrap` at its default of `preserve`, so Markdown prose is never reflowed and `yarn prettier --check` passes on this file — I ran it.

Breaking after `case that does not.` puts it back in line with its neighbours.

kowinauyeung:
## Re-review — `60efaec`

**Verdict: APPROVE.** All six threads from the first round are fixed, verified and resolved. What is left is one suggestion and two nitpicks, none of which changes behaviour.

### First, a correction to my own review

**My first pass diffed the wrong commit.** I ran `git diff origin/release/0.1.0...HEAD` against a stale local `HEAD` at `0a41b16`, while the pull request head was already `2a976f2`. So the review covered three files and never saw `docs(legal): bound the liability that was excluded, and name a way to reach the operator`.

That has a consequence beyond a missing file. My summary had a section headed "On the two things deliberately left out", arguing that the `terms.ts` docblock contradicted its own clause — **which you had already fixed, two hours before I posted it.** Reviewing an absence that was not there is worse than missing the commit, because it reads as a considered position. I have now reviewed `2a976f2` on its merits and left a comment on it.

The fix on my side is to take the head from the API rather than from a local checkout, which is what I did this round.

### The four fixes, verified

| Thread | Verified how | Status |
|---|---|---|
| `updatedAt` borrowing the paging argument | Read the split paragraph; `grep -rn 'orderBy(' src/` still returns three call sites and no fourth | Resolved |
| Four bounds never red | Neutralised each one **individually**, four separate emulator runs | Resolved |
| Test name narrower than its assertions | Renamed; all three assertions are now instances of the name | Resolved |
| Workflow filter + `README.md` | `gh pr checks 35` — no `deploy-dev` on this pull request | Resolved |

On the second row, the reason I re-ran rather than took the measurement: removing all four bounds at once shows only that *at least one* of them is load-bearing, which is the same weakness as a docblock naming seven fields and reaching four. One at a time:

```
removed publishedId       -> Tests  1 failed | 71 passed (72)
removed publishedVersion  -> Tests  1 failed | 71 passed (72)
removed copiedFrom        -> Tests  1 failed | 71 passed (72)
removed finishedAt        -> Tests  2 failed | 70 passed (72)
```

Each is individually red-capable. `finishedAt` taking two is correct — `d.finishedAt is timestamp` is also what refuses the session in the ordered-timestamp test. Suite green at head: **72 passed**. `yarn prettier --check` clean on every changed file.

On the fourth row, the filter is verified by observation rather than by reading the pattern: this pull request's base is `release/0.1.0` and its checks are `e2e`, `verify`, `emulator` and no `deploy-dev`.

### `2a976f2`, reviewed now

The clause no longer excludes ordinary negligence outright; it bounds it by kind of damage. One line of the docblock is worth more than the clause it explains — *a cap at "the amount paid" is a total exclusion wearing a number* — because that is the trap a free service walks into while believing it has written a limitation.

I checked for the same hole under a different heading, since that is this pull request's whole subject. There is none: バックアップ (`terms.ts:84`) and 内容の正確性 (`:69`) say 保証しません, which is a no-warranty statement rather than a liability exclusion, so 責任について is the only place liability is allocated and the docblock now describes it. Not a lawyer; the file says the same about itself.

### Open, none blocking

1. 🟡 **`privacy.ts` changed text without touching `updated`, and is correct only because of the calendar.** The date was already `2026-08-14` and the commit landed on `2026-08-14`. The same diff tomorrow ships a changed policy under a stale date with nothing red — while `terms.ts:99` and `privacy.ts:105` both promise 「このページの更新日を改めます」. A promise in the product text with nothing enforcing it is the shape this pull request closes in `firestore.rules`. `content.test.ts` already reads `client.ts` and `site.feedbackFormUrl` rather than hard-coding what it checks; the same trick would work here.
2. 🔵 `release/**` in the YAML, `release/*` in the comment above it and in `README.md:182`. Different patterns — `*` does not match `/`.
3. 🔵 `README.md:185` is 91 characters against a paragraph wrapped at 75–80. Prettier will not catch it: `proseWrap` is at its default of `preserve`.

(1) is worth an issue if it does not fit in a release. (2) and (3) are one line each.

src/content/terms.ts
kowinauyeung: @kowinauyeung
Thanks — and for going looking under the other headings rather than reading the one section that was reported. That is the check I did not do: if バックアップ or 内容の正確性 had said 責任を負いません, the total exclusion would have come back by another route and the docblock would have been wrong again for a new reason.

That they say 保証しません is the distinction that makes the file coherent — a warranty disclaimer allocates no liability, so 責任について really is the only place it is allocated. Worth having confirmed rather than assumed, since the two read similarly in Japanese and the difference is the whole of it.

src/content/privacy.ts
kowinauyeung: @kowinauyeung
Agreed on the diagnosis, and the framing is the part that makes it more than housekeeping:

> a promise made in one place and kept only by whoever remembers

`terms.ts:99` and `privacy.ts:105` both say 「このページの更新日を改めます」, and nothing enforces it — which is the same shape as the rule this pull request closed in `firestore.rules`, one layer up. `content.test.ts` pins the *format* of `updated`, not that it moves when the body does, and the mechanical version really is the trick that file already plays twice: it reads `client.ts` for App Check and `site.feedbackFormUrl` for Google Forms rather than hard-coding either.

**Not in this pull request**, for your reason and one more. Yours: a release is not the moment to add a test that pins content. Mine: it is now 2026-08-15, so adding it here would mean touching the documents to bump their dates in the same change that introduces the check — which is a worse first run than one where the check simply passes.

Raised as an issue instead, with the hashing approach and the `content.test.ts` precedents named so it does not arrive as a bare "add a test".

The date is correct today, and correct by coincidence, which is exactly what you said.

.github/workflows/deploy-dev.yml
kowinauyeung: @kowinauyeung
Upheld, and precise. GitHub's filter patterns follow glob rules, so `*` stops at a `/` and `**` does not — `release/2026/q1` is described by the narrower pattern and not matched by it.

`release/**` is what the configuration has and what it should have, so both descriptions moved to it: the comment here and `README.md:182`. The comment now also says *why* it is `**`, so the next person to tidy it back to `*` meets the reason first.

> the same class of defect the rest of this pull request is fixing: a description that is very slightly wider or narrower than the mechanism it describes

That is the third instance in this pull request — the `updatedAt` justification, the `README` preview claim, and now this. All three were introduced by changes whose purpose was to stop exactly that.

Fixed in 3597d12

.github/workflows/deploy-dev.yml
kowinauyeung: @kowinauyeung
Thanks — and `gh pr checks 35` returning no `deploy-dev` is the check that settles it. A filter pattern is one of the few things where reading the syntax and reading the result can disagree and only the second one counts.

README.md
Agreed, and thank you for establishing *why* `prettier --check` passes rather than just noting that it does. `proseWrap` defaulting to `preserve` means Markdown prose is never reflowed, so `printWidth: 100` describes what Prettier would produce and not what it enforces — a distinction that is invisible until a hand-wrapped file drifts.

The line was 91 against neighbours at 75–80, because I appended a sentence to an existing line instead of rewrapping the paragraph.

Rewrapped, and split in two while there: which pull requests get a preview, and why previews are hosting-only, were one paragraph doing two jobs.

Fixed in 3597d12

src/content/terms.ts
kowinauyeung: Nothing outstanding. Resolving.

src/content/privacy.ts
kowinauyeung: Deferral accepted, and I checked the follow-up exists rather than taking "raised as an issue" on the word — a deferred finding with no ticket behind it is just a closed thread.

#36 is there and it is not a stub: it carries the promise both documents make, the concrete near-miss that surfaced it, the hashing approach, both `content.test.ts` precedents, and the open question of whether the hash belongs in the test or beside the document. Somebody picking it up cold has what they need.

Your second reason is better than mine. Adding the check today would mean bumping both documents' dates inside the same commit that introduces it — so its first run proves nothing, because the thing it exists to catch was hand-corrected on the way in. A check whose first green is manufactured is one nobody trusts later.

Resolving.

.github/workflows/deploy-dev.yml
kowinauyeung: Verified in `3597d12`. Both descriptions now say `release/**`, and the comment gives the reason — `*` stops at a `/`, `**` does not — so the next person to tidy it back meets the argument before the edit.

On the tally: three instances in one pull request, each introduced by a change whose purpose was to stop exactly that. Worth reading as a property of the class rather than as three mistakes. A description sits next to the mechanism, is never executed, and stays true only while somebody re-reads it — which is the same argument this branch makes about `updatedAt` being optional in the rules and ordered-by in the adapter. The two `content.test.ts` tests that read `client.ts` and `site.feedbackFormUrl` are the shape that actually holds, and #36 is the next one.

Resolving.

.github/workflows/deploy-dev.yml
kowinauyeung: Nothing outstanding. Resolving.

README.md
kowinauyeung: Verified in `3597d12`: 75–81 characters across the paragraph, in line with the rest of the file.

Splitting it in two was the right call and more than tidying — which pull requests get a preview and why previews are hosting-only are separate claims, and the second is the one a reader needs when they wonder why their rules change did not appear. Resolving.

kowinauyeung:
## Re-review — `3597d12`

**Verdict: APPROVE.** Every thread from both rounds is fixed, verified and resolved. Nothing outstanding, nothing deferred without a ticket.

### Verified at head

| | |
|---|---|
| `yarn test` | 573 unit and component, 72 emulator, all green |
| `yarn prettier --check .` | clean |
| `yarn lint` | 0 errors, 17 warnings — all in files this branch does not touch |
| `gh pr checks 35` | `verify`, `emulator`, `e2e` pass; no `deploy-dev`, which is the filter working |

`3597d12` is documentation only, so I read it rather than re-deriving the rules behaviour. `README.md` is 75–81 characters across the paragraph and now split in two, and both descriptions of the filter say `release/**`.

### The deferral, checked rather than accepted

#36 exists and is not a stub. It carries the promise both documents make, the near-miss that surfaced it, the hashing approach, both `content.test.ts` precedents, and the open question of where the hash belongs. A deferred finding with no ticket behind it is just a closed thread; this one has somewhere to go.

Your second reason for keeping it out is better than mine. Adding the check today would mean bumping both documents' dates in the same commit that introduces it — so its first run would prove nothing, because the thing it exists to catch was hand-corrected on the way in.

### What this branch is, at the end of it

The rules change is the substance and it survived every check I could put on it: each of the four bounds individually red-capable, the two required-timestamp guards red when reverted, every adapter write path traced against the new predicates, and the migration's 67 documents confirmed to carry both stamps.

The rest of the branch turned out to be about one thing, and it is worth naming because three separate instances of it appeared *during* the review: **a description that drifts a little wider or narrower than the mechanism it sits beside.** The `updatedAt` justification, the `README` preview claim, and `release/*` against `release/**` — each introduced by a change whose purpose was to stop exactly that. The two tests in `content.test.ts` that read `client.ts` and `site.feedbackFormUrl` rather than hard-coding an expectation are the shape that actually holds, and #36 is the next one of those.

One correction of mine stands on the record: my first pass diffed a stale local `HEAD` and missed `2a976f2` entirely, then argued in its summary for something that commit had already done. Reviewed on its merits in round two.

Ready to merge.

PR #37, opened 2026-08-14
chore(release): sync 0.1.0 back to develop

kowinauyeung:
## Summary

- sync the release-only migration claim validation from #34 back to `develop`
- sync the release-only Firestore rule bounds and required query fields from #35
- restore `develop` as the baseline for all new feature work

## Testing

- no additional code changes are introduced by this synchronization PR
- the existing CI suite will verify the combined release state against `develop`


kowinauyeung:
Superseded by a pull request from `main` rather than from `release/0.1.0`.

Same content — `main` holds nothing the release branch does not — but merging from `main` makes it an ancestor of `develop`. That costs nothing today and matters the first time a hotfix lands on `main` directly: with this shape the next back-merge carries it, and without it the fix has to be remembered and cherry-picked.

PR #53, opened 2026-08-16
chore: retire the migration folder

### LINKED ISSUE #26: upload.mjs still writes allowedUsers and reports it as confirming access

kowinauyeung:
The 67 hand-written notes are loaded into production and were exercised through the app, which is the condition `migration/README.md` set for removing itself. This does that, and records the accents that went with them.

## Two commits, and why they are not one

**`feat(migration): fill 37 pitch accents`** — every entry still carried `pitchAccent: null`. Without this commit the last committed `output.json` would not match what is actually in production, and the artefact would be deleted having never recorded what it uploaded.

**`chore: retire the migration folder`** — the removal.

## How the accents were chosen

`yarn migrate:accents prompt` emits the word list with the mora count already worked out per row, so the question asked is only *which mora the pitch drops on* — counting mora is the part an assistant gets wrong.

Two models answered it independently. 37 numbers matched exactly, 18 words were `null` on both sides, 7 numbers disagreed and 5 more were a number from one side only. **Only the 37 agreements were written**, and `apply` re-checks every value against that word's own mora count before writing it — none was refused.

The 7 disagreements were then researched against NHK 日本語発音アクセント新辞典, 大辞林 and OJAD with an instruction to answer `null` without a source. None could be sourced, so all 7 stay `null`. The 5 one-sided numbers were declined on the prompt's own terms: four are colloquial, and `身に染みる` is several accent phrases rather than one word.

30 of 67 remain `null` and that is the correct state, not a gap. A wrong accent is invisible — nobody can tell 2 from 3 by looking at the line — so it gets memorised, which is worse than not knowing.

## Test layer, and the red verification

No new tests. The suite loses one file: `tests/unit/migrationOutput.test.ts`, whose job was to prove the data conformed **before** the upload. That upload has happened, so the guard is discharged rather than dropped.

Before removing it, it was made to fail, because it had been green over 67 nulls and green proves nothing there. Injecting a quoted `"0"` (what an assistant routinely returns) and a `2.7` into `output.json` turned **exactly that file red and nothing else**:

```
❯ tests/unit/migrationOutput.test.ts (14 tests | 1 failed)
+   "sagesumu.pitchAccent: 2.7 → null"
+   "choukou.pitchAccent: \"0\" → 0"
Test Files  1 failed | 40 passed (41)
```

Restored and re-run green. That test was the only layer that could see this data: the upload runs through the Admin SDK and `tests/integration` uses `mockUserToken: 'owner'`, and both bypass rules entirely, so `validEntry`'s null-or-int bound was unreachable at upload time.

## What the removal touches

The folder, plus every point that referenced it:

| | |
| --- | --- |
| `tests/unit/migrationOutput.test.ts` | removed — see above |
| `eslint.config.js` | two ignores, two `files` globs |
| `tsconfig.scripts.json` | `include` is now `["admin"]` |
| `package.json` | three `migrate:*` scripts |
| `README.md` | five references |
| `LICENSE` | the scope paragraph |
| `.gitignore` | two entries and a comment block that no longer applied |
| `firestore-rules.test.ts`, `vocabulary.spec.ts` | comments pointing at deleted paths |

**`migrationKey` stays on the entry allowlist in `firestore.rules`.** Every migrated document carries it, nothing in `src` names it, and an allowlist rebuilt by reading the TypeScript type would be complete, reviewable, green in the emulator, and would refuse every edit to all 67 migrated words in production only. That was the finding in #29. What makes it survivable is that `tests/rules/firestore-rules.test.ts` pins it and never lived in the deleted folder — the comment there now says so.

## One unrelated correction, found because the paragraph named the folder

`src/lib/accountData.ts` documented an invariant that user settings ended:

> this function deliberately does not delete `users/{uid}`, because that document does not exist. Nothing writes it: … `UserRepository` has no adapter, and `firestore.rules` allows the client `read, delete` on the path and nothing else
>
> **That is an invariant, not an observation**, and it is the reason there is no parent step below.

All three claims are false since #50: `createUserRepository` writes the document, `userSettings.tsx` calls `save()`, and the rules allow `create, update`. The **code was already correct** — `deleteEverything` removes the profile after the subcollections — so only the rationale above it was stale. Rewritten to say what the function does and why the ordering is what it is.

## Checks

`yarn typecheck` (all three passes) · `yarn lint` 0 errors / 17 warnings, unchanged · `yarn format:check` · `yarn test:unit` 596 passed · `yarn test:emulator` 77 passed.

## Issues

Closes #26 and #14 in substance — both were filed against files this removes, and #26 had additionally been fixed by #34 already. Closed by hand rather than by keyword, since `Closes` only fires on a merge to the default branch.

## Summary by CodeRabbit

- **Documentation**
  - Updated repository, licensing, authentication, and deployment guidance to reflect the current workflow.
  - Clarified that historical import artefacts remain available only in repository history.

- **Chores**
  - Removed obsolete data-import scripts, documentation, generated outputs, and related commands.
  - Updated linting and script-check configuration to cover only active administrative tooling.
  - Refined ignore rules for service-account files and backup documentation.

- **Tests**
  - Removed obsolete import-output conformance tests and updated remaining test documentation.


kowinauyeung:
@qodo-merge review

Issue #65, opened 2026-08-18
The manifest and the hosting headers have no test, and offline navigation is never exercised


kowinauyeung:
Nothing added for the manifest, deliberately, because a manifest has no
behaviour a test can go red on until something reads it. Three things then do,
and each belongs at a different layer.

**`tests/unit` — the manifest's shape.** Required fields present, both icon
sizes Chromium demands, a `maskable` entry, and `start_url` and `scope`
agreeing. This is the layer that catches an icon renamed in `public/` and not
in the manifest, which otherwise surfaces as an install that silently has no
icon.

**`tests/unit` — the hosting cache headers.** Once #62 lands, assert the rules
in `firebase.json` the way `tests/unit/csp.test.ts` already asserts the policy:
the worker, `index.html` and the manifest revalidate, and `/assets/**` does not.

**`tests/e2e` — offline navigation, and only once a service worker exists.**
`browserContext.setOffline(true)`, then reach the dashboard. This is the only
layer that can see the defect it is for: every route in `src/router.tsx` is a
`lazy: () => import(...)`, so an uncached chunk fails to load offline and the
boundary renders `ErrorScreen`. No unit or component test can observe that —
the failure is in what the network did not serve.

Per CLAUDE.md, prove each fails before claiming it passes, and add no screenshot
baseline without asking.

Follow-up to #66.

kowinauyeung:
Landed on `develop` in #91 (`b41d0e2`). Closing by hand for the same reason as #63 and #64: the `Closes` keyword only fires when a pull request merges into the default branch, and this repository merges into `develop` first.

Issue #64, opened 2026-08-18
viewport-fit=cover is set but no layout reads env(safe-area-inset-*), so a standalone window collides with the home indicator


kowinauyeung:
`index.html` has carried `viewport-fit=cover` since before the manifest, which
tells the browser to lay the page out under the notch and the home indicator.
Nothing then compensates: `src/index.css` contains no `env(safe-area-inset-*)`
anywhere, and neither does any component.

In a browser tab this is mostly hidden, because Safari's own chrome occupies
the unsafe strips. Installed, there is no chrome — the app owns the full screen,
and the header runs under the notch while anything anchored to the bottom sits
under the home indicator.

To do:

- Add the insets where the layout meets the edges, which is `AppLayout.tsx`
  first.
- Check navigation with no browser back button. Every route reachable in
  `standalone` has to be leavable from inside the app.

Follow-up to #66.

kowinauyeung:
Landed on `develop` in #86 (`7f39f06`). Closing by hand: the `Closes #64` keyword only fires when a pull request merges into the default branch, and this repository merges into `develop` first.

Issue #63, opened 2026-08-18
A dropped connection renders as a failed read, with nothing telling it apart from a denial


kowinauyeung:
`src/infra/firebase/client.ts` already configures `persistentLocalCache`, so
the data half of working offline is done: a review session on a train keeps
reading and syncs when the connection returns. Nothing in `src/` reads
`navigator.onLine` or listens for `online`/`offline`, so none of that is
visible to the person using it.

What is missing:

- **A state indicator.** Offline is a state the app is in, not an error that
  happened, and it currently has no representation at all.
- **`src/lib/loadError.ts` cannot name it.** The module splits denial from
  everything else and puts a dropped connection in "everything else", where it
  gets the subject-specific fallback — 単語を読み込めませんでした — for a
  condition that is neither about the word nor about the read. The distinction
  it already draws for `permission-denied` is the one to extend.
- **App Check offline.** `initializeAppCheck` refreshes its token over the
  network. Whether an expired token turns cached reads into `permission-denied`
  while offline has not been established, and it decides whether the cache is
  actually reachable in the state it exists for.
- **`persistentSingleTabManager`.** One person on one device was the assumption
  and an installed app breaks it: the installed window and a browser tab are two
  tabs, and the second one loses persistence. `persistentMultipleTabManager` is
  the swap, and it is a swap with a cost worth measuring before taking it.

Separately, the fonts. `index.html` pulls three families from Google Fonts, so
they are cross-origin, and a service worker can only cache them as opaque
responses — no status, no way to tell a hit from a failure. For an app whose
subject is Japanese text with furigana, the fallback is not a cosmetic
degradation. Self-hosting is the fix, and it also closes an existing gap: the
Content-Security-Policy allows neither `fonts.googleapis.com` in `style-src`
nor `fonts.gstatic.com` in `font-src`, so promoting it from Report-Only breaks
them today.

Follow-up to #66.

kowinauyeung:
Landed on `develop` in #82 (`6b1618c`). Closing by hand for the same reason as #64: the `Closes` keyword only fires when a pull request merges into the default branch, and this repository merges into `develop` first.

Issue #40, opened 2026-08-16
A deleted account's ID token stays valid, so a second tab can write the data back


kowinauyeung:
## The race

One account is enough. Two tabs, or a phone and a laptop:

1. Device A presses delete. `deleteEverything` re-authenticates, drains every
   collection, and calls `auth.deleteAccount()`.
2. Device B is signed in as the same uid and already holds a minted ID token.
3. A finishes. The Firestore data is gone and the Auth user is gone.
4. B writes — a practice session finishing, an edit in an open form — and the
   write succeeds, because the token is still inside its validity window.

The result is an account that no longer exists owning documents that do, under a
uid nothing will ever issue a token for again. They are unreachable by the app
and invisible to the export and delete paths, which both start from a signed-in
session.

## Why the token still works

Firestore rules verify the token's signature and claims. There is no revocation
check and cannot be one: rules are evaluated without a round trip to Auth.
A Firebase ID token is valid for about an hour after it is minted, so deleting
the Auth user does not invalidate tokens already in the wild.

https://firebase.google.com/docs/auth/admin/manage-sessions

## Correcting the earlier reasoning

This was previously deferred as "it only matters with more than one account".
That is wrong. Having one account lowers the *impact* — the only data at risk is
the operator's own — but the defect is entirely within a single uid and is
reachable by anyone who leaves a second tab open. Deferring it is a decision
about impact, not a statement that it cannot happen.

## What is missing

`src/lib/accountData.ts` and `src/infra/firebase/authAdapter.ts` carry no
tombstone and no revocation metadata, so nothing after step 3 can tell the
difference between a live session and a dead one.

Directions, none of them free:

- **A tombstone document** the rules require to be absent. Costs a read on every
  write, which on Spark is the resource under most pressure.
- **`revokeRefreshTokens` plus a `auth_time` floor in rules.** Needs the Admin
  SDK, so it needs a backend.
- **Delete the Auth user first**, accepting that a failure mid-way strands the
  data instead — trading an unbounded window for a bounded one, and arguably
  worse.

Blocks inviting a second person. Does not block the owner-only first release.

PR #72, opened 2026-08-20
[Draft] Release 0.2.0

kowinauyeung:
`main` is 39 commits and one week behind `develop`, and the gap has started
costing things rather than just sitting there.

## Why now

**The `Closes #N` keyword has stopped working.** It only fires when a pull
request merges into the default branch, and everything here merges into
`develop` first — so #63, #64 and #65 all had to be closed by hand, each with a
comment explaining why. That is not a one-off; it is the new normal until the
two branches meet.

**Reviews pointed at the default branch report fixed problems as live ones.** A
review of the release risk came back naming three comment/implementation
mismatches. All three were real — on `main`. `firestore.rules:287` still says
`mode` is "one of four words" there; on `develop` #74 rewrote that comment and
it now records the correction explicitly. `accountData.ts` on `main` never
deletes the parent document its own docblock describes deleting; on `develop` it
does. `client.ts` on `main` still says App Check enforcement should wait for
metrics; on `develop` it says production was enforced from its first deploy and
why. Every one of those cost time to disprove.

## What this does not do

**It does not deploy anything.** `deploy-prod.yml` is `workflow_dispatch` only
and takes an explicit 40-character SHA, so production stays a deliberate act.
This merge makes `main` current and nothing else.

## What is in it

Nine features, seven fixes, three test and documentation changes, and the usual
dependency traffic. The parts worth naming:

- **The app is installable.** A manifest and its icons (#66), a service worker
  that precaches the shell and the route chunks with a prompt rather than an
  ambush (#78), `Cache-Control` so a stale shell cannot outlive a deploy (#77),
  offline rendered as a state rather than a failure (#82), and the device's
  safe-area insets read at last (#86) — with `/login` given a way out, since a
  `standalone` window has no back button.
- **Every user-written field is bounded** (#74), in the rules as well as the
  forms, and a long one no longer widens the page.
- **Localised settings and authenticated flows** (#50), the provider profile
  picture (#69), a pronunciation button (#56), refined tag filters (#38).
- **Four iOS and import fixes**: the focus zoom (#73), substituted quotes (#71),
  a clipboard that refuses (#59), and the heatmap opening on today (#61).
- **Testing**: replayable monkey runs (#51), integration tests no longer trusting
  the emulator's clock resolution (#76), public documents pinned to prose hashes
  (#87), and the manifest and offline navigation covered (#91).

## Carried, deliberately

Two accepted residual risks go out with this, both written up rather than
forgotten:

- **#39** — rules bound list and map *cardinality* but cannot reach inside an
  element, so a nested value can still approach 1 MiB. Re-measured on `develop`
  after #74 and still open. Accepted for one operator account on Spark, where
  the ceiling is an outage rather than a bill; a decision is required before a
  second account exists.
- **#40** — a deleted account's ID token stays valid for about an hour, so a
  second tab can write data back after the account is gone. One account is
  enough to hit it; having one account lowers the impact, not the likelihood.

## One thing to settle before this leaves draft

**The title says 0.2.0 and `package.json` says `0.1.0`, on both branches.**
Nothing in these 39 commits bumps it. That matters beyond tidiness: the version
is compiled into the bundle by `build-info.ts` and rendered on the account page,
which is where a bug report's build line is read from — so shipping this as
0.2.0 by title alone would put a release name in the notes that no running
client can report.

Either bump `package.json` on `develop` before merging, or retitle. It is a
release decision, not a cleanup.


kowinauyeung:
wrong branch

PR #144, opened 2026-08-28
feat(dashboard): load vocabulary stats on demand

### LINKED ISSUE #140: Backfill vocabulary dashboard stats after deploy
## Context

Dashboard reads `users/{uid}/stats/vocabulary` for heatmap and distribution counters. A missing document uses a read-only paged fallback; the browser never publishes a full stats snapshot because it could overwrite a concurrent entry mutation.

## Post-deploy migration steps

1. Deploy the app and Firestore rules that support `users/{uid}/stats/vocabulary`.
2. Authenticate an operator identity with Firestore read/write access to production:

   ```sh
   yarn auth:login
   ```

3. Run the privileged, idempotent backfill and source parity check:

   ```sh
   yarn backfill:vocabulary-stats prod --sample 20
   ```

4. Record the command output. It reports active users rebuilt, stale deleted-account stats removed, and sampled users verified. A source/cache mismatch exits non-zero; correct the failure and rerun the same command.
5. Watch Dashboard load errors and Firestore rule denials after the backfill. Keep the client fallback for at least one release.
6. Close this issue only after the production command completes successfully and its sample verification is clean.

## Consistency protocol

The Admin SDK job enumerates every `users/{uid}` document, including empty notebooks. For a missing stats document it first transactionally creates a zeroed document. It then runs a second transaction that reads both the source entries and stats document before writing exact `total`, `countsByDay`, `jlptCounts`, and `posCounts` counters. A concurrent client delta conflicts with that transaction and is retried or applied after the source snapshot; it is never silently overwritten by a client-side full scan. Re-running the job for unchanged entries produces the same counters.

## Acceptance criteria

- Every existing user, including empty notebooks, has a stats document after a successful run.
- Empty notebooks receive a zeroed stats document.
- The operator command is privileged and idempotent.
- The requested production sample compares stored stats with a fresh source scan and fails non-zero on a mismatch.
- The concurrent missing-cache regression is covered by an Admin SDK migration integration test.

kowinauyeung:
### Summary

Load Dashboard vocabulary data on demand instead of mounting the full notebook. Persisted dashboard counters supply distributions and heatmap data, while recent words, selected-day words, and word of the day use targeted reads and pagination.

### Changes

- Added owner-gated vocabulary stats reads and atomic counter deltas.
- Kept account deletion, selected-day pagination, and Dashboard dialog refreshes consistent with the cache.
- Added an Admin SDK `yarn backfill:vocabulary-stats` command that rebuilds every user cache, sample-verifies source parity, and documents the production runbook.
- Removed client-side full-scan cache writes; a missing cache uses a read-only fallback so concurrent mutations cannot be overwritten.

### Test coverage

- Unit: cached day-bucket period totals, backfill argument validation, and source counter calculation.
- Adapter integration: cache bootstrap, immediate mutation visibility, paged selected days, and an Admin SDK missing-cache concurrent-create regression.
- Red-to-green proof: restoring query-then-plain-set drops the concurrent entry (`total: 1`); the transaction implementation preserves it. Restoring fire-and-forget stats deltas reads `total: 0` immediately after create; awaiting the delta passes.

### Verification

- `yarn format`
- `yarn typecheck`
- `yarn lint` (18 existing warnings, no errors)
- `yarn test:emulator` (113 tests)
- `yarn test:unit` (920/921 passed; existing local font fixture failure: `inter-latin-400-normal.woff2` missing from `node_modules`)

Closes #140

## Summary by CodeRabbit

* **New Features**
  * Added vocabulary dashboard statistics, including totals, daily activity, JLPT levels, and parts of speech.
  * Added recent learned entries, date-filtered browsing, and a deterministic Word of the Day.
  * Added reliable statistics rebuilding and verification tools for administrators.
* **Improvements**
  * Vocabulary details now load automatically when an entry is not already cached.
  * Entry changes update dashboard data more reliably.
  * Authenticated homepages load faster by avoiding unnecessary entry requests.
* **Documentation**
  * Added setup and production guidance for rebuilding vocabulary statistics.

### DIFF
--- a/HANDOFF-dashboard-on-demand-vocab.md
+++ b/HANDOFF-dashboard-on-demand-vocab.md
@@ -38,13 +38,14 @@ deploy` (<https://github.com/kowinauyeung/japanese-learning-notes/issues/140>)
 
 ## Status update
 
-The review blockers below have now been implemented in the uncommitted
-worktree. `saveDashboardStats()` bootstraps a complete cache document after the
-Dashboard fallback drain. CRUD awaits the stats delta before resolving, and the
-Dashboard derives period totals from cached day buckets rather than three server
-aggregates. The focused adapter regression was proven red by restoring
-`void writeStatsDelta(...)`, which read `total: 0` after create, then green with
-the awaited implementation. Full emulator verification passes with 113 tests.
+The client-side full-scan writer was removed after review identified its stale
+snapshot race. Dashboard now has a read-only fallback for a missing cache. The
+privileged `yarn backfill:vocabulary-stats` Admin SDK script creates and rebuilds
+every cache document transactionally, then sample-verifies source parity. CRUD
+awaits stats deltas, and Dashboard derives period totals from cached day buckets
+rather than three server aggregates. The focused adapter regression was proven
+red by restoring `void writeStatsDelta(...)`, which read `total: 0` after create,
+then green with the awaited implementation.
 
 ## Feature implemented so far
 
--- a/HANDOFF-dashboard-on-demand-vocab.md
+++ b/HANDOFF-dashboard-on-demand-vocab.md
@@ -1,0 +1,200 @@
+# Dashboard on-demand vocabulary handoff
+
+## Working agreement
+
+- Reply to the user in Cantonese, using Traditional Chinese. Keep repository
+  artifacts (commits, branches, PRs, GitHub issues, and review replies) in
+  English.
+- Read `AGENTS.md` before changing code. In particular, shared port changes
+  require enumerating downstream consumers, and regression tests must be proven
+  red before the fix is claimed green.
+- Use `apply_patch` for edits. Use `gh` for GitHub operations.
+- Do not commit without proposing the exact commit message and receiving the
+  user's explicit approval.
+
+## Repository state
+
+- Repository: `/Users/kowin/Documents/japanese-learning-notes`
+- Branch: `feat/dashboard-on-demand-vocab`
+- Base: `develop`
+- Existing GitHub issue: #140, `Backfill vocabulary dashboard stats after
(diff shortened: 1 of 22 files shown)


kowinauyeung:
Closing without merging.

This branch set out to stop the Dashboard loading the whole notebook, and does
it by adding a derived `users/{uid}/stats/vocabulary` cache that the client
maintains. Two problems, and the second is why it is being closed rather than
fixed.

**1. The cache cannot be kept correct from the client.** The reviews here are
right. `update()` derives its delta from a locally read `before`, so two devices
editing the same entry both subtract the same old value and the distribution
counters drift permanently. `update()` and `remove()` also skip the delta
outright when that pre-read fails — the ordinary offline path — and never retry
it. `increment()` makes concurrent creates and deletes safe; nothing here makes
an edit safe. Getting this right needs a trusted server-side processor, which is
out of scope for the current plan.

**2. It does not address the load it was meant to address.** `AppLayout` passes
`loadOnMount={location.pathname !== '/'}`, so the full paged drain in
`EntriesProvider` is skipped on the dashboard route only. The vocabulary list —
the unbounded read this was about — still pulls every page into memory exactly
as it did before.

The real bottleneck is `EntriesProvider`: it renders below the auth gate, so
every route pays for the whole notebook. Replacement work starts there, with
paging for the vocabulary list. It keeps the cheap direct reads this branch
introduced — `countLearnedSince`, `recentLearned`, `learnedOn`, `wordOfDay` —
without a cache behind them.

Dashboard aggregates that genuinely require a full scan, the heatmap and the
JLPT and part-of-speech distributions, are deferred rather than approximated.

This also affects two issues: #140 exists only to migrate the cache and should
close with this, and #173 assumes the stats counter is available to rules.

The branch stays for reference.

